PR #29, opened 2015-01-17
Check type parameters in error:has_type/2

mndrix:
Types like between(L,U) and oneof(L) require that their parameters
be of a specific type.  Verify this is true before relying on the
assumption.

### DIFF
--- a/library/error.pl
+++ b/library/error.pl
@@ -279,7 +279,7 @@ has_type(atom, X)	  :- atom(X).
 has_type(atomic, X)	  :- atomic(X).
 has_type(between(L,U), X) :- (   integer(L)
 			     ->  integer(X), between(L,U,X)
-			     ;   number(X), X >= L, X =< U
+			     ;   number(X), number(L), number(U), X >= L, X =< U
 			     ).
 has_type(boolean, X)	  :- (X==true;X==false), !.
 has_type(callable, X)	  :- callable(X).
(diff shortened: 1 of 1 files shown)


mndrix:
Incidentally, this makes queries like this possible:

```prolog
?- dif(T,impossible), error:has_type(T,7).
T = any ;
T = atomic ;
T = constant ;
T = ground ;
T = acyclic ;
T = integer ;
T = nonneg ;
T = positive_integer ;
T = nonvar ;
T = number ;
T = rational ;
false.
```

It's helpful for certain kinds of static analysis.

drbeco:
Using

```
:~$ uname -a
Linux raposa 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2 (2016-04-08) x86_64 GNU/Linux
```

I got only these 2 values for `T`:


```
:~$ swipl
Welcome to SWI-Prolog (Multi-threaded, 64 bits, Version 6.6.6)
Copyright (c) 1990-2013 University of Amsterdam, VU Amsterdam
SWI-Prolog comes with ABSOLUTELY NO WARRANTY. This is free software,
and you are welcome to redistribute it under certain conditions.
Please visit http://www.swi-prolog.org for details.

For help, use ?- help(Topic). or ?- apropos(Word).

?- dif(T,impossible), error:has_type(T,7).
T = any ;
T = atomic ;
ERROR: >=/2: Arguments are not sufficiently instantiated
?- 
```

triska:
Relatedly, we currently have:

<pre>
?- must_be(impossible, t).
<b>ERROR: Arguments are not sufficiently instantiated</b>
</pre>

But this is the wrong error to throw in this case! We expect a **domain&nbsp;error**, because `impossible` is not a valid type.

JanWielemaker:
error:has_type/2 is not a public interface.   The same error applies for is_of_type/2, which is public.  It has mode +,@ and thus your call is invalid.  Pushed some changes to make it a bit better behaved.
For this type of static analysis we should add additional primitives rather than calling internal predicates.

PR #30, opened 2015-01-19
Fix pack installation for GitHub releases

mndrix:
Trying to install a pack whose releases are hosted on GitHub was failing
with errors like "ERROR: V0.9.0.zip: A package archive must be
named <pack>-<version>.<ext>"  When downloading a release from GitHub,
store it locally using the proscribed naming convention.

This fixes issue mndrix/julian#8


JanWielemaker:
Thanks.  Merged by hand and pushed after changing commit message.  By starting such message
as [A-Z]+:, they get into the changelog.

PR #33, opened 2015-02-02
ls/1 fails if a filename without wildcards does not exist

borisvassilev:
This fixes the following unexpected behavior of ls/1:

~~~~
?- ls('foobar*').
Warning: No match: foobar*
false.

?- ls('foobar').
% foobar   
true.
~~~~


JanWielemaker:
Hmm.  Don't think this is neat.  The tagging is supposed to be det and used in a maplist/3.
This also doubles the number of filesystem calls.  Finally, we don't want a 'No Match' as
we did not provide a pattern, but an existence error.  I've pushed 87ee9e34d790b0c9420bfda2986e092ad0d674cf to resolve the issue.

PR #35, opened 2015-02-17
Goal expansion for foldl/4+N

borisvassilev:
This is NOT a pull request, just a request for comments.

I _think_ this is a correct implementation. Before a merge:

- There is a lot of code duplication between maplist/2+N and foldl/4+N expansion which has to be cleaned up
- I don't fully understand the code in prolog_clause:unify_goal :( so I have no idea if it is correct

Please comment so that I can take care of the outstanding issues (if you think this functionality is necessary in the first place.

### DIFF
--- a/library/apply_macros.pl
+++ b/library/apply_macros.pl
@@ -45,6 +45,7 @@ translated into an call  to  a   generated  auxilary  predicate  that is
 compiled using compile_aux_clauses/1. Currently this module supports:
 
 	* maplist/2..
+        * foldl/4..
 	* forall/2
 	* once/1
 	* ignore/1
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Hi Boris,

My grand plan goes in another direction: use partial evaluation to get rid of such meta calling loops.   Another grand plan is to have a special object that avoids the overhead of meta-calling by providing a direct pointer to the predicate.  That would allow maplist and friends to obtain this handle before entering the loop and it also allows goal expansion to take care of this part of the job.

The reason for the first is quite obvious if you look at the horrible code in library(apply_macros) :-)

That said, I'm not really against merging this after cleanup.  The role of the unify_goal hook is to
make the gui tracer understand the relation between the source term (holding foldl) and the
clause holding the modified goal.  So, it must tell that foldl(...) unifies with '__aux....'(....)

borisvassilev:
Hi Jan,

I am not emotionally attached to my code :) so if it doesn't fit in your plans, it is better left alone.

My reasons for writing it: I like using maplist, forall, foldl. I actually only noticed the expansion when I was playing with a small program that used `library(clpfd)`, which apparently uses `library(apply_macros)`. So I thought it doesn't hurt to have an expansion for foldl, too, so I wrote it. But as my email to the list also shows, I realized there are some really annoying limitations.

But now that I have your attention: how about `foldr/N+3`? I implemented this too (since I used it before I realized it's not available), but at the moment, it is not tail recursive:

~~~~
foldr(Goal, List, V0, V) :-
    foldr_1(List, Goal, V0, V).

foldr_1([], _, V, V).
foldr_1([H|T], Goal, V0, V) :-
    foldr_2(T, Goal, H, V0, V).

foldr_2([], Goal, X, V0, V) :-
    call(Goal, X, V0, V).
foldr_2([H|T], Goal, X, V0, V) :-
    foldr_2(T, Goal, H, V0, V1),
    call(Goal, X, V1, V).
~~~~

Is that worth adding (together with the version for 2, 3, 4 lists)? Or is it better to simply `reverse` and use `foldl`, and leave this out, too?

JanWielemaker:
As far as I recall, these predicates were based on Richard O'Keefe's [Prolog library proposal](http://www.cs.otago.ac.nz/staffpriv/ok/pllib.htm).  I see that Richard has done a lot of updates,
including proposing foldr.  I do think that a reverse and foldl is probably a better implementation
than as you do now (and easier).

Please take a look at Richard's work.  He also claims that Logtalk has a compiler for these things.
Maybe we can steal that (Paulo will be happy as long as we acknowledge)?   Further integration of parts of Richard's proposal are generally welcome.  He has a *lot* of experience and has taken the effort to study current practice.

JanWielemaker:

https://swi-prolog.discourse.group/t/add-to-library-apply-any-2-or-include-2-foldl-helper/5971/2

JanWielemaker:

https://swi-prolog.discourse.group/t/problem-with-foldl-4-and-library-apply-macros/7312/2

PR #37, opened 2015-03-23
offset condition was the wrong way around.

samer--:


JanWielemaker:
Thanks Samer.  Modified your patch by swapping the arguments rather than the comparator.
I think that makes the code a bit easier to read.

PR #59, opened 2015-06-23
Allows unset MACRO Wunused-result with gcc -DWUNUSED=""

drbeco:
This would allow a temporary (or permanent if so desired) disable of GCC -Wunsused-result, for functions that have __attribute__ ((warn_unsused_result)) set.

A GCC command line would add -DWUNUSED=""

This can allow a developer to keep the warn ON (no need to add -Wno-unused-result) for the entire project, without receiving warnings from SWI-Prolog.

### DIFF
--- a/src/SWI-Prolog.h
+++ b/src/SWI-Prolog.h
@@ -117,12 +117,14 @@ duplicated this stuff.
 		 *	  GCC ATTRIBUTES	*
 		 *******************************/
 
-#ifndef WUNUSED
-#if __GNUC__ >= 4
-#define WUNUSED __attribute__((warn_unused_result))
-#else
-#define WUNUSED
+#ifndef WNOUNUSED
+#  if __GNUC__ >= 4
+#    define WUNUSED __attribute__((warn_unused_result))
+#  endif
 #endif
+
+#ifndef WUNUSED
+#  define WUNUSED
 #endif
 
 
--- a/src/SWI-Prolog.h
+++ b/src/SWI-Prolog.h
@@ -117,10 +117,12 @@ duplicated this stuff.
 		 *	  GCC ATTRIBUTES	*
 		 *******************************/
 
+#ifndef WUNUSED
 #if __GNUC__ >= 4
 #define WUNUSED __attribute__((warn_unused_result))
 #else
 #define WUNUSED
+#endif
 #endif
 
 


JanWielemaker:
Thanks.  I'm not really sure we want this though.  Ignoring these return values typically leads to
crashes on stack overflows.  Worse, the crash not happen immediately (as in e.g. ignoring a malloc() return after which one typically writes to the memory quickly, causing a crash), but they lead to a failure recovering from a stack overflow much later.

drbeco:
Hi Jan. Thanks for taking your time to read this pull. I wrote an email to you earlier, to the one in the source code, but it returned. And so I decided to write the modification myself here at github. I'll send it again.

Regarding the modification, its more about giving options for a better, clean debug. Also, if a sequence of predicates are called from C, you can't handle all returned values.

Take a look at page 279, SWI-Prolog manual, 6.6.0. You can see there PL_get_chars(), PL_get_string_chars(), PL_get_name_arity(), all called, and no return value assigned.

Sometimes you need to get the return value, sometimes you don't. And if you, like me, love a code without warnings, we better have an option to at least disable temporarily, during build / debug stages.

Turning off -Wno-unused-warning is worst, because you would turn off warnings you might never realize from other sources, system_calls, etc. And hacking as I did in the stack-overflow answer I sent you by email is just a workaround of something that might be more useful if implemented directly inside the SWI-Prolog.h.

drbeco:
Actually, better would be line 114 to be:

#ifndef WNOUNUSED

So GCC would be as easy as -DWNOUNUSED

And prevents some nutcase to write something as -DWUNUSED="-- attribute -- ((deprecated))" or alike. 

Being WNOUNUSED just to check defined/non-defined prevents this, and works even better.

JanWielemaker:
Hi Ruben,

On 06/23/2015 10:02 AM, Ruben Carlo Benante wrote:
> Hi Jan. Thanks for taking your time to read this pull. I wrote an email
> to you earlier, to the one in the source code, but it returned. And so I
> decided to write the modification myself here at github. I'll send it again.
>
> Regarding the modification, its more about giving options for a better,
> clean debug. Also, if a sequence of predicates are called from C, you
> can't handle all returned values.

You have to.  Typically, you write

	if ( PL_...() &&
	     PL_...() ... )

> Take a look at page 279, SWI-Prolog manual, 6.6.0. You can see there
> PL_get_chars(), PL_get_string_chars(), PL_get_name_arity(), all called,
> and no return value assigned.

I almost thought you had a case, but PL_get_chars() can easily fail (if
char* cannot represent the characters) and thus this code can easily
crash. The example as is only works in a strict ISO-Latin-1 world should
be updated.

> Sometimes you need to get the return value, sometimes you don't. And if
> you, like me, love a code without warnings, we better have an option to
> at least disable temporarily, during build / debug stages.

The WUNUSED is tactically placed for all API functions where you really
should check the return value (you typically do not want to store the
return as most functions are boolean) as there is nothing in advance you
can use to guarantee you will not get a failure and crash. Well, may be
there are a few too many. I'd be happy to remove those. But in general,
PL_unify_* and PL_put_* use stack and thus may run out. PL_call_*() is
obvious. PL_get_*() can typically fail because the C type cannot handle
the often far more general Prolog type (char* cannot represent Unicode
strings, long long cannot handle big integers). Here, there are a couple
of exceptions. E.g., if something is an atom (checked by type),
PL_get_atom() will succeed. Same for PL_get_float(). I'd be happy to use
another macro for these and make that dependent on a preprocessor value.

> Turning off -Wno-unused-warning is worst, because you would turn off
> warnings you might never realize from other sources, system_calls, etc.
> And hacking as I did in the stack-overflow answer I sent you by email is
> just a workaround of something that might be more useful if implemented
> directly inside the SWI-Prolog.h.

I dispute that.  Ignoring the return values from most of these APIs is
pretty disasterous.  Your code seems fine, but in the event of a stack
overflow it suddenly breaks in a way that is very hard to relate back
to the particular missed test.

So, all I'll accept is a split in things you really should not ignore
and (the few) where it is defensible to ignore, assuming you did the
proper type test before.

	Cheers --- Jan

P.s.	Note that older versions (< 5.9 or so) performed a longjmp()
	on a resource error instead of returning.  Getting the old
	code fixed was an important reason for adding the warning.
P.s.	You can also use a #pragma in the broken source (but I'd still
	advice against that).

drbeco:
Hi Jan,

I agree with you that it is (of course) far better to simply check all the results, if you must. Thats ok.

But what is discussed here is a simple change that allows an option to turn of this checking, while keeping the warn on for other sources that might come from different languages (as I do integration with Prolog / C ). 

I don't see this option to be anything different of -Wno-unused-result, but for the fact that you can keep gcc with -Wunused-result on and clean the output a little bit from Prolog source only.

Not discussing the importance of all return code checks. Lets add this option, and be advised its an advanced option only to be used during DEBUG or so. I hope I made this point a little bit more clear.

JanWielemaker:
I'm sorry, but I see no reason to make it easy to suppress these messages for Prolog API calls and leave the in for anything else.  As I argued, ignoring is particularly harmful for the Prolog API because incorrect behaviour of your program is not unlikely and it can be really hard to trace back the error to the forgotten ignored test.  Adding this macro requires documenting it, which implies documenting something you should not want to begin with.

drbeco:
I see. Thanks anyway for the discussion. I hope I can collaborate in the future with other ideas. BTW, did you get my email?

JanWielemaker:
Hi Ruben,

On 06/23/2015 05:09 PM, Ruben Carlo Benante wrote:
> I see. Thanks anyway for the discussion. I hope I can collaborate in the
> future with other ideas. BTW, did you get my email?

Not the first time, but eventually yes.   I do accept a large part of
the pull requests, although often after some discussion or modification.
If you plan to put a lot of effort making a large pull request,
please contact me first.

	Cheers --- Jan

drbeco:
There is one more thing I would like to take a closer look. Its probably not here the place to discuss it, but here is the problem: [welcome msg](http://stackoverflow.com/questions/28152469/how-to-suppress-welcome-message-when-embedding-swi-prolog-in-c-c/30962215)

And for this issue here, I'll keep using a MACRO I discussed here [Ignoring ret](http://stackoverflow.com/questions/11888594/ignoring-return-values-in-c/30994180)
`#define igr(M) if(1==((int)M)){;}` or `#define failif(M) if(1==((int)M)){PL_fail;}`

The last could inspire you to add it, or some kind of variation of it, to swipl API.

JanWielemaker:
On 06/23/2015 05:18 PM, Ruben Carlo Benante wrote:
> There is one more thing I would like to take a closer look. Its probably
> not here the place to discuss it, but here is the problem: welcome msg
> <http://stackoverflow.com/questions/28152469/how-to-suppress-welcome-message-when-embedding-swi-prolog-in-c-c/30962215>
>
> And for this issue here, I'll keep using a MACRO I discussed here
> Ignoring ret
> <http://stackoverflow.com/questions/11888594/ignoring-return-values-in-c/30994180>
> |#define igr(M) if(1==((int)M)){;}| or |#define failif(M)
> if(1==((int)M)){PL_fail;}|
>
> The last could inspire you to add it, or some kind of variation of it,
> to swipl API.

I used to have failif(something) as TRY(something).  In the end I've
removed most of them from the code.  Also stopped using PL_succeed
and PL_fail, so functions only return using `return something`.  That
makes life a lot easier if you wonder why the bottom of your function
is not executed :-)

I've come to the conclusion that such utilities are better left to the 
developer.  It is easy enough.

I like combining sequences of calls to the APIs using &&, as in:

	if ( PL_... &&
	     PL_... &&

The last advice is to make not too many foreign predicates and keep
their interface to Prolog simple.  Prolog is much better at (dis)
assembling complicated Prolog terms.

	Cheers --- Jan

PR #61, opened 2015-06-23
PL_TRY(M): MACRO

drbeco:
To keep code clean while testing return values and doing proper debug/exit in case of problems.

This is a new approach to the -Wunused-result I did in other rejected pull, for good reasons.

Now, this MACRO can be used safely to keep an eye in middle results that are not so important to code, but still can cause crashs if not taken care of.

### DIFF
--- a/src/pl-builtin.h
+++ b/src/pl-builtin.h
@@ -207,6 +207,14 @@ is also printed if stdio is not available.
 #define SECURE(g) ((void)0)
 #endif
 
+/* If PL_func fail, exit */
+#define PL_TRY(M) do { if(0==((int)(M)))\
+                       {\
+                         DEBUG(1, Sdprintf("[PL_TRY fail file:%s line:%d]\n", __FILE__, __LINE__));\
+                         exit(EXIT_FAILURE);\
+                       }\
+                     } while(0)
+
 
 		 /*******************************
 		 *   NON-DET PREDICATE CONTEXT	*


JanWielemaker:
In the first place is pl-builtin.h not part of the public interface.  More importantly though, we should
not exit after a failing API call.  They are just normal and lead to Prolog backtracking or a Prolog exception when handled properly.  If they happen inside a foreign predicate, the caller must return with failure or inspect the reason (logical failure or exception) and act accordingly.  If they happen in an embedded situation the caller must always inspect the reason and act appropriately.  exit() is typically not appropriate.

drbeco:
I chose `pl-builtin.h` exactly because this MACRO is not supposed to be used by public. And near the `DEBUG` option to state that it's in the DEBUG category. But I'm ok to the info you provided that `exit(0)` is not supposed to be called. I think you believe `exit(1)` here would be used by the general public, not by developers. In pursuing of bugs, and in a not-production-code, `exit(1)` with a error code just for a debug session is perfectly acceptable.

drbeco:
Jam, I'd like to add that if the only problem you see now is the `exit(1)` I'd be perfectly ok with the exact same code without the `exit(1)` line. It would still give a debug message. Or we can call `atexit()`. If you allow me, I'll present another more complex piece of MACRO with `atexit()`. But now you need to give me some incentive ;) to work on it, because It seems I'm pulling things in a dead rock decision of yours.

JanWielemaker:
On 06/24/2015 08:19 PM, Ruben Carlo Benante wrote:
> Jam, I'd like to add that if the only problem you see now is the
> |exit(1)| I'd be perfectly ok with the exact same code without the
> |exit(1)| line. It would still give a debug message. Or we can call
> |atexit()|. If you allow me, I'll present another more complex piece of
> MACRO with |atexit()|. But now you need to give me some incentive ;) to
> work on it, because It seems I'm pulling things in a dead rock decision
> of yours.

My problem is twofold.  First of all, I've come to dislike macros that
return because they harm understanding the control flow, in particular
in the presence of locks or allocated resources.  Second, failing API
calls do not indicate an error (most of the time) and should (again
most of the time) be returned to Prolog where they will be mapped to
failure or a Prolog exception.

	Cheers --- Jan

PR #69, opened 2015-08-19
Rewriting sum(X) template of aggregate_all/3 to use nb_setarg/3

eazar001:
I merged the count and sum(X) template because they were so similar.

### DIFF
--- a/library/aggregate.pl
+++ b/library/aggregate.pl
@@ -162,15 +162,23 @@ aggregate(Template, Discriminator, Goal0, Result) :-
 %	solutions, i.e., the minumum and  maximum   of  an  empty set is
 %	undefined.
 
-aggregate_all(count, Goal, Count) :- !,
-	State = state(0, _),
-	(   Goal,
-	    arg(1, State, C0),
-	    C1 is C0+1,
-	    nb_setarg(1, State, C1),
-	    fail
-	;   arg(1, State, Count)
-	).
+aggregate_all(Template, Goal, Result) :-
+  (  Template = sum(X),
+     ignore(contains_var(X, Goal))
+  ;
+     Template = count,
+     X = 1
+  ),
+  !, % optimization
+  State = state(0),
+  (  Goal,
+     arg(_, State, S0),
+     S is S0 + X,
+     nb_setarg(1, State, S),
+     fail
+  ;
+     arg(1, State, Result)
+  ).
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I agree with the idea.  Just, the implementation is a bit weird.  First of all, what is `ignore(contains_var(X, Goal))` doing there?  contains_var/2 just succeeds or fails without doing anything, so if you wrap that in ignore/1, nothing will ever happen, right?

Style wise, I don't like the disjunction very much, surely not when followed by a cut.  It creates a long clause trying to do two things.  There is a much nicer way to write this:

```
aggregate_all(sum(X), Goal, Sum) :- !,
    ... % very much like the old count code.
aggregate_all(count, Goal, Count) :- !,
    aggregate_all(sum(1), Goal, Count).
```

If you make a new pull request along these lines, I'm happy to merge it.

eazar001:
I also agree with you that `ignore(contains_var(X, Goal))` is strange; however it was the most succint way I could maintain behavior of the original `aggregate_all/3`.

Take my example of the original `aggregate_all/3`
```
term(42).
term(42).
?- aggregate_all(sum(2), term(42), Result).
Result = 4.
```
Here the original predicate maintains this nice behavior where every ground `term(42)` will increment the counter by 2. My implementation also maintains this behavior. However, if I were to remove the `ignore/1` wrapper:
```
?- aggregate_all(sum(2), term(32), Result).
false.
```
Again, this was the most succint way that came to mind. If you still disagree with this choice I can find an alternative route. As for your second point regarding the style of my disjunction, I 100% agree with you that your `sum(1)` version is superior.

I'll wait for your final assessment/decision and then make the necessary changes and make a second attempt at a push sometime tomorrow. Thanks for your response.

JanWielemaker:
The result of `aggregate_all(sum(2), term(32), Result).`  should of course be `0`.  But, what is
the reason for the contains_var/2 call?  I agree that if you put it there you also need ignore around
it, but `ignore(contains_var(X,Goal))` is simply a slow alternative for `true`, unless I'm really confused :)

eazar001:
You beat me to the puch. I was just about to suggest completely removing `contains_var/2`.

eazar001:
Okay, well thanks for clearing that up =]. I'll remove the unnecessary `contains_var/2`, and reformat the disjunction per your suggestion. I'll send you the revision when I next get the chance.

PR #70, opened 2015-08-20
Rewriting sum(X) template to use nb_setarg/3 in constant space.

eazar001:
Sum(X) rewritten using nb_setarg/3. Count template rewritten in terms of sum(1).


JanWielemaker:
Merged.  I did edit the layout a bit and added a guard aggregate_all(Var, _, _) to catch instantiation errors.  That bug was already there ...

PR #71, opened 2015-08-20
Rewriting min(X) and max(X) to use constant space.

eazar001:
I put the `\+ \+ Goal` in there to cause failure when Goal is a ground term that is not a valid solution. Perhaps there is a better way?

### DIFF
--- a/library/aggregate.pl
+++ b/library/aggregate.pl
@@ -176,6 +176,12 @@ aggregate_all(sum(X), Goal, Sum) :- !,
 	   fail
 	;  arg(1, State, Sum)
 	).
+aggregate_all(max(X), Goal, Max) :- !,
+	aggregate_all_compare(max(_,X), Goal, Max).
+
+aggregate_all(min(X), Goal, Min) :- !,
+	aggregate_all_compare(min(_,X), Goal, Min).
+
 aggregate_all(Template, Goal0, Result) :-
 	template_to_pattern(all, Template, Pattern, Goal0, Goal, Aggregate),
 	findall(Pattern, Goal, List),
@@ -195,6 +201,18 @@ aggregate_all(Template, Discriminator, Goal0, Result) :-
 	pairs_values(Pairs, List),
 	aggregate_list(Aggregate, List, Result).
 
+aggregate_all_compare(Comp, Goal, Max) :-
+	\+ \+ Goal,
+	Comp =.. [_, S, X],
+	State = state(X),
+	(  Goal,
+	   arg(_, State, S),
+	   M is Comp,
+	   nb_setarg(1, State, M),
+	   fail
+	;
+	   arg(1, State, Max)
+	).
 
 template_to_pattern(All, Template, Pattern, Goal0, Goal, Aggregate) :-
 	template_to_pattern(Template, Pattern, Post, Vars, Aggregate),


JanWielemaker:
That can't be right.  Basically you need a state that allows you to distinguish the first solution.  On the
first solution you need to initialise your max/min.  On subsequent ones you update it.  If Goal
finally fails you check you have seen the first solution and then pick the min/max.  I don't see
this happening in your example.  I also fail to see the reason for the `min(_,X)` term.

eazar001:
That's probably because the predicate I made is not very transparent. The univ call in aggregate_call_compare will initialize the state to X from min(S,X). The first call of Goal will initalize that state to the first solution that unifies with X. the min(S,X) and max(S,X) calls are there so that I can state that `M is max(S,X)`. So for example if the first solution is term(2), then the first call will be max(2,2). The state will be state(2) on the first solution. I update the state with M. I apologize if this is written in a confusing way.

JanWielemaker:
I now see the intend.  It won't work though.  the `arg(_,State,S)` will bind the first argument, but on
backtracking this will be undone.  Should probably also be `arg(1,State,S)`?  Even if it works, it won't be very robust.  Suppose X is left unbound at some point?  That should raise an instantiation error.

eazar001:
I've run several test runs on it and it seems to run fine. I'm not sure I'm understanding you with respect to your concerns of X being unbound at some point, but _if_ i'm understanding you correctly the instantiation error should be raised by virtue of how `is/2` works if I'm not mistaken.

Nonetheless, I'm keenly aware that you have quite the advantage here in terms of experience and the fact that this is your implementation. Hence if see a can of worms here, there must be considierable merit in this vision. I'll respectfuly rescind this pull request if you don't think this is a valid reimplementation.

JanWielemaker:
After studying the code carefully, I must admit it is a rather brilliant solution to avoid complicated `is this the first solution' tests.   I guess the `arg(_,State,S)` was one of the things that confused me.  I now rewrote it into this, and likewise for min.  Sharing the two in one implementation makes it mostly harder to understand and doesn't safe that much IMHO.  We can avoid the \+ \+ Goal by demanding
that the max/min is bound.  As we are only talking numbers, this is reliable. 

```
aggregate_all(max(X), Goal, Max) :- !,
	State = state(X),
	(  call(Goal),
	   arg(1, State, M0),
	   M is max(M0,X),
	   nb_setarg(1, State, M),
	   fail
	;  arg(1, State, Max),
	   nonvar(Max)
	).
```

Edited and pushed.  Thanks!

eazar001:
No, thank _you_! =]

eazar001:
As a followup:
There is one remaining pair of templates that can be reimplemented to use constance space, namely min(X, Witness) as well as max(X, Witness). I wrote a draft version which seems to pass my test cases, the only drawback being it is pretty complicated compared to the last two commits. As an example, here is the min(X, Witness) predicate:
```
agg_all(min(X,W), Goal, min(Min,Witness)) :- !,
	State = state(X,W),
	Entry = state(false),
	(  call(Goal),
	   arg(1, State, M0), % Current min
	   arg(2, State, W0), % Current witness
	   M is min(M0,X),
	   (  M == M0
	   ->
	      arg(1, Entry, false),
	      nb_setarg(1,State,M),
	      nb_setarg(2,State,W0),
	      nb_setarg(1, Entry, true)
	   ;
	      nb_setarg(1, State, M),
	      nb_setarg(2, State, W)
	   ),
	   fail
	;  arg(1, State, Min),
	   arg(2, State, Witness),
	   nonvar(Min)
	).
```
It is rather unwieldy, and because of the way the specifications are for (min/max(X), Witness), I had to keep track of first solution explicitly. Should this be entered as a revision? Or do you think a better solution is in order?

PR #72, opened 2015-08-25
ENHANCED: Rewriting (min/max, Witness) templates to use constant space.

eazar001:
Rewriting the Witness variations of `min` and `max` templates.

### DIFF
--- a/library/aggregate.pl
+++ b/library/aggregate.pl
@@ -196,6 +196,58 @@ aggregate_all(min(X), Goal, Min) :- !,
 	;  arg(1, State, Min),
 	   nonvar(Min)
 	).
+aggregate_all(min(X,W), Goal, min(Min,Witness)) :- !,
+	State = state(X,W),
+	(  (  once(call(Goal))
+	   -> true
+	   ;  !, fail
+	   ),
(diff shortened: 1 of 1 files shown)


eazar001:
Just realized that I probably don't need those `once/1 wrappers`

JanWielemaker:
Evaluating Goal twice is not acceptable.  Goal can be prohibitively expensive or it can have side
effects.  I expect there is a similar solution as without witness.  I'd also prefer to stay away from
dif/2.  I don't like the library dependency.  I also wonder whether dif/2 behaves as expected regardless
of what Goal does.  Some built-ins and certainly some code relying on foreign extensions do not cooperate well with constraints.   Finally, X is min(M,X) is cute, but X < M is IMO simpler.

eazar001:
Hmm, I did't think about the case of a very expensive `Goal` actually. I originally did write a version that didn't rely on calling a `once(Goal)`, and then calling `Goal` for all the solutions on a second run, but they relied on redundant checks. Perhaps I missed a more obvious strategy the first time around. I'll close this for now, and (hopefully) a better solution will present itself.

PR #73, opened 2015-08-27
ENHANCED: Rewriting min/max Witness templates to use constant space.

eazar001:
I chose to use this "heavier looking code" for several reasons. Here are some alternatives that I attempted:
```
agg_all_1(min(X,W), Goal, min(Min,Witness)) :- !,
  State = state(X,W),
  (  call(Goal),
     arg(1, State, M0),
     M is min(M0,X),
     nb_setarg(1, State, M),
     M \== M0,
     nb_setarg(2, State, W),
     fail
  ;
     arg(1, State, Min),
     arg(2, State, Witness)
  ).
```
In this variation you will see it is clearly shorter than the implementation I chose. However, you'll also notice that the minimum is calculated, set as the state, and then checked again with respect to equality of the last minimum. This is necessary, because we don't want the wrong witness being bound to a repeat X in a solution. So here, although the code "looks cleaner", there is twice the amount of checks in this algorithm.

Also:
```
agg_all_2(min(X,W), Goal, min(Min,Witness)) :- !,
  State = state(X,W,LastMin),
  (  call(Goal),
     arg(1, State, M),
     (  X \== LastMin,
	X < M
     -> true
     ;  nb_setarg(1, State, X),
	nb_setarg(3, State, X),
	fail
     ),
     nb_setarg(1, State, X),
     nb_setarg(2, State, W),
     nb_setarg(3, State, X),
     fail
  ;
     arg(1, State, Min),
     arg(2, State, Witness)
  ).
```
In this version there is not too much of an improvement either. There is the possibility of avoiding writes alltogether, however we still have the possibility of reundant checks. Furthermore, the implementation that I eventually chose accounts for the odd mode variations in `Goal` (i.e. the ground, and mixed mode bound variations). These alternatives do not account for maintaining the behavior of the original. So if these variations were chosen, the final versions would most likely not be pristine. Ultimately, the version I chose to implement does this, and utilizes no redundant checks. It introduces two more state variables: `Flag` and `Solutions`. `Flag` checks whether or not we are passed the first solution (hence behavior changes once changed to true). `Solutions` checks to make sure that `Goal` actually has solutions in the first place. This acts as a check for questionable results returned in mode variations other than `(-,-)`.

### DIFF
--- a/library/aggregate.pl
+++ b/library/aggregate.pl
@@ -196,6 +196,60 @@ aggregate_all(min(X), Goal, Min) :- !,
 	;  arg(1, State, Min),
 	   nonvar(Min)
 	).
+aggregate_all(min(X,W), Goal, min(Min,Witness)) :- !,
+	State = state(X,W,Flag,Solutions),
+	Flag = false,
+	Solutions = false,
+	(  call(Goal),
+	   (  call(Flag),
+	      arg(1, State, M)
+	   ->
+	      X < M,
+	      nb_setarg(1, State, X),
+	      nb_setarg(2, State, W),
+	      fail
+	   ;
+	      nb_setarg(1, State, X),
+	      nb_setarg(2, State, W),
+	      (  nonvar(X)
+	      -> nb_setarg(3, State, true),
+		 nb_setarg(4, State, true)
+	      ;  instantiation_error(X)
+	      ),
+	      fail
+	   )
+	;
+	   arg(4, State, true),
+	   arg(1, State, Min),
+	   arg(2, State, Witness)
+	).
+aggregate_all(max(X,W), Goal, max(Max,Witness)) :- !,
+	State = state(X,W,Flag,Solutions),
+	Flag = false,
+	Solutions = false,
+	(  call(Goal),
+	   (  call(Flag),
+	      arg(1, State, M)
+	   ->
+	      X > M,
+	      nb_setarg(1, State, X),
+	      nb_setarg(2, State, W),
+	      fail
+	   ;
+	      nb_setarg(1, State, X),
+	      nb_setarg(2, State, W),
+	      (  nonvar(X)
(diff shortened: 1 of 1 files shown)


eazar001:
I also forgot to mention, when the `Flag` is set to false, the intialization routine is set up. All this work is necessary to prevent the need for "last minimum checks" or "equality checks" once the flag is set to true in the false branch. In the "true" branch, we simply have the intuitive comparison needed, and the corresponding mutations of the compound state term.

JanWielemaker:
I think I found a better and simpler one :)

```
agg_all(min(X,W), Goal, min(Min,Witness)) :- !,
  State = state(false, _Min, _Witness),
  (   call(Goal),
      (   State = state(true, Min0, _)
      ->  X < Min0,
          nb_setarg(2, State, X),
          nb_setarg(3, State, W)
      ;   nb_setarg(1, State, true),
          nb_setarg(2, State, X),
          nb_setarg(3, State, W)
      ),
      fail
  ;   State = state(true, Min, Witness)
  ).
```

You can edit and play at http://swish.swi-prolog.org/p/zUgPNOEB.swinb

eazar001:
Wow! That is a pretty sweet version. Backtracking over `State` bindings with the conditional in the fashion above makes for a much cleaner implementation indeed. Thank you.

eazar001:
However, I'm definitely more in agreement with the latest edit you have on swish. This revision handles all the other edge cases as well.

JanWielemaker:
Made another change to the SWISH version.  Do you make a new pull request or shall I put
this in library(aggregate)?

eazar001:
I'm perfectly fine with you putting this in. I'm just happy to have the upgrade =]
If by the time I arrive home you haven't pushed it yourself, then I'll just start a new request if you prefer it that way.

PR #84, opened 2015-10-26
DOC: Document interaction between options `expand/1' and `file_type/1…

wouterbeek:
…' in `absolute_file_name/3'.


JanWielemaker:
I don't agree with that.  It is not a specific interaction between these two options.  There is a general pattern that some options affect the search (expand, extensions, relative_to, ...) and some are conditions (file_type is a mix, access is a clear condition).  If the end result is empty it raises an existence error.   I'd be happy with a change to the overall documentation to make this mechanism more explicit.

PR #88, opened 2015-11-07
Updated listing/1  to show the proper defining module

logicmoo:
This fixes the issue with autoloaded predicates displaying the defining module incorrectly the first time you type 

?- listing(ls).

(the second calling to listing was correct before this patch, and continues to correct after this patch)


logicmoo:
Jan fix already

PR #91, opened 2015-11-17
FIXED:  existence_error(variable,'$variable_names')  In var_property/…

logicmoo:
FIXED:  existence_error(variable,'$variable_names')  In var_property/2   for newly created threads

Perhaps new threads should create it if its missing? 

% [Thread 21] Got exception existence_error(variable,'$variable_names') (Ctx0=system:b_getval/2, Catcher=system:'<meta-call>'/1)

### DIFF
--- a/boot/expand.pl
+++ b/boot/expand.pl
@@ -365,7 +365,7 @@ prop_var(fresh(Fresh), Var) :-
 	 ;   Fresh = true
 	 ).
 prop_var(name(Name), Var) :-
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I think using nb_current is fine.  That is also what happens in `prolog_load_context/2`.  Basically you are supposed to use these predicates only during goal/term expansion as called from the compiler,
but there may of course be other reasons to call them ...

PR #93, opened 2015-11-21
FIXED - Listing no longer removes meta_predicate/1s and module_transp…

logicmoo:
FIXED - Listing no longer removes meta_predicate/1s and module_transarent/1s

I know secretly meta_predicate/1 one day might only work if put right after the module exports but for now it'd be good to leave in the listing.


logicmoo:
Closing this pull request as it will been better implemented from a pull request making `module_transparent/1` only shows up when `meta_predicate/1` wont imply `module_transparent/1`.

PR #98, opened 2015-12-25
MANPAGE: Document "swipl file1 ..." invocation, as in swipl --help.

triska:
This documents the important

    swipl file1 file2 ...

invocation method of SWI-Prolog in the manpage. This also aligns the manpage with the `swipl --help` message a bit better.


JanWielemaker:
Modified and little and added a new commit to explain the four calling recipes

PR #104, opened 2016-01-04
FIXED: listing of predicates that were made transparent by means other than meta_predicate/1

logicmoo:
FIXED: listing of predicates that were made transparent by means other than meta_predicate/1

Such as declared by foreign code.

### DIFF
--- a/library/listing.pl
+++ b/library/listing.pl
@@ -229,17 +229,27 @@ decl(volatile,	   volatile).
 decl(multifile,	   multifile).
 decl(public,	   public).
 
+meta_implies_transparent(Head):-
+	compound(Head),
+	arg(_,Head,Arg),
+	% DM: Do we need '^' as well?
+	(number(Arg); Arg=':'), !.
+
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Merged after edit.  Note that this is mostly theoretical as there is no point in meta_predicate if it is not for making the predicate module-sensitive (and thus _transparent_).

logicmoo:
On Sun, Jan 3, 2016 at 11:54 PM, Jan Wielemaker <notifications@github.com>
wrote:

> Merged after edit. Note that this is mostly theoretical as there is no
> point in meta_predicate if it is not for making the predicate
> module-sensitive (and thus *transparent*).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SWI-Prolog/swipl-devel/pull/104#issuecomment-168605250>
> .
>

Thank you for the edits.


I agree I'll send the high bandwidth why I do discussion towards the
mailing list.

Issue #106, opened 2016-01-18
assignAttVar: Assertion failed


triska:
Let `sat3.pl` consist of:

    :- use_module(library(clpfd)).

    sat(X) :- X in 0..9.

    num(L) :-
	    solve(As,Bs,Cs,Ds),
	    append([As,Bs,Cs,Ds], Vs),
	    findall(., labeling([ff], Vs), Ls),
	    length(Ls, L).

    solve([A1,A2,A3,A4],[B1,B2,B3,B4],[C1,C2,C3,C4],[D1,D2,D3,D4]) :-
	    maplist(sat, [A1,A2,A3,A4,B1,B2,B3,B4,C1,C2,C3,C4,D1,D2,D3,D4]),
	    A1 + A2 + A3 + A4 #= B1 + B2 + B3 + B4,
	    A1 + A2 + A3 + A4 #= C1 + C2 + C3 + C4,
	    A1 + A2 + A3 + A4 #= D1 + D2 + D3 + D4,
	    A1 + A2 + A3 + A4 #= A1 + B1 + C1 + D1,
	    A1 + B1 + C1 + D1 #= A2 + B2 + C2 + D2,
	    A1 + B1 + C1 + D1 #= A3 + B3 + C3 + D3,
	    A1 + B1 + C1 + D1 #= A4 + B4 + C4 + D4,
	    A1 + A2 + A3 + A4 #= A1 + B2 + C3 + D4,
	    A1 + B2 + C3 + D4 #= A4 + B3 + C2 + D1.

Then I get:

    $ swipl -O sat3.pl
    Welcome to SWI-Prolog (Multi-threaded, 64 bits, Version 7.3.15-7-gad4db9c)
    Copyright (c) 1990-2015 University of Amsterdam, VU Amsterdam
    SWI-Prolog comes with ABSOLUTELY NO WARRANTY. This is free software,
    and you are welcome to redistribute it under certain conditions.
    Please visit http://www.swi-prolog.org for details.

    For help, use ?- help(Topic). or ?- apropos(Word).

    ?- num(_).
    [Thread 1 (main) at Mon Jan 18 19:59:27 2016] pl-attvar.c:174: assignAttVar: Assertion failed: (__PL_ld->stacks.global.top)+8 <= (__PL_ld->stacks.global.max) && (__PL_ld->stac
    ks.trail.top)+6 <= (__PL_ld->stacks.trail.max)
    C-stack trace labeled "assert_fail":
[...]

PR #107, opened 2016-01-18
Raise the GLOBAL_OVERFLOW (if that is really happening)

logicmoo:
https://github.com/SWI-Prolog/swipl-devel/issues/106

@triska  Hi Markus, this should be helpful.    

Notes:   Old code used to perform this check in do_unify() .. But since then we have split the responsibility of assignment out of do_unify() the right thing would be to remove the checks for there .  Might include that into this patch

### DIFF
--- a/src/pl-attvar.c
+++ b/src/pl-attvar.c
@@ -1387,6 +1387,8 @@ PRED_IMPL("$attvar_assign", 2, dattvar_assign, 0)
   deRef(av);
   if ( isAttVar(*av) )
   { deRef2(valTermRef(A2), value);
+    /* FALSE on the next line isn't a hiding of trouble it will still raise the GLOBAL_OVERFLOW */
+    if(!allocGlobal(0)) return FALSE;
     assignAttVar(av, value, ATT_ASSIGNONLY PASS_LD);
   } else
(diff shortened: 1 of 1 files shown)


logicmoo:
Nope, I looked.  I think it should be left in do_unify() (as well as where it's been added) since do_unify() still expects some extra space created for scheduling wakeups

logicmoo:
> But since then we have split the responsibility of assignment out of do_unify()  ..

What I should have said:

Since we have split the responsibility between  do_unify() and attvar_assign/2 ,  We are just as responsible for the same checks in attvar_assign/2 as what is done in do_unify().  

(Also we need it since even though it says AllocGlobal(0) that means to allocate 8 cells. In my eclipse_c branch I might experiment a tad with only requesting what is absolutely needed. In actuality

do_unifys()  AllocGlobal(0)s needs to reserve 8
bindConst()  AllocGlobal(0) needs to reserve 8
attvar_assign/2 should neede non reserved. 

You may ask, then why did I purposely do it wrong in this patch?
the old bindConst() never reserved the 8 and somehow stayed lucky.. So I error on the side of caution that if new bindConst() needs it we will still get our assert error.

PR #109, opened 2016-01-21
Patch as per request https://github.com/SWI-Prolog/swipl-devel/issues/108#issuecomment-173537799

logicmoo:
Patch as per request https://github.com/SWI-Prolog/swipl-devel/issues/108#issuecomment-173537799

The surviving attvar gets the attributes in the case of a swap

### DIFF
--- a/src/pl-attvar.c
+++ b/src/pl-attvar.c
@@ -182,7 +182,9 @@ assignAttVar(Word av, Word value, int flags ARG_LD)
       av = value;
       value = tmp;
       if (flags & ATT_ASSIGNONLY)
-      { *valPAttVar(*value)=*valPAttVar(*av);
+      { a = valPAttVar(*value);
+        TrailAssignment(a);
+        *a = *valPAttVar(*av);
       }
     } else if ( av == value )
       return;
--- a/src/pl-attvar.c
+++ b/src/pl-attvar.c
@@ -181,6 +181,9 @@ assignAttVar(Word av, Word value, int flags ARG_LD)
     { Word tmp = av;
       av = value;
       value = tmp;
+      if (flags & ATT_ASSIGNONLY)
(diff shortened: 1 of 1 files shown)


triska:
The issue explained in #108 still exists with this branch.

logicmoo:
I now believe the bug is caused by not undoing "all of the bindings" in unification before verify_attributes/3 I made this change local and the bug clears up.   However other issues can be present afterwards dealing with occurs and cycle checking which require yet another fix

logicmoo:
Proof of bug:   that was in 7.3.15

````
Welcome to SWI-Prolog (Multi-threaded, 64 bits, Version 7.3.15-32-g1336b66)
Copyright (c) 1990-2015 University of Amsterdam, VU Amsterdam
SWI-Prolog comes with ABSOLUTELY NO WARRANTY. This is free software,
and you are welcome to redistribute it under certain conditions.
Please visit http://www.swi-prolog.org for details.

For help, use ?- help(Topic). or ?- apropos(Word).

1 ?- asserta((va_test:verify_attributes(Var,_,[]):-get_attr(Var,va_test,Val),must_be(var,Val))).
true.

2 ?- put_attr(X,va_test,Y), t(Y,X)=t(1,1).
ERROR: Uninstantiated argument expected, found 1
3 ?-
````
Y should not yet be 1 !
since we were supposedly still in the "pre-unification" stages of `verify_attributes/3`

logicmoo:
Sorry one last thing:

See, if you belive Y=1 then you believe unification has taken place.. Thus it be impossible to believe that X was a still a variable. This helps explain the quantum effects that when the debugger has. And further explains why Markus' `verify_attributes/3` code only worked in December and no longer in January when these sorts of changes got made.

PR #115, opened 2016-02-01
Fix documentation of first_solution/3.

edechter:
A tiny fix to the documentation. I don't really understand whether the submodule commit hash changes are supposed to be there. If they are not I'm not sure how to get rid of them.


JanWielemaker:
Decided to edit in my version from the diffs.  I think the preferred way to do these things is to keep or own master in sync with mine.  Then if you want to suggest a change, create a branch, commit into the branch and push to github.  Then make a pull request for the branch.   That way I keep a clean history.

PR #127, opened 2016-02-16
Add missing closing bracket in documentation of read_term/2, option quasi_quotations

fnogatz:


fnogatz:
Don't want to annoy you with small documentation fixes, @JanWielemaker. Let me know if I should gather them first and send a PR every four weeks or so. I only fear to forget about them after some days...

JanWielemaker:
With github it is not a lot of work to review and confirm them.  Do as it suits you best.  Thanks.

PR #133, opened 2016-03-09
Fix documentation of between(L, U) in error.pl

edechter:


JanWielemaker:
I'm afraid this isn't right.  Changed to:

```
%	| between(IntL,IntU) | Integer [IntL..IntU] |
%	| between(FloatL,FloatU) | Number [FloatL..FloatU] |
```

PR #134, opened 2016-03-10
FIXED: make --without-* work

honnix:
Commit 8257acd doesn't work any more with current build structure
because packages are configured during make and hence `DISABLE_PKGS`
is not visible to `packages/configure` script.

The fix is simply passing `--without-*` to `packages/configure`
without touching at the top level.


JanWielemaker:
Thanks.  Merged after rebase and small rewording of the commit message.   I normally build using `build[.templ]`, which doesn't suffer from this.  Still seems to work.

PR #138, opened 2016-03-27
MODIFIED: read_pending_[chars|codes]/3 -> get_pending_[chars|codes]/3.

triska:
The rational is that "read" is always used for full Prolog terms, while
"get" is used for other I/O, like get_char/1 and get_code/1.

Please consider merging this pull request. If you apply it, then `read_pending_codes/3` needs to be replaced by `get_pending_codes/3` in the modules. I have sent pull requests for the packages too.

### DIFF
--- a/library/backcomp.pl
+++ b/library/backcomp.pl
@@ -253,10 +253,10 @@ read_variables(Stream, Term, Vars) :-
 
 %%	read_pending_input(+Stream, -Codes, ?Tail) is det.
 %
-%	@deprecated Use read_pending_codes/3.
+%	@deprecated Use get_pending_codes/3.
 
 read_pending_input(Stream, Codes, Tail) :-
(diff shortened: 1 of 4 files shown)


JanWielemaker:
I'll hold this.  get* is used for single chars/codes.  read* for longer sequences.  This seems consistent with many other languages.  We also have read_line_to_codes/2, read_string/3, etc.   As far as I'm concerned, read_pending_{input,chars,codes} is mostly intemal.  I see little reason to change the name, but if other Prolog systems wish to implement this I'm happy to synchronise the name and semantics with the majority.  Isolated, this feels like a random choice between two valid alternatives.

JanWielemaker:
Closing as part of general cleanup of issues/PRs.   See comments above why this wasn't applied.

PR #144, opened 2016-05-12
Initialisation function from memory 

olsky:
Dear team,

I'd like to checkin the 

`PL_initialise_mem(int argc, char **argv, char *mem_data, long int mem_size)`

 message, which will initialize swipl with the savedsate in memory.

Could you please review and add it to master and next release?

Thank you.
Cheers
Oleh

### DIFF
--- a/src/SWI-Prolog.h
+++ b/src/SWI-Prolog.h
@@ -880,7 +880,7 @@ PL_EXPORT(int)	PL_wchars_to_term(const pl_wchar_t *chars,
 		 *	    EMBEDDING		*
 		 *******************************/
 
-PL_EXPORT(int)		PL_initialise(int argc, char **argv);
+PL_EXPORT(int)		PL_is_initialised(int *argc, char ***argv);
 PL_EXPORT(int)		PL_initialise(int argc, char **argv);
 PL_EXPORT(int)		PL_initialise_mem(int argc, char **argv, char *mem_data, long int mem_size);
 #ifdef __CYGWIN__
--- a/src/SWI-Prolog.h
+++ b/src/SWI-Prolog.h
@@ -882,7 +882,7 @@ PL_EXPORT(int)	PL_wchars_to_term(const pl_wchar_t *chars,
 
 PL_EXPORT(int)		PL_is_initialised(int *argc, char ***argv);
 PL_EXPORT(int)		PL_initialise(int argc, char **argv);
-PL_EXPORT(int)		PL_boot_from_memory(int argc, char **argv, char *mem_data, long int mem_size);
+PL_EXPORT(int)		PL_boot_from_memory(int argc, char **argv, char *mem_data, size_t mem_size);
 #ifdef __CYGWIN__
 PL_EXPORT(void)		PL_install_readline(void);
 #else
--- a/src/SWI-Prolog.h
+++ b/src/SWI-Prolog.h
@@ -881,7 +881,8 @@ PL_EXPORT(int)	PL_wchars_to_term(const pl_wchar_t *chars,
 		 *******************************/
 
 PL_EXPORT(int)		PL_initialise(int argc, char **argv);
-PL_EXPORT(int)		PL_is_initialised(int *argc, char ***argv);
+PL_EXPORT(int)		PL_initialise(int argc, char **argv);
+PL_EXPORT(int)		PL_initialise_mem(int argc, char **argv, char *mem_data, long int mem_size);
 #ifdef __CYGWIN__
 PL_EXPORT(void)		PL_install_readline(void);
 #else
--- a/src/SWI-Prolog.h
+++ b/src/SWI-Prolog.h
@@ -882,7 +882,7 @@ PL_EXPORT(int)	PL_wchars_to_term(const pl_wchar_t *chars,
 
 PL_EXPORT(int)		PL_is_initialised(int *argc, char ***argv);
 PL_EXPORT(int)		PL_initialise(int argc, char **argv);
-PL_EXPORT(int)		PL_initialise_mem(int argc, char **argv, char *mem_data, long int mem_size);
+PL_EXPORT(int)		PL_boot_from_memory(int argc, char **argv, char *mem_data, long int mem_size);
 #ifdef __CYGWIN__
 PL_EXPORT(void)		PL_install_readline(void);
 #else
--- a/src/pl-init.c
+++ b/src/pl-init.c
@@ -917,7 +917,7 @@ PL_initialise_int(int argc, char **argv, char *mem_data, long int mem_size)
 
 // init prolog from memory
 int
-PL_boot_from_memory(int argc, char **argv, char *mem_data, long int mem_size)
+PL_boot_from_memory(int argc, char **argv, char *mem_data, size_t mem_size)
 {
 
 
--- a/src/pl-init.c
+++ b/src/pl-init.c
@@ -784,41 +784,18 @@ PL_initialise_int(int argc, char **argv, char *mem_data, long int mem_size)
   setPrologFlagMask(PLFLAG_SIGNALS);	/* default: handle signals */
 #endif
 
-/*********************************************************************************
-  if (    (GD->resourceDB = rc_open_archive(GD->paths.executable, RC_RDONLY))
+  // if resourceDB is present -- skip  rc_open_archive call (case for PL_boot_from_xxx calls)
+  if ( !GD->resourceDB) {
+    if (    (GD->resourceDB = rc_open_archive(GD->paths.executable, RC_RDONLY))
 #ifdef __WINDOWS__
-       || (GD->resourceDB = rc_open_archive(GD->paths.module, RC_RDONLY))
+         || (GD->resourceDB = rc_open_archive(GD->paths.module, RC_RDONLY))
 #endif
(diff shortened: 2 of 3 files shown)


src/SWI-Prolog.h
```diff
@@ -881,7 +881,8 @@ PL_EXPORT(int)	PL_wchars_to_term(const pl_wchar_t *chars,
 		 *******************************/
 
 PL_EXPORT(int)		PL_initialise(int argc, char **argv);
-PL_EXPORT(int)		PL_is_initialised(int *argc, char ***argv);
```
JanWielemaker: I guess removing this prototype is a typo?

JanWielemaker:
I'm happy with the functionality.  I have some remarks though

  - Please use `size_t` rather than `long int` for portability and consistency.
  - I wonder whether it is not a better idea to add a function to be called __before__ `PL_initialise()`
    that sets the location and format of the state.   This can write in additional fields in GD (see `pl-global.h`).
    That avoids ending up with more and more copies of PL_initialise().  Another sensible way to
    fetch the resources is from a `jar` file, Windows resource, etc.  I'd says something like

    - PL_boot_from_memory(char *data, size_t len)
    - PL_boot_from_file(char *file)
    - ...

If the number of formats get out of hand we can install callback functions.  Actually, I do have
extensions in mind such as a format that materializes the state lazily to reduce startup time.

Does that make sense?

RobertoBagnara:
This is a great feature: I have wanted it for years!

olsky:
Hi @JanWielemaker 

yes, it makes sense! Sounds good, let me rework it a bit.

Thank you & Cheers
Oleh

src/SWI-Prolog.h
olsky: oops... strange... yes, it was not intended

RobertoBagnara:
Please let me know if I can help completing the implementation of this feature.

olsky:
Hi Roberto, 

thank you for your reminder. I'm (unfortunately) not an advanced github user... I'm stuck with pull and revert :)

Here is what I'm planning to commit:

Changes to pl-init.c:

FROM:
```

  if (    (GD->resourceDB = rc_open_archive(GD->paths.executable, RC_RDONLY))
#ifdef __WINDOWS__
       || (GD->resourceDB = rc_open_archive(GD->paths.module, RC_RDONLY))
#endif
     )
  { rcpath = ((RcArchive)GD->resourceDB)->path;
    initDefaultOptions();
  }

```


TO:
```
  // if resourceDB is present -- skip  rc_open_archive call (case for PL_boot_from_xxx calls)
  if ( !GD->resourceDB) {
	  if (    (GD->resourceDB = rc_open_archive(GD->paths.executable, RC_RDONLY))
#ifdef __WINDOWS__
	       || (GD->resourceDB = rc_open_archive(GD->paths.module, RC_RDONLY))
#endif
	     )
	  { rcpath = ((RcArchive)GD->resourceDB)->path;
	    initDefaultOptions();
	  }
  }
```

this extra  if ( !GD->resourceDB) check would be transparent for other PL_boot_from_xxx functions.

and then simplify init from memory with this:

```
// init prolog from memory
int
PL_boot_from_memory(int argc, char **argv, char *mem_data, long int mem_size)
{
	if (mem_size != 0 && mem_data != NULL) {
		if (    (GD->resourceDB = rc_open_archive_mem(GD->paths.executable, mem_data, mem_size, RC_RDONLY))
#ifdef __WINDOWS__
		        || (GD->resourceDB = rc_open_archive_mem(GD->paths.module, mem_data, mem_size, RC_RDONLY))
#endif
		   )
		{	
			return PL_initialise(argc, argv);
		}
	}
	return FALSE;
}
```

and the rc_open_archive_mem to remain in rc/access.c

What do you think?

Also, I would suggest to include the unzip-function (zlib?) to the prolog co, so we could boot from zipped files and jars... but it's probably not the right thread for this...

RobertoBagnara:
On 06/23/2016 01:02 PM, olsky wrote:
> Hi Roberto,
> 
> thank you for your reminder. I'm (unfortunately) not an advanced github user... I'm stuck with pull and revert :)
> 
> Here is what I'm planning to commit:
> 
> Changes to pl-init.c:
> 
> FROM:
> 
> |if ( (GD->resourceDB = rc_open_archive(GD->paths.executable, RC_RDONLY)) #ifdef __WINDOWS__ || (GD->resourceDB = rc_open_archive(GD->paths.module, RC_RDONLY)) #endif ) { rcpath = ((RcArchive)GD->resourceDB)->path; initDefaultOptions(); } |
> 
> TO:
> 
> |// if resourceDB is present -- skip rc_open_archive call (case for PL_boot_from_xxx calls) if ( !GD->resourceDB) { if ( (GD->resourceDB = rc_open_archive(GD->paths.executable, RC_RDONLY)) #ifdef __WINDOWS__ || (GD->resourceDB = rc_open_archive(GD->paths.module, RC_RDONLY)) #endif ) { rcpath = ((RcArchive)GD->resourceDB)->path; initDefaultOptions(); } } |
> 
> this extra if ( !GD->resourceDB) check would be transparent for other PL_boot_from_xxx functions.
> 
> and then simplify init from memory with this:
> 
> |// init prolog from memory int PL_boot_from_memory(int argc, char **argv, char *mem_data, long int mem_size) { if (mem_size != 0 && mem_data != NULL) { if ( (GD->resourceDB = rc_open_archive_mem(GD->paths.executable, mem_data, mem_size, RC_RDONLY)) #ifdef __WINDOWS__ || (GD->resourceDB = rc_open_archive_mem(GD->paths.module, mem_data, mem_size, RC_RDONLY)) #endif ) { return PL_initialise(argc, argv); } } return FALSE; } |
> 
> and the rc_open_archive_mem to remain in rc/access.c
> 
> What do you think?

Hi Oleh,

this looks good to me (and consistent with what Jan has suggested, but
we should better ask him).

> Also, I would suggest to include the unzip-function (zlib?) to the prolog co, so we could boot from zipped files and jars... but it's probably not the right thread for this...

Indeed, my first use for this new feature will be to boot
from scrambled and compressed files: I will read the file,
unscramble and uncompress it in memory, then I will call
PL_boot_from_memory() and PL_initialise().
Saved states are highly compressible: I see a factor 6
here using xz -9.
Kind regards,

   Roberto

-- 
     Prof. Roberto Bagnara

Applied Formal Methods Laboratory - University of Parma, Italy
mailto:bagnara@cs.unipr.it
                              BUGSENG srl - http://bugseng.com
                              mailto:roberto.bagnara@bugseng.com

JanWielemaker:
Looks promising!  Plan to have a detailed look tomorrow.  Support for booting from a zip file would be nice.  That would also allow for booting from an encrypted state.  Not sure how useful that is.  One of the issues is that zip encryption is claimed to be pretty weak.  On the other hand, once you have a Prolog state, it is hard to really prevent people from being able to run it without being able to get a listing of its content.

olsky:
Hi Roberto, 

I'm having somewhat similar use-case as yours, and zipping saved state is of at least 50% space save. In my setup, I load zip file in memory, inflate it and pass to prolog, that's why I needed init_from_mem.

Somehow I got lazy to patch it always thus asked to include it in the release :)

olsky:
Hi Jan, 

your idea of having encryption is good: be it zip-based or even PKI... why not? +1! 

Another view on this would be: if we have embedded prolog (other host system), with its own top_level then there is no concern with the listing. Dont you think?

In this case, the host system should take care of storing keys and isolating prolog from external world.

JanWielemaker:
> your idea of having encryption is good: be it zip-based or even PKI... why not? +1!
> 
> Another view on this would be: if we have embedded prolog (other host system), with its own top_level then there is no concern with the listing. Dont you think?
> 
> In this case, the host system should take care of storing keys and isolating prolog from external world.

If the embedded system can somehow call the ZIP encryption API without leaking the key and does not allow the user to access arbitrary predicates it is a good protection.  Especially the first is a big *if* though.  It is really hard to hide a key, especially if you need to give it to a known API.  Even after stripping symbols a bit experienced hacker can easily find the ZIP API and instruct a debugger to stop there.

JanWielemaker:
I've changed it all a little.  You find my current version on the branch `memory-boot`.  If you agree with this, I'll squash the commits, leaving you as author.   Can you send me a mail where you state that you agree that your code will be re-licensed under the Simplified BSD license?  All but one contributor has already done so.  We are working on replacing his code.   I don't want that one more time ...

olsky:
Super! thank you! :)

I'll send you an email as you suggested.

When is the next stable version scheduled? :)

JanWielemaker:
> When is the next stable version scheduled? :)

Is was more or less planned, but the extra work dealing with the license change and the new stuff wrt. tabling and upcoming Lean Prolog style _interactors_ delay it.  Anyway, 7.3.x is pretty stable, so just
pick a version thereof.

JanWielemaker:
The modified version has been squashed to a single commit and is now in master.  Thanks.

This provides a way around to use zip archives for booting, but it is of course not the same.  I'm a bit in doubt.  The current resource db handling is way simpler and more lightweight and doesn't involve any dependencies.  Using `libz` provides better interoperability, compression and encryption.   Please open a new topic if you want to discus how to continue on this.

PR #152, opened 2016-07-09
BUILD: include ltx2htm in packages

triska:
The instructions in `man/README` assume that this package exists.


JanWielemaker:
No, we don't want that.  ltx2htm should be considered specifically for generating the documentation HTML.  We don't want to bother the normal Prolog user with that.   Simply run

    git submodule update --init packages/ltx2htm

and add `ltx2htm` to `EXTRA_PKGS` in `build`.  You only have to do that once.

triska:
The instructions in `man/README` assume that this package exists. Please at least adjust the instructions so that users have a chance to build the documentation.

JanWielemaker:
Done (and turned into Markdown, so it shows up more readable on Github).

PR #164, opened 2016-09-12
Fixed missing cuts after has_type/2

jp-diegidio:
Since has_type/2 is semidet but also multifile, the code is not safe at the moment in a couple of places, i.e. where has_type is called without a cut following.  (A cut is present in the other two places where has_type is called.)  I have fixed this by adding the cuts.


JanWielemaker:
Merged after editing commit message to suit auto-generation of change logs.

PR #181, opened 2016-12-04
DEBUG: Give the programmer information about how to resolve the issue…

wouterbeek:
… of non-directive settings.


JanWielemaker:
There are quite a few similar issues in the library.  We need something more generic.   It is definitely a good idea to have a better error though.

PR #189, opened 2016-12-29
user-data-type-attvars

logicmoo:
Hi @JanWielemaker , when you have time to look this over.

This adds a new predicate a new datatype that pretends to be an attvar.

How much performance impact would this have on master?  When I tested I could not see a difference.

This tiny bit of machinery would go a long way and allow several extends to be developed as add-on packs.

And eventually including a verify_attributes/3 extension pack.

### DIFF
--- a/src/pl-attvar.c
+++ b/src/pl-attvar.c
@@ -183,6 +183,19 @@ assignAttVar(Word av, Word value ARG_LD)
   registerWakeup(a, value PASS_LD);
 
   TrailAssignment(av);
+
+  /* When first attribute is $VAR$ skip binding (allows to be done elseware) */
+  { Word l = a;
+    deRef(l);
+    if ( isTerm(*l) )
+     { Functor f = valueTerm(*l);
+      if ( f->definition == FUNCTOR_att3 )
+      { Word n; 
+	    deRef2(&f->arguments[0], n);
+	    if ( *n == ATOM_dvard ) return;
+    }}}
+  
+
   if ( isAttVar(*value) )
   { DEBUG(1, Sdprintf("Unifying two attvars\n"));
     *av = makeRef(value);
@@ -1378,6 +1391,33 @@ PRED_IMPL("$call_residue_vars_end", 0, call_residue_vars_end, 0)
 #endif /*O_CALL_RESIDUE*/
 
 
+/* - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+ attv_bind_trail(+AttVar, +Value, +NilOrTrail) is det.  
+    Binds AttVar with Value without calling wakeup
+ - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - */
+static
+PRED_IMPL("attv_bind_trail", 3, attv_bind_trail, 0)
+{ PRED_LD
+  Word av;
+  
+  deRef2(valTermRef(A1), av);
+  if (!isAttVar(*av) )
+  { return PL_error("attvar_bind_trail", 3, NULL, ERR_UNINSTANTIATION, 1, A1);
+  }
+  
+  /* Maybe Not needed */ 
+  if ( !hasGlobalSpace(0) )
+  { int rc;
+    if ( (rc=ensureGlobalSpace(0, ALLOW_GC)) != TRUE )
+      return raiseStackOverflow(rc);
+  }
+
+  /* 3rd arg is for a little experiment */
+  if(!PL_get_nil(A3)) TrailAssignment(av);
+  *av = linkVal(valTermRef(A2));
+  return TRUE;
+}
+
 		 /*******************************
 		 *	    REGISTRATION	*
 		 *******************************/
@@ -1398,6 +1438,7 @@ BeginPredDefs(attvar)
   PRED_DEF("$call_residue_vars_start", 0, call_residue_vars_start, 0)
   PRED_DEF("$call_residue_vars_end", 0, call_residue_vars_end, 0)
 #endif
+  PRED_DEF("attv_bind_trail", 3, attv_bind_trail, 0)
 EndPredDefs
 
 #endif /*O_ATTVAR*/


DouglasRMiles:
Hi Jan,

If the patch isn't acceptable.   That is OK..  my motivation is I need to
do some Windows work on a deployment leveraging that patch. that  (I
haven't had success building SWI on windows since whenever Makefile.mak was
dropped (probably over a decade ago?) )

So what I'd need is perhaps a build log of it getting built on windows to
make sure I understand what I might be doing wrong.
And find out hat specific GMP is used on MinGW.

Thanks in advance!





On Thu, Dec 29, 2016 at 11:01 AM, Douglas R. Miles <notifications@github.com
> wrote:

> Hi @JanWielemaker <https://github.com/JanWielemaker> , when you have time
> to look this over.
>
> This adds a new predicate a new datatype that pretends to be an attvar.
>
> How much performance impact would this have on master? When I tested I
> could not see a difference.
>
​Anything I'd do to mitigate (like a flag) would cost more than just the
derefing done there. ​




> This tiny bit of machinery would go a long way and allow several extends
> to be developed as add-on packs.
>
> And eventually including a verify_attributes/3 extension pack.
>
​Though "real attvars" would need to respect unifiable/3​ in which as of
right now I only use these new terms as objects.

JanWielemaker:
Hi Douglas,

I'm sorry to say I haven't looked at it yet.  I'm mostly on holiday and right now overwhelmed in some low level stream handling :(  The size of the patch sounds promising!

     Sorry --- Jan

The only viable way to make the Windows version right now is to cross-compile it from Linux.  README.mingw is a mostly machine readable specification what needs to be done.  I use it on Ubuntu, so if you setup a VM for it, Ubuntu is the safest option.  It probably works with most Linux distros though.

JanWielemaker:
Hi Douglas,

As it is a rainy new year's day I had a look :)  My impression is that this is too specific to be placed in the standard distribution.  Without proper user-level tools this is a mere hack that is there, if I understand you correctly, to provide you with a binary for Windows to do your things.  I think that is easier solved by setting up the environment for building the Windows version.  Considering the scope of the patch it should be trivial for you to maintain your own version.

The other part of the story is that I still do not believe in the direction of the verify_attributes hook.   I had a lot of exchanges with Joachim Schimpf from ECLiPSe about synchronising the string support between SWI and ECLiPSe shortly after the failed attempt to add verify_attributes.  He said he followed some of this discussion and in his opinion this is the wrong direction.  As ECLiPSe is one of the most advanced Prolog systems when it comes to constraint handling we should be serious about his opinion on these matters.   I have not discussed the details with Joachim, but before taking any step modifying the attributed variable interface we (including me) should carefully study ECLiPSe and discuss this with Joachim.

P.s.  `attv_bind_trail/3` is broken.  You pick a Word type pointer and then optionally do a shift/GC.  That will invalidate `av`.  You need to do the space check first.

logicmoo:
Hi Jan.

On Sun, Jan 1, 2017 at 7:02 AM, Jan Wielemaker <notifications@github.com>
wrote:

Hi Douglas,
>
> As it is a rainy new year's day I had a look :) My impression is that this
> is too specific to be placed in the standard distribution. Without proper
> user-level tools this is a mere hack that is there, if I understand you
> correctly, to provide you with a binary for Windows to do your things. I
> think that is easier solved by setting up the environment for building the
> Windows version. Considering the scope of the patch it should be trivial
> for you to maintain your own version.
>
​OK Makes sense​.. will do.



> The other part of the story is that I still do not believe in the
> direction of the verify_attributes hook. I had a lot of exchanges with
> Joachim Schimpf from ECLiPSe about synchronising the string support between
> SWI and ECLiPSe shortly after the failed attempt to add verify_attributes.
> He said he followed some of this discussion and in his opinion this is the
> wrong direction.
>

I really owe it to everyone to clarify. ​I was not actually interested in
verify_attributes/3 ​that was only for a gift to those whom wanted it. As
My goal was merely to win enough favor I could work on a new foreign
datatype based on the ECLiPSe interface.    My constraint handler code at
work only uses attvars so i can avoid marshaling memory back an forth to
prolog  which calls C++ >Unify()
.  C++ >Unify()  uses a whole different constraint handling system that has
nothing to do with attributed vars.


Correct me  if I am wrong, but if in ECLiPSe  if I wanted a new String
datatype I create it as an ATTVAR.


Since the attvar allows the override of =/2 , copy_term/2, =@=/2  and ==/2 .
  I can implement all the new datatypes I wanted:  lisp symbols, python
streams, prolog dictionaries, C# dictionaries, XPCE refs etc,  Strings,
 and and so forth.

Being a [att]VAR allows them to mingle with all prolog datatypes.




In other words if BLOB offered to unify with anything as

   AND let me control the impl of   =/2 , copy_term/2, =@=/2, ==/2.

I'd have no need for attvars in any of my uses.


As ECLiPSe is one of the most advanced Prolog systems when it comes to
> constraint handling we should be serious about his opinion on these
> matters. I have not discussed the details with Joachim, but before taking
> any step modifying the attributed variable interface we (including me)
> should carefully study ECLiPSe and discuss this with Joachim
> ​​
>
> P.s.
> ​
>
> attv_bind_trail/3
>
> ​
>  is broken. You pick a Word type pointer and then optionally do a
> shift/GC. That will invalidate av. You need to do the space check first.
>
> ​Thanks! fixed locally now​.

I never noticed becasue I'll never need to attv_bind_trail/3 myself since I
never allow them to be replaced by prolog data as they are "Objects" in the
sense of XPCE or BLOBs .

logicmoo:
>
>
>
>
>
> In other words if BLOB offered to unify with anything as
>
>    AND let me control the impl of   =/2 , copy_term/2, =@=/2, ==/2.
>
> I'd have no need for attvars in any of my uses.
>
>


Actually the way you designed BLOBs has such an intent with with it's
Handlers Struct.

​typedef struct PL_blob_t

{ uintptr_t     magic;          /* PL_BLOB_MAGIC */
  uintptr_t     flags;          /* Bitwise or of PL_BLOB_* */
  char *        name;           /* name of the type */
  int           (*release)(atom_t a);
  int           (*compare)(atom_t a, atom_t b);
  int           (*write)(IOSTREAM *s, atom_t a, int flags);
  void          (*acquire)(atom_t a);
  ...

} PL_blob_t;​


​
ECLiPSe
​​
​ATTVAR is very similar it allows the attributes to control its handlers to
 ​
Handler DeclarationsThe predefined operations and the corresponding handler
arguments are the following:unify

Operation : unification

Handler : handler(+Term, ?Attribute [, ?SuspendAttribute])

Description : The handler for the usual unification. Term is the term that
was unified with the attributed variable, it is either a nonvariable or an
attributed variable. Attribute is the contents of the attribute slot
corresponding to the extension. Note that, at this point in execution, the
original attributed variable no longer exists, because it has already been
bound to Term. The optional third argument is the suspend-attribute of the
former variable; it may be needed to wake the variable's 'constrained'
suspension list.

The handler's job is to determine whether the binding is allowed with
respect to the attribute. This could for example involve checking whether
the bound term is in a domain described by the attribute. For
variable-variable bindings, typically the remaining attribute must be
updated to reflect the intersection of the two individual attributes. In
case of success, suspension lists inside the attributes may need to be
scheduled for waking.

If an attributed variable is unified with a standard variable, the variable
is bound to the attributed variable and no handlers are invoked. If an
attributed variable is unified with another attributed variable or a
non-variable, the attributed variable is bound (like a standard variable)
to the other term and all handlers for the unify operation are invoked.
Note that several attributed variable bindings can occur e.g. during a head
unification and also during a single unification of compound terms. The
handlers are only invoked at certain trigger points (usually before the
next regular predicate call). Woken goals will start executing once all
unify-handlers are done.

test_unify

Operation : unification test

Handler : handler(+Term, ?Attribute)

Description : The handler for a unifiability test which is not supposed to
trigger constraints propagation. It is used e.g. in the
​

not_unify
<http://eclipseclp.org/docs/Dev/current/bips/kernel/termcomp/not_unify-2.html>/2
predicate. The handler arguments are equivalent to those of the unification
handler, Term is the term that was unified with the attributed variable,
Attribute is the attribute of this extension. The handler's job is to
determine whether Attribute allows unification with Term (not considering
effects of woken goals). During the execution of the handler the attributed
variable may be bound to Term, however when all local handlers succeed, all
bindings are undone, and no waking occurs.

compare_instances

Operation : instance and variant tests

Handler : handler(-Res, ?TermL, ?TermR)

Description : The handler for the
​
​
variant
<http://eclipseclp.org/docs/Dev/current/bips/kernel/termcomp/variant-2.html>
​​
/2,
​
instance
<http://eclipseclp.org/docs/Dev/current/bips/kernel/termcomp/instance-2.html>/2
and
​
compare_instances
<http://eclipseclp.org/docs/Dev/current/bips/kernel/termcomp/compare_instances-3.html>/3
instance-testing predicates. The handler arguments are similar to those of
compare_instances/3. At least one of TermL or TermR will be an attributed
variable whenever the handler is invoked. The handler should bind Res to <
if the attributes imply that TermL is a proper instance of TermR, > if
TermR is a proper instance of TermL, and = if the two attributed variables
are variants of each other (e.g. they have identical domains). If the terms
are incomparable (not unifiable), the handler must fail. If the attribute
being declared has no bearing on the instance-relationship, this handler
should remain undefined.

copy_term

Operation : copying an attributed variable

Handler : handler(?AttrVar, ?Var)

Description : The handler for the
​
copy_term
<http://eclipseclp.org/docs/Dev/current/bips/kernel/termmanip/copy_term-2.html>/2
predicate. AttrVar is the attributed variable encountered in the copied
term, Var is its corresponding variable in the copy. All extension handlers
receive the same arguments. This means that if the attributed variable
should be copied as an attributed variable, the handler must check if Var
is still a free variable or if it was already bound to an attributed
variable by a previous handler.

​
Just like BLOBs,   ECLiPSe  Att
​V​
ars can be used to implement all sort
​s​
of new datatypes.    (The difference is BLOBS dont try to unify with term
​s
​, atoms, numbers, streams .. etc)​


​

PR #199, opened 2017-02-06
Fix rbtrees:rb_in/3 and remove rb_lookupall/3

samer--:
These changes are aimed at clarifying the status of variables in keys and duplicate keys. After discussions on the mailing list, we established that is is ok for keys to contain variables, in which case, key matching is based on (==)/2 equivalence of terms (and not on unifiability, subsumption or structural equality).
    
It also seems to be the case that a tree with more than one entry for a given key is an invalid tree (and is rejected by is_rbtree/1).
    
Taking these together, we can conclude that the old rb_lookupall/3 was redundant, since there should only ever be one value for a given key whether or not that key contains variables. Instead, rb_lookup/3 should be used. I've taken the possibly drastic step of removing rb_lookupall/3 completely, but another option would be to leave it in  but deprecated and with dire warnings.
    
Also, the old rb_in/3 was diverting to lookup/3 for nonvariable keys. This was giving incorrect behaviour for partially instantiated keys, since the tree may contain several keys that unify with the given term. My initial fix was to replace the `nonvar(Key)` with a `\+ground(Key)`. However, further consideration has lead me to remove that clause of rb_in/3 entirely. The reasoning behind this is:
       (a) if you want fast lookup, use rb_lookup/3;
       (b) making rb_in/3 check the key argument with ground/1 forces it to examine the whole term and removing that clause makes rb_in/3 cleaner;
       (c) the rest of the rbtrees interface is relies on (==)/2 equality between keys, where as the nondeterministic mode of rb_in/3 relies on unification of keys. This is fine, but mixing both meanings in one predicate depending on the instantiation state of the key feels not quite right.

In further support of this change, this is exactly what assoc:gen_assoc/3 does, leaving get_assoc/3 the job of doing O(log N) lookups. Also, assoc:ord_list_to_assoc/2 prevents the creation of duplicate keys.


JanWielemaker:
Thanks.  I agree.  Did a few more documentation edits.  Nothing serious.  Looking through this I wonder whether we want to keep rb_visit/3 and rb_keys/3.  They are basically difference list versions, but disregarding the argument order conventions.

Another point I'm wondering about is whether or not this should be in 7.4.0.  This cleans up the library, but is potentially breaking compatibility.

samer--:
Thanks for including this Jan. I think I agree about rb_visit/3 and rb_keys/3 - first of all, I don't know how often one would want to append another list to the list of pairs or keys from an rbtree, and so why bother with the difference list versions, and also, they are both easily handled by rb_fold/4 (though at the expense of a metacall for each element).

Re. the possibility of this being a breaking change - I'm in the dark as I don't know how people are currently using this library. My initial post didn't get much of a response so maybe it's ok!

PR #210, opened 2017-03-17
ENHANCED: Allow reproducible builds by using the SOURCE_DATE_EPOCH variable

susnux:
To allow [reproducible builds](https://reproducible-builds.org/) (same source results in same binary) replace
`__DATE__` and `__TIME__` with system provided values.

This is used e.g. by [debian](https://wiki.debian.org/ReproducibleBuilds), openSUSE also recommends to patch usage of `__DATE__` and `__TIME__` as leads to a lot of rebuilds even when source did not change, for more information: https://reproducible-builds.org/specs/source-date-epoch/

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -166,6 +166,15 @@ COMMON= README
 UNICODE=blocks.pl unicode_data.pl
 builddirs=../lib ../lib/$(PLARCH) os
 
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.  Merged.

PR #230, opened 2017-05-16
ADDED: Prolog flag `functional_notation` to disable functional notation.

### LINKED ISSUE #229: Add a prolog flag controlling the expansion of functional notation
Adding a flag would allow me to disable the functional notation on a per-module basis so that I can reuse ./2 for my own purposes.

webstrand:
Resolves #229

### DIFF
--- a/boot/expand.pl
+++ b/boot/expand.pl
@@ -915,6 +915,7 @@ expand_functions(G0, P0, G, P, M, MList, Term) :-
 %   @tbd: make functions module-local
 
 expand_functional_notation(G0, P0, G, P, M, _MList, _Term) :-
+    current_prolog_flag(M:functional_notation, true),
     contains_functions(G0),
     replace_functions(G0, P0, Eval, EvalPos, G1, G1Pos, M),
     Eval \== true,
--- a/man/overview.doc
+++ b/man/overview.doc
@@ -1347,6 +1347,10 @@ search result satisfies the conditions. Default is 10 seconds, which
 typically avoids most repetitive searches for (library) files during
 compilation. Setting this value to 0 (zero) disables the cache.
 
+    \prologflagitem{functional_notation}{bool}{rw}
+If \const{true} (default) the functional notation ./2 is expanded. If false,
+./2 can be used for other purposes. This flag must be set on a per-module basis.
+
     \prologflagitem{gc}{bool}{rw}
 If true (default), the garbage collector is active.  If false, neither
 garbage collection, nor stack shifts will take place, even not on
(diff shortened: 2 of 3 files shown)


JanWielemaker:
Flags are not module-specific but thread-specific.  Some flags are hacked internally and re-routed to module properties.   This started to be as much as possible in line with some ISO standards and has been used since then for other properties.  To implement this properly you'd have to got the same route as flags such as `double_quotes`.

That said, also if properly implemented, I'm not sure I'd accept such a flag.  `./2` could never be used sensibly as an operator as it was ambiguous due to the list notation as well as the floating point notation.  I.e., in the old days people tended to turn `.` into an infix operator and write a list as `a.b.c.[]` rather than `[a,c,b]`.  However, `1.2.3.[]` would read as `[1.2,3]` (two elements).  That is hard to explain.

So, I claim that the `.` cannot meaningfully be used for anything but constructing a list in traditional Prolog due to list and float issues (which means it is not useful at all as we have list notation for that
purpose)  and it cannot be used meaningfully for anything but functional notation in SWI-Prolog 7.
Adding a flag merely promotes something that doesn't work properly anyway and most likely will cause confusion on its own.

In extreme cases `term_expansion/2` can be used to avoid functional expansion.

So, what is the real problem you need to solve?

webstrand:
I'm writing a build system that uses predicates for it's production rules. Since, by and large, most filenames only contain [a-z_], I had hoped to represent paths in my build scripts naturally: `usr/lib/libswipl.so.7.5` would be `/(/(usr,lib),.(libswipl,.(so,.(7,5))))` This way, the filenames would be easy read and unification could still be leveraged for most common file matching operations. 

Unfortunately, I hadn't thought about how this notation would interact with floating point. I can't see any way to disambiguate between floating point and ./2 without checking the types of the two adjacent terms. That would be strangely inconsistent with the rest of the language and fairly confusing.

Therefore, I agree with you that the idea behind this patch is flawed and will be dropping it, unless you have a better idea. Thank you for taking the time to explain.

JanWielemaker:
Dealing with / is indeed doable (that is why SWI-Prolog also allows for `[library(http/http_open)]` without quotes.  `.so.7.5` is something you cannot deal with.  Well, if you really want you can use `read_term/3` for reading the code yourself using the `subterm_positions/1`, get the file positions for the term and read the raw file for the original lexical representation.  Yuk!  I used that trick for programs that manipulate Prolog source and for which we wait to keep e.g., `0'a` instead of translating it to `97`.

Apart from that, I think the options are to tell your users to always quote their file names (possibly using double quotes to create strings) or define your own rule language syntax and read and compile that to Prolog predicates.

PR #232, opened 2017-05-20
Point submodules to public URLs

cbarrick:
This patch changes the submodules from relative to absolute URLs, e.g. `../bench.git` becomes `https://github.com/SWI-Prolog/bench.git`.

As a result, cloning the project is easier and clicking on a submodule in GitHub takes you to the corresponding repository.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,111 +1,111 @@
 [submodule "packages/chr"]
 	path = packages/chr
-	url = ../packages-chr.git
+	url = https://github.com/SWI-Prolog/packages-chr.git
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I don't want that.  I want to checkout using `git@` or `https:` or have a mirror elsewhere, etc.  If you want a github clone of the core and keep sharing the rest, clone the core, make a clone on github and use `git remote set-url origin <your clone>` or, possibly better, `git remote add my-clone <your clone>`, so you can easily access both the origin and your clone and keep things merged.

PR #233, opened 2017-05-30
thread_get_message with timeout can block forever on Windows

thetrime:
dwMilliseconds is a DWORD, which is unsigned. This means the check to… see whether the timeout has already occurred will only succeed if the deadline is precisely now; otherwise we will wait up to 7 weeks for the deadline in the case where it has already passed

Here's a simple test case that reproduces (note that it depends on the speed of your computer)

```
go:-
        message_queue_create(Queue),
        forall(between(0, 1000000, T),
               thread_send_message(Queue, a(T))),
        writeln(reading),
        thread_get_message(Queue, a(9999999), [timeout(0)]).
```


thetrime:
Replaced with #243

Issue #236, opened 2017-06-02
pack_upgrade/1 broken for Github


wouterbeek:
`pack_upgrade/1` currently fails for Packs that are hosten on Github.  This may be due to an API change in Github: the Wildcards that are used to match version-denoting URLs [1] do not seem to match the URLs that Github is currently using to denote versions [2].

```
[1] `releases/*.zip`
[2] `releases/tag/<VERSION>.zip`
```

Possibly related, but does not seem to search for release URLs:

```prolog
?- pack_upgrade(googleclient).
% Querying "https://github.com/JanWielemaker/googleclient/archive/" to find new versions ...
% > GET /JanWielemaker/googleclient/archive/ HTTP/1.1
% > Host: github.com
% > User-Agent: SWI-Prolog
% > Connection: close
ERROR: url `'https://github.com/JanWielemaker/googleclient/archive/'' does not exist (status(404,Not Found))
```

fnogatz:
This is still a problem for calls of the form `pack_install('https://github.com/<owner>/<pack>/archive/<version>.zip')` as mentioned in the documentation. For example:

```prolog
?- pack_install('https://github.com/jonakalkus/xsd/archive/v0.1.2.zip').
ERROR: Version mismatch: pack.pl: 0.1.2, file claims version 0.1.2.2
false.
```

I tracked down the issue to the point that `github_version/4` creates a `VersionURI` which does not end with `.zip` (maybe because GitHub's API does not provide a file extension) and therefore the `available_download_versions/2` returns a wrong download link without the `.zip` at the end. As this URL is used by `file_name_extension/3`, it takes the last digit of the version as the filename extension and raises an error because of mismatching version numbers.

This seems to be a minor bug in @wouterbeek's new approach of using GitHub's API, isn't it?

wouterbeek:
My 'fix' indeed appears to have been incorrect, but the previous situation in which the HTML was scraped for the download link (IIRC then that was actually what was done) also no longer worked( (possibly because Github changed the structure of its HTML).  We need to properly fix this by using the official Github API.

JanWielemaker:
[...]

PR #237, opened 2017-06-04
FIXED: Packs distributed as Github releases could no longer be upgraded.

wouterbeek:
See issue https://github.com/SWI-Prolog/swipl-devel/issues/236 for more
information.

This approach uses the Github API to retrieve the tags i.o. scraping the
HTML.


JanWielemaker:
Merged after some minor fixes.

PR #242, opened 2017-06-15
DOC: Mention crypto_n_random_bytes/2 in the documentation of random/1.

triska:
I think a warning and pointer to the newly available cryptographically secure generation of random numbers is highly warranted at this point. Please merge as you see fit.

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -6731,6 +6731,12 @@ numbers are shared between threads and the seed is initialised from the
 clock when SWI-Prolog was started. The predicate set_random/1 can be
 used to control the random number generator.
 
+\textbf{Warning!} This method does \textit{not} produce
+cryptographically secure random numbers! To generate cryptographically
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I'll check tomorrow.  AFAIK, the docs are wrong.  the GMP random generated is initialized from a cryptographically safe source (if available on the OS).

triska:
That may be the case, but the mentioned algorithm itself (Mersenne&nbsp;Twister) is *not* cryptographically secure. 

Any pseudorandom number sequence generated by a linear recursion is insecure, since from a sufficiently long subsequence of the output, you can predict *all* the rest!

JanWielemaker:
Merge after some more rewrite.   I wasn't really aware of this.  Note that you can still use the
random functions and predicates for crypto purposes, but you should call `set_random(seed(random))`
for each call.

PR #243, opened 2017-06-16
Fix potential infinite timeout in thread_get_message in win32

thetrime:
Note that this is a rewrite, based on Jan's suggestions, of #233, which I will now close.

dwMilliseconds is a DWORD, which is unsigned. This means the check to see whether the timeout has already occurred will only succeed if the deadline is precisely now; otherwise we will wait up to 7 weeks for the deadline in the case where it has already passed


JanWielemaker:
I generalised the stuff a little further and handle the case where a DWORD is too small.  Didn't test it, except for testing that it compiled.

PR #268, opened 2017-09-20
Feature titled console

samer--:
This PR adds `attach_console/1`, which works just like `attach_console/0`, but accepts an arbitrary title instead of building a title from the thread id.


JanWielemaker:
Merged.  Added a second patch to add interactor/1.  

P.s. Could you please make sure that PRs are relative to the main master?  I can now only merge these using some git magic ...

PR #269, opened 2017-09-29
Fix expand term

samer--:
This PR refactors the term expansion system to provide a more consistent and complete processing of expanded terms, while, in the process, fixing a bug that meant that if a term expansion rule produces a DCG clause (`Head --> Body`) as part of a _sequence_ of expanded terms, then the DCG clause is not expanded using `dcg_translate_rule` but instead ends up in the database as a clause of `(-->)/2`. 

The expansion pipeline consists of an initial phase of one-to-many term mapping using the `term_expansion/{2,4}` hooks in several modules, followed by three stages of one-to-one mappings to implement DCG expansions, function expansions in clause heads, body goal expansions and predicate renamings. 

The one-to-many mappings are handled consistently by (a) immediately normalising the results of `term_expansion` to a _list_ of term-layout pairs and (b) using `extend/3` to lift any mapping of type `A -> list(A)` to `list(A) -> list(A)`. (This is equivalent to the `bind` operation of the list monad, which is called `extend` when the arguments are flipped--hence the name.)
The one-to-one mappings are handled using `maplist/3`. Together, this means that all the lower level expansion predicates never have to deal with lists.

The system is based on a type specification for terms (see `boot/notes.txt`) which all transformations respect. DCG expansion, goal expansion, and predicate renaming are all implemented using `map_term/3` and `map_qual/5`, which are basically a strict structural over the term type. This ensures consistency in all term transformations and localises the specification of the term type.

The interface presented to `init.pl` is now `expand_term_to_terms/2`, which uses `expand_term_to_terms/4` and _always_ produces a list of term-layout pairs. The whole
system should be strictly deterministic. The old `expand_term/{2,4}` interface is preserved (not least for `library/prolog_source.pl' to use).

[Additional note]
The last commit introduced the `map_term/3` and `map_qual/4` predicates to factor out the common structural recursion patterns that were used in DCG expansion, body expansion and predicate renaming. If these and the other high-order predicates are too slow, they could be re-expanded by partial evaluation, but I've left the meta-predicates in for now as they make it clear that the same thing is happening in each case.

### DIFF
--- a/boot/dcg.pl
+++ b/boot/dcg.pl
@@ -319,9 +319,9 @@ dcg_terminal_pos(Pos, _) :-
     var(Pos),
     !.
 dcg_terminal_pos(list_position(F,T,_Elms,_Tail),
-                 term_position(F,T,_,_,_)).
+                 term_position(F,T,_,_,_)) :- !.
 dcg_terminal_pos(F-T,
-                 term_position(F,T,_,_,_)).
+                 term_position(F,T,_,_,_)) :- !.
 dcg_terminal_pos(Pos, _) :-
     expected_layout(terminal, Pos).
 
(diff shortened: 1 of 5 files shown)


JanWielemaker:
Closing.  Too old and term/goal expansion is planned to be completely redesigned.

PR #278, opened 2017-11-07
ADDED: Multiline CSV rows where newlines are embedded in quoted fields.

likelion:
The assumption is: if a CSV file line contains odd number of quotes (`"`) then the line is not complete and the next line is appended before invoking `row//2` parser. The lines are appended until there is an even number of quotes in the concatenated set. This handles newlines embedded in the quoted fields and quote characters represented by double-quotes (`""`).


likelion:
I think, this should fix the `@tbd` in `csv_read_file_row/3`.

library/csv.pl
```diff
@@ -396,21 +396,32 @@
...
     Row = end_of_file.
 csv_read_row(Stream, Row, Record) :-
-    read_lines_to_codes(Stream, Codes),
+    read_lines_to_codes(Stream, Codes, even),
     phrase(row(Row0, Record), Codes),
     !,
     Row = Row0.
 
-read_lines_to_codes(Stream, Codes) :-
+read_lines_to_codes(Stream, Codes, Quotes) :-
```
honnix: The `Quotes` name is a bit weird.

library/csv.pl
```diff
@@ -396,21 +396,32 @@
...
-    ;   Codes = Codes0
+    ;   check_quotes(Codes0, Quotes, even)
+    ->  Codes = Codes0
     ;   append(Codes0, [0'\n|Tail], Codes),
-        read_lines_to_codes(Stream, Tail)
+        read_lines_to_codes(Stream, Tail, odd)
     ).
 
+check_quotes([], Quotes, Quotes) :- !.
+
```
honnix: No empty line I guess normally. Applies to the following as well.

likelion: There is a caveat with empty lines. Inside a quoted field an empty line should be preserved, but it’s not (due to padding in `read_string/5`). Otherwise, empty line should fail `csv_read_file_row/3`. I will look into it closer...

honnix: Ok. Here I was referring to the code itself. :) Seems like a convention at least in this file that no empty line between simple predicates.

likelion: Got you! :) I'm externalizing my previous comment then.

likelion:
There is a caveat with empty lines. Inside a quoted field an empty line should be preserved, but it’s not (due to padding in `read_string/5`). Otherwise, empty line should fail `csv_read_file_row/3`, which won't work if padding is removed. More investigation is required.

likelion:
Empty lines inside quoted fields are preserved with a single `\n`. Empty lines between CSV lines will result in row with arity 1 and the empty field.

likelion:
@JanWielemaker Ok, I think this fix is ready now. Please, have a look. There are tests that cover various cases that I wanted the library to handle.

JanWielemaker:
Thanks.  Might take a few days as I'm a little busy right now.  It already works for you, so I suppose you are not in a hurry yourself.

likelion:
No worries, take your time. Thanks.

JanWielemaker:
Thanks.  I wish I got all patches like this!  Only moved the !s to their own line to sync with the layout standard (although I admit I prefer them at the end of the line too ...)

Merge after rebase, which causes the patch to be considered conflicting.

PR #288, opened 2018-02-18
"prepare" script rewritten

dtonhofer:
prepare script rewritten for legibility and to reorganize really bizarre (and suspiciously problematic) stuff that looks like an accumulation of historical "quick fixes". Especially the download of documentation was overly complex and frankly inscrutable. Used bash-isms instead of pure "sh" mold. 
The code has been checked with https://www.shellcheck.net/ 
Works on Linux. Unfortunately I have no access to a mingw installation, so it needs to be run there too.

### DIFF
--- a/prepare
+++ b/prepare
@@ -14,275 +14,504 @@
 #
 # On Windows installations, this script may be run from the MsysGit bash
 # shell to initialise the submodules and download the documentation.
+#
+# 2018-02-18: Rewritten for legibility and to reorganize really
+#             bizarre stuff that looks like an accumulation of 
+#             historical "quick fixes".
+#             Script has been checked with https://www.shellcheck.net/
 
-# default submodules to pull.  Use --all to get all registered submodule
-# You can also create a file modules in this directory, listing the
-# desired modules one-per-line
+set -o nounset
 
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Sorry for not integrating yet.  It will happen.  Need some time to review it and possibly make it a little
less verbose.  Does this really solve a bug/porting issue or is it merely you couldn't stand the outdated
and bad style?

Thanks!

dtonhofer:
Hi Jan. Sorry for the late reply. Actually I just rewrote it to scratch one of those itches i.e. I didn't understand exactly what it was doing - it seemed to have several elements that made no sense - I didn't know about "git modules" and I wanted to retrain in bash. It has been sitting on my disk half-finished for a year, so I though, just finish it and get it out. As is my usual problem, it took many hours more to polish than I expected. 
It's really not particularly verbose, unless there are is functionality you want to drop. I like "uncompressed" code that one can understand by reading through it. Especially in bash, being too elegant will bite later. I ran it through shellcheck too to eliminate avoidable errors. 
Best regards,
-- David

dtonhofer:
Cool. Thanks Jan.

PR #295, opened 2018-04-29
Fixed error of lacking messages' colors

TymonSerock:
During Prolog initialization and file load messages (especially error messages) did not get properly colored to draw users' attention. It was due to wrong Prolog's start sequence.


JanWielemaker:
Addressed in cf963d60b7414c7d799a9f57814827cdafac530e.  The solution is a little different, in particular loading __after__ loading the user init file such that the user can disable colors from the
init file.   Thanks!

PR #319, opened 2018-07-21
Fix option:select_option/4; only delete an option if found

royratcliffe:
`del_dict/4` fails when defaulting, and therefore `select_option/4` also fails even though a default exists. Yet the predicate is deterministic. This fix only deletes the key-value if found, and unifies `RestOptions` with `Options` unchanged otherwise.


JanWielemaker:
Thanks.  Edited a little as we can use `del_dict/4` right away.

royratcliffe:
Nice one. And thank you sir for your very excellent work on this project. I'm a big fan of yours. Salute you!

PR #321, opened 2018-07-25
Use NODERAWFS during wasm build

rla:
This removes src/Makefile.in hacks to remap file paths for the boot file creation introduced in:
<https://github.com/SWI-Prolog/swipl-devel/pull/310>

It also disables MMAP for wasm build.

Issue reference: <https://github.com/SWI-Prolog/swipl-wasm/issues/3>.

It is one step closer for automated tests for the better swi-js term interface.

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -109,11 +109,6 @@ EXPORTS=$(INCLUDEDIR)/SWI-Exports
 CINCLUDE=$(INCLUDEDIR)/SWI-Prolog.h
 STREAMH=$(INCLUDEDIR)/SWI-Stream.h
 STARTUPPATH=$(PL).prc
-ifeq (@WASM@,true)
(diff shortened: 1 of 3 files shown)


JanWielemaker:
Do I read correctly that this writes to `/swipl-devel/src/$(PL).prc`, or is this the virtual filesystem we are writing to?  If there is some way around we should avoid the need to write files outside the build directory.

rla:
It is removing those virtual paths (`/swipl-devel/`) and is now writing directly to `src/$(PL).prc` as it is done with builds for other platforms. My previous PR introduced those virtual paths (PLINIT_IN and STARTUPPATH_OUT), this one is removing them and restoring prc file generation instructions to the way they were before.

JanWielemaker:
Stupid me ... Sorry!  Merged.  Somehow the rebase due to changes I already had locally makes github beleive the change is not merged, but is is.  Closing.  Thanks!

PR #333, opened 2018-09-19
Remove duplicated meta_predicate/1 directives and use numbers for closures

pmoura:
This PR fixes issues is the `rbtrees` library module.


JanWielemaker:
Thanks!

PR #334, opened 2018-09-19
List other permissive components when there are no GPL/LGPL components

pmoura:
As an example of the modified output, with Logtalk loaded but no GPL/LGPL components, we now get:

```
?- license.
% This program contains only components covered by permissive license
% conditions. SWI-Prolog is covered by the Simplified BSD license:
% 
...
% This program contains other components with permissive licenses:
% 
%   The following components are covered by the "Apache License 2.0" license
%     (see http://www.apache.org/licenses/LICENSE-2.0)
%     Logtalk
true.
```

The output is the same as before when there's a GPL or a LGPL component.

Also relevant, note that the output no longer claims that, without any GPL/LGPL components present, that the Simplified BSD license covers everything.

Tagged as a WIP PR for now to prevent accidental merging while we discuss these changes.

### DIFF
--- a/boot/license.pl
+++ b/boot/license.pl
@@ -354,7 +354,7 @@ license_title(License) -->
     ->  true
     ;   C = License
     },
-    [ '  The following modules are covered by the "~w" license'-[C] ].
+    [ '  The following components are covered by the "~w" license'-[C] ].
 
 license_url(License) -->
     { license(License, _, Att),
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Merged.  Sorry for the late reply ... You know what I was doing ...  Note that github doesn't see the merge because I rebased it offline.

PR #350, opened 2018-10-17
A slightly more efficient version of the select/3 predicate

KommuSoft:


JanWielemaker:
Thanks.  Some measurements indicate the performance enhancement is dubious and possibly even negative on the current SWI-Prolog.  Still, providing determinism on the last element is probably already worth the trouble.  Merged after some rewrite (docs and longer aux name).

PR #356, opened 2018-11-05
Add travis-ci configuration

jgoldfar:
Not sure if there's any interest in enabling something like this, but it does build correctly! Thanks for your work on SWI-Prolog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,67 @@
+language: cpp
+
+dist: trusty
+
+os:
+  - linux
+  - osx
+
+addons:
+  apt:
+    packages:
+      - build-essential
+      - autoconf
+      - curl
+      - chrpath
+      - pkg-config
+      - ncurses-dev
+      - libreadline-dev
+      - libedit-dev
+      - libgmp-dev
+      - libssl-dev
+      - unixodbc-dev
+      - zlib1g-dev
+      - libarchive-dev
+      - libossp-uuid-dev
+      - libxext-dev
+      - libice-dev
+      - libjpeg-dev
+      - libxinerama-dev
+      - libxft-dev
+      - libxpm-dev
+      - libxt-dev
+      - libdb-dev
+      - libpcre3-dev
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,67 @@
...
+      - berkeley-db
+      - pcre
+      - jpeg
+
+git:
+  submodules: false
+
+before_script:
+  - mkdir build
+  - if [ "$TRAVIS_OS_NAME" == "linux" ] ; then sed -i 's~url = ..~url = https://github.com/SWI-Prolog~' .gitmodules; fi
```
JanWielemaker: Why is this needed?   Seems hacky and possibly this can be avoided at a different level?

jgoldfar: I agree, and I admit I didn’t look for the most elegant way to get this particular part working. Having a sub module configuration with relative paths means that in order to run a CI service one has to have all of the corresponding git repositories cloned to your username on GitHub. This isn’t an issue for your (upstream) repository, but anyone else (e.g. me, to set this up) would be unable to get things running. I could remove or comment out this line.

JanWielemaker: At least I now understand the why :)  If you only want to deal with the core you can also simply not clone the packages.  That should work fine if you use `-DINSTALL_DOCUMENTATION=OFF`. I propose to include the file above, commenting the git hackery.  Agree?  I'm happy to do so (will apply and amend the commit).

jgoldfar:
That all sounds good to me. Thanks

JanWielemaker:
Merged after update as agreed.  Please create a new pull request relative to the current master if you want further changes.

JanWielemaker:

https://swi-prolog.discourse.group/t/consolidating-the-71-github-repositories-to-simplify-maintenance-and-contribution/492/3

JanWielemaker:

https://swi-prolog.discourse.group/t/being-friendly-to-quick-contributions/493/1

Issue #358, opened 2018-11-13
SWI-Prolog on Android (termux)


erlanger:
**UPDATE**: As of SWI Prolog version 7.7.25 we are building for all Android  termux architectures: `arm`, `aarch64`, `i686`, and `x86_64`.  
### :grin: Quick'n Easy start
1. Install [termux](https://termux.com/) on your android device
2. Start termux, and paste the following in the terminal:
```sh
pkg install curl
curl  https://raw.githubusercontent.com/erlanger/swipl-termux/master/install -sSf | sh
```
More information at the [swipl-termux repo](https://github.com/erlanger/swipl-termux). 

----------
UPDATE: As of SWI Prolog version 7.7.23 we are passing all tests (with the exception of JPL). You can get it  [here](https://github.com/erlanger/swipl-termux). 

----------
Hi,
   I have been working on cross-compiling SWI-Prolog to an  **Android-arm** executable (using an x86_64 host) with [Termux](https://termux.com/). It is very promising, and I have been able to generate a termux `.deb` package to use on Android. I have even been able to run the `swipl` binary on Android, but I need some help to get to the finish line.

Problem
-----------
* Can't generate `boot32.prc` on the cross-compiling x86_64 host.

Solution:
-----------
* I need a static `swipl` binary
* __Explanation__:  With a static swipl binary I can tell  CMake to use  `qemu-arm-static` (to build `boot32.prc` on the x86_64 host) using the [CMAKE_CROSS_COMPILING_EMULATOR
](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING_EMULATOR.html) variable
* I know __this works__, because I copied the `boot32.prc` file from a raspberry-pi (32-bit) and was able to run the `swipl` executable under termux on Android.
* I can give more details and upload everything to a git repo as soon as I get it working

What I need: A Polite request
-------------------------------------
* @JanWielemaker is there a possibility to extend the SWIPL_SHARED_LIB cmake option so that it produces a static __binary__ (not just a static `libswipl`)? This will solve my problem, and we will all get SWI-Prolog on Android!! (without having to install a linux distro).
* A static `defatom` and `mkvmi` will make things much easier, but I can fix that myself if need be.

NOTE: the reason why I need a static binary is because I need to run the Android/arm `swipl` **on the x86_64 host** (to generate `boot32.prc` - I can't use a 64 bit version of swipl because it will generate the wrong boot.prc file)

[...]

PR #359, opened 2018-11-18
PORT: Android minizip USE_FILE32API only in 32 bit targets

erlanger:
See issue #358 for more information.


JanWielemaker:
Thanks.  Merged after edit (hence the conflict).  There is a more robust way to test for 32-bits and I'd like to stick to the cmake convention that reconfiguring is almost silent.

erlanger:
Thanks for merging!

PR #360, opened 2018-11-18
PORT: Use host compiler to build mkvmi and defatom

erlanger:
As discussed in #358


JanWielemaker:
Thanks.  I've done a little editing: (1) moved the build and use together in CMakefiles.txt for readability, (2) use a variable PROG_MKVMI such that we can still use the plain target for native compilation (which may require special tricks to run the executable) (3) define and use `CMAKE_HOST_CC`.   Tested native, for wasm and windows.   I think we can close this one.

erlanger:
Thanks for merging, tested it and it works for android/termux.

PR #363, opened 2018-11-20
DOC: improve SWIPL_NATIVE_FRIEND explanation

erlanger:


erlanger:
Thanks for merging! By the way, I wanted to thank you for your diligence, patience and perseverance with SWI-Prolog.

JanWielemaker:
This makes progress without having to explain much, so I'm more than happy.

PR #366, opened 2018-11-22
ENHANCED: allow installation of tests with -DINSTALL_TESTS

erlanger:
First try to install tests (to be executed in the target device) and enhance `test.pl` to test packages.
Discussed in #358.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,9 @@ option(BUILD_TESTING
 option(BUILD_SWIPL_LD
        "Create the swipl-ld utility"
        ON)
+ option(INSTALL_TESTS
+       "Install script and files needed to run tests of the final installation"
+       OFF)
 
 include(Utils)
 include(BuildType)
@@ -109,6 +112,9 @@ endif(WIN32)
 set(SWIPL_INSTALL_LIBRARY  ${SWIPL_INSTALL_PREFIX}/library)
 set(SWIPL_INSTALL_BOOT     ${SWIPL_INSTALL_PREFIX}/boot)
 set(SWIPL_INSTALL_INCLUDE  ${SWIPL_INSTALL_PREFIX}/include)
+if(INSTALL_TESTS)
+   set(INSTALL_TESTS_DIR   ${SWIPL_INSTALL_PREFIX}/test)
+endif()
 
 if(MSVC)
   add_compile_options(/W3)
@@ -166,6 +172,24 @@ if(INSTALL_DOCUMENTATION)
   install(FILES packages/index.html
 	  DESTINATION ${SWIPL_INSTALL_PREFIX}/doc/packages)
 endif(INSTALL_DOCUMENTATION)
+
+# Install a prolog script to run tests on target device
+# in which ctest is not available
+if(INSTALL_TESTS)
+   set(INSTALL_TESTS_DIR ${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_PREFIX}/test)
+   set(PKGS_SRC_DIR ${CMAKE_CURRENT_SOURCE_DIR}/packages)
+   set(INSTALL_TESTS_DB ${CMAKE_BINARY_DIR}/cmake_pkg_tests.db)
+   #Move test db to installation
+   install(FILES       ${INSTALL_TESTS_DB}
+           DESTINATION ${INSTALL_TESTS_DIR}/)
+   file(REMOVE ${INSTALL_TESTS_DB})
+   #Copy generic package test files to the installation
+   install(DIRECTORY   ${PKGS_SRC_DIR}
+           DESTINATION ${INSTALL_TESTS_DIR}
+           FILES_MATCHING REGEX
+             ".*/.*/test_.*$|^.*/Test[s]*/.*$|^.*/test[s]*/.*"
+          )
+endif(INSTALL_TESTS)
 
 # Add the packages
 foreach(package ${SWIPL_PACKAGE_LIST})
--- a/cmake/CheckInstallation.cmake
+++ b/cmake/CheckInstallation.cmake
@@ -1,0 +1,5 @@
+# Build check_installation.pl in the home directory
+# it is calles by InstallSource.cmake
+configure_file(../../library/check_installation.pl.in
+               ${SWIPL_BUILD_LIBRARY}/check_installation.pl
+               @ONLY)
--- a/library/check_installation.pl.in
+++ b/library/check_installation.pl.in
@@ -35,7 +35,9 @@
 
 :- module(check_installation,
           [ check_installation/0,
-            check_installation/1                % -Issues
+            check_installation/1,               % -Issues
+            test_installation/0,
+            test_installation/1                 % -Options
           ]).
 
 /** <module> Check installation issues and features
@@ -135,6 +137,8 @@ issue_base('http://www.swi-prolog.org/build/issues/').
 %        are present (e.g., unbounded arithmetic support)
 %     2. Test that all standard libraries that depend on foreign
 %        code are present.
+%     3. provides a test_installation predicate to run the tests
+%        at runtime if the system was built with -DINSTALL_TESTS
 %
 %   If issues are found it prints a   diagnostic message with a link
 %   to a wiki page with additional information about the issue.
@@ -149,7 +153,7 @@ check_installation :-
         print_message(warning, installation(imperfect(Count)))
     ).
 
-%!  check_insmtallation(-Issues:list(pair)) is det.
+%!  check_installation(-Issues:list(pair)) is det.
 %
 %   As check_installation/0, but additionally  returns   a  list  of
 %   Component-Problem pairs. Problem is  one of `optional_not_found`
@@ -171,6 +175,28 @@ check_installation_(Issues) :-
     forall(component(Source, _Properties),
            check_component(Source)),
     findall(I, retract(issue(I)), Issues).
+
+test_installation :-
+    test_installation([]).
+
+test_installation(Options) :-
+    '@INSTALL_TESTS@' = 'ON'
+    ->  test_installation_run(Options)
+    ;   print_message(informational, installation(testing(no_installed_tests))).
+
+test_installation_run(Options) :-
+   current_prolog_flag(home,H),
+   atom_concat(H,'/test',TstDir),
+   atom_concat(TstDir,'/test.pl',TestPl),
+   load_files(TestPl, [ silent(true), if(changed), module(user) ]),
+   set_test_options([sto(false)]),
+   setup_call_cleanup(
+       working_directory(OldDir, TstDir),
+       test([], Options),
+       ( unload_file(TestPl),
+         working_directory(_, OldDir)
+       )
+   ).
 
 check_component(Source) :-
     component(Source, Properties),
@@ -513,6 +539,9 @@ message(jquery(found(Path))) -->
     [ '  jQuery from ~w'-[Path] ].
 message(jquery(not_found(File))) -->
     [ '  Cannot find jQuery (~w)'-[File] ].
+message(testing(no_installed_tests)) -->
+    [ '  Runtime testing is not enabled.', nl],
+    [ '  Please recompile the system with INSTALL_TESTS enabled.' ].
 
 
 public_executable(EXE, PublicProg) :-
--- a/packages/cmake/PrologPackage.cmake
+++ b/packages/cmake/PrologPackage.cmake
@@ -278,16 +278,22 @@ endfunction()
 
 # test_lib(name
 #	   [PACKAGES ...]
-#	   [PARENT_LIB])
+#	   [PARENT_LIB]
+#	   [TEST_FILES_REGEX <regex>])
 #
 # Run test_${name} in test_${name}.pl
+#
+# TEST_FILES_REGEX is only needed if the test files are *not* under a
+# directory named test[s] or Test[s]. It is one regular expression
+# which will be matched against the full path of the package source
+# and build directories.
 
 if(NOT SWIPL_PATH_SEP)
   set(SWIPL_PATH_SEP ":")
 endif()
 
 function(test_lib name)
-  cmake_parse_arguments(my "PARENT_LIB" "NAME" "PACKAGES" ${ARGN})
+  cmake_parse_arguments(my "PARENT_LIB" "NAME;TEST_FILES_REGEX" "PACKAGES" ${ARGN})
   set(test_goal "test_${name}")
   set(test_source "${CMAKE_CURRENT_SOURCE_DIR}/test_${name}.pl")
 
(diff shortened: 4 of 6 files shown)


erlanger:
I missed the fact that some packages don't have a test_<pkgname>.pl script including all the other tests.

Should I rework the test.pl script to run any test_*.pl file within the package directory? Or is it better to require a package to have a test_<pkgname>.pl script?

erlanger:
Some tests actually call `halt` (e.g. inclpr)or in the case of JPL on android are aborting (probably `SIGABRT`).

 This means the `test.pl` framework used for packages can't use `load_files`, I am thinking of simply starting an new `swipl` process for each package. Do you think this is the right approach?

JanWielemaker:
I scanned though the patch.  As we have some open issues I won't merge right now.  Remarks:

  - I'd be in favour of just copying the test_*.pl files from the packages.  That should work for all
    but a few that require additional data.  For those we can add a conditional bit to the package
    CMakeLists.txt to copy the test data.
 - test_* scripts that call halt should be rewritten.  
 - test scripts that may abort should be rewritten.  I think that is realistically only the java test.
   The script may simply execute Prolog itself using process_create/3 using the executable
   as available through the Prolog flag.   I guess the main driver can set some flag so we know
   this is needed.
 - I don't think we need a test_<package>.pl  Not sure here.  I think there should be none/very few
   exceptions and if these exist we can probably rename the file or add a block-list to test.pl

Does this make sense?

erlanger:
> Does this make sense?

Sure, good guidance!
I'll get to work on it.

> test scripts that may abort should be rewritten. I think that is realistically only the java test.
The script may simply execute Prolog itself using process_create/3 using the executable
as available through the Prolog flag.

The only problem here is that running on Android requires the environment variable `LD_PRELOAD=libswipl.so:libm.so` (and for JPL) `libnative.so` before running the executable. `process_create/3` is not passing the environment variables (which is the reason why the 3 `saved_state` tests fail, since they use `process_create` to run the newly minted executable.

This happens because the Android dynamic linker (the equivalent of `ld.so`) is broken, and makes symbols globally visible only for libraries in LD_PRELOAD (or for symbols on the main file).

I'll experiment linking each package `.so` with `libswipl.so` and `libm.so` and to set some variable in the package CMakeLists.txt to point to extra libraries that may be needed (like `libnativehelper.so` for JPL on Android). I think I'll try to solve this first, and then continue the work on `INSTALL_TESTS`.
Does this seem reasonable?

JanWielemaker:
> `process_create/3` is not passing the environment variables

If that is true, it is broken.   In Linux (Fedora 28 today), I get a nice environment using

    ?- process_create(path(printenv), [], []).

erlanger:
> If that is true, it is broken. In Linux (Fedora 28 today), I get a nice environment using
> ?- process_create(path(printenv), [], []).

hmm, you are right,  I presumed there was a problem with that due the [documentation](http://www.swi-prolog.org/pldoc/doc_for?object=process_create/3) in  which says:
 "Note that the current implementation does not pass any environment variable"

I just tested, and `process_create` is properly passing the environment variables. But there is some other problem with it:
```
$ pwd
/data/data/com.termux/files/usr/lib/swipl/test/Tests/save
$ swipl -o p -g echo_true -c input/plain.pl
$ ls -al p
-rwx------    1 u0_a241  u0_a241     667471 Nov 22 07:53 p
$ swipl
Welcome to SWI-Prolog (threaded, 32 bits, version 7.7.22-28-g3d27e9a6a-DIRTY)
SWI-Prolog comes with ABSOLUTELY NO WARRANTY. This is free software.
Please run ?- license. for legal details.

For online help and background, visit http://www.swi-prolog.org
For built-in help, use ?- help(Topic). or ?- apropos(Word).

?- process_create('./p',[],[]).
/data/data/com.termux/files/usr/lib/swipl/test/Tests/save/p: No such file or directory
ERROR: Process "/data/data/com.termux/files/usr/lib/swipl/test/Tests/save/p": exit status: 1
ERROR: In:
ERROR:    [9] process:process_create('/data/data/com.termux/files/usr/lib/swipl/test/Tests/save/p',[])
ERROR:    [7] <user>
ERROR:
ERROR: Note: some frames are missing due to last-call optimization.
ERROR: Re-run your program in debug mode (:- debug.) to get more detail.
?-
% halt
$ ./p
true.
$
```

But this works:
```
?- process_create(path(ls),['-al','./p'],[]).
-rwx------    1 u0_a241  u0_a241     667471 Nov 22 07:53 ./p
true.
```

and this:
```
$ MYVAR=hello swipl
Welcome to SWI-Prolog (threaded, 32 bits, version 7.7.22-28-g3d27e9a6a-DIRTY)
SWI-Prolog comes with ABSOLUTELY NO WARRANTY. This is free software.
Please run ?- license. for legal details.

For online help and background, visit http://www.swi-prolog.org
For built-in help, use ?- help(Topic). or ?- apropos(Word).

?- process_create(path(env),[],[]).
USER=u0_a241
SSH_CLIENT=<deleted>
EXTERNAL_STORAGE=/storage/emulated/legacy
LD_LIBRARY_PATH=/data/data/com.termux/files/usr/lib
SHLVL=1
OLDPWD=/data/data/com.termux/files/usr/lib/swipl/test/Tests
HOME=/data/data/com.termux/files/home
SSH_TTY=/dev/pts/1
TMPDIR=/data/data/com.termux/files/usr/tmp
LOGNAME=u0_a241
_=/data/data/com.termux/files/usr/bin/swipl
TERM=xterm-256color
ANDROID_DATA=/data
PATH=/data/data/com.termux/files/usr/bin:/data/data/com.termux/files/usr/bin/applets
ANDROID_ROOT=/system
LANG=en_US.UTF-8
LD_PRELOAD=libswipl.so:libm.so
TMP=/data/data/com.termux/files/home/tmp
PREFIX=/data/data/com.termux/files/usr
SHELL=/data/data/com.termux/files/usr/bin/sh
MYVAR=hello
PWD=/data/data/com.termux/files/usr/lib/swipl/test/Tests/save
SSH_CONNECTION=<deleted>
LINES=31
COLUMNS=195
true.
```

JanWielemaker:
> Does this make sense to you?

Not for `process_create/3` as this comes from `packages/clib/process.c`.  Otherwise it makes sense to replace all hard-coded "/bin/sh" with the UNIX_SHELL.  Note that this particular one is written by library(qsave) from Prolog.  I guess we need a Prolog flag `shell` that is initialized from UNIX_SHELL in
`os/pl-prologflag.c`.  Ideally we also use this shell from shell/0 and shell/1, such that the user can set the
shell from Prolog.

JanWielemaker:
Will you add the flag `shell` or shall I do so?

erlanger:
> Not for process_create/3 as this comes from `packages/clib/process.c`.

Ahh, you're right about the different file, but we have the same problem in [`packages/clib/process.c`](https://github.com/SWI-Prolog/packages-clib/blob/master/process.c#L1658):
```c
   if ( info->envp )
      execve(info->exe, info->argv, info->envp);
    else
      execv(info->exe, info->argv);
```
So if we call without an environment then `execv` is called (and not `execve`).


> Otherwise it makes sense to replace all hard-coded "/bin/sh" with the UNIX_SHELL. Note that this particular one is written by library(qsave) from Prolog. I guess we need a Prolog flag shell that is initialized from UNIX_SHELL in os/pl-prologflag.c. Ideally we also use this shell from shell/0 and shell/1, such that the user can set the shell from Prolog.

Great idea! I'll work on it.

erlanger:
> Will you add the flag shell or shall I do so?

I'll work on it

JanWielemaker:
> info->envp

Note that process_create/3 uses posix_spawn() when available and fork/exec only if a directory is specified as posix_spawn strangely cannot change directory for the child.

The environment handling requires some cleanup and a way to extend the environment, leaving existing variables alone.    I'll have a look.

erlanger:
For Android:
```
-- Looking for posix_spawn
-- Looking for posix_spawn - not found
```

erlanger:
> Ok. Of course we can simply change the execv to execve and pass environ. I guess we can so so unconditionally?

Yes, I think this is the best. It will allow running things normally under termux.

BTW, I also sent the PR with qsave using the shell prolog flag. I think we should also use the shell prolog flag for as much as we can, because we should not depend on `termux-exec` to shuffle shebangs around.

erlanger:
New implementation of -DINSTALL_TESTS
-----------------------------------------------------------
Okay, I think I have a much cleaner way to run the tests at run time (issue #366). Included are the patches that do the following:
* During build time write a `cmake_pkg_tests.db`
* `cmake_pkg_tests.db` is a list of facts indicating the different tests that `ctest` runs. It has enough information to run each test within `test.pl`.
* `test.pl` is modified to be able to run these tests with the `--packages` or `--package=XXX` command line options.
* Updated some packages so that they load modules using `file_search_path` so that we avoid conflicts with duplicate files, prevent duplicate copies, and allow the developers to test with local modules in the source directory.

Don't know what the conflict with `packages/ssl` is, I probably just missed a previous commit you made.

erlanger:
I would like to run `test.pl` from within swipl, having a predicate similar to `check_installation`. Something like `test_installation`. It will just run all the tests if the build was compiled with `-DINSTALL_TESTS` otherwise it prints an informational message saying the no tests are available because the build was done without `-DINSTALL_TESTS`. What do you think?

JanWielemaker:
Looks good (not tested though).   I would like to see the installation of specific files for packages such as Yaml, protobufs, etc., to be moved into the packages themselves.  This improves modularity and ensures we install only tests for packages that are actually installed.   The general copy of test files for a package can probably go to add_test(), so again we only install for packages that are really active.

Agree?

Getting really close!   Being able to install the tests is a nice addition.  They are not that big and doing so might be worthwhile in many cases were the build environment differs from the deployment environment.

JanWielemaker:
> I would like to run `test.pl` from within swipl, having a predicate similar to `check_installation`. Something like `test_installation`. It will just run all the tests if the build was compiled with `-DINSTALL_TESTS` otherwise it prints an informational message saying the no tests are available because the build was done without `-DINSTALL_TESTS`. What do you think?

I'm in favor.  I think that can be added to `check_installation.pl`.  It is related closely enough.

erlanger:
> Looks good (not tested though). I would like to see the installation of specific files for packages such as Yaml, protobufs, etc., to be moved into the packages themselves. This improves modularity and ensures we install only tests for packages that are actually installed. The general copy of test files for a package can probably go to add_test(), so again we only install for packages that are really active.
> 
> Agree?

Great, yes, this makes total sense, I should have thought about it.

erlanger:
> I'm in favor. I think that can be added to `check_installation.pl`. It is related closely enough.
Good, I'll work on it.

erlanger:
Workng
----------
Phew! That was quite a bit of work, but it is done!
The following is working:
* `ctest -j 8` from the build directory
*  `<install_dir>/bin/swipl -f none test.pl` from the `<PREFIX>/lib/swipl/test` directory
*  `<install_dir>/bin/swipl -f none test.pl --packages` from the `<PREFIX>/lib/swipl/test` directory
*  `<install_dir>/bin/swipl -f none test.pl --package=<pkg>` from the `<PREFIX>/lib/swipl/test` directory

Help needed
-----------------
* Running `test_installation` from the top level gives this error (all the other tests work):
```
(test_call.pl)...................
ERROR: /tmp/s1/lib/swipl/test/Tests/core/test_call.pl:159:
        test c_cm1 is subject to occurs check (STO):
ERROR: Finite trees (error checking): received error: plunit_cross_module_call:c_cm1/1: Undefined procedure: foo:cm1/1
ERROR: Rational trees: received error: plunit_cross_module_call:c_cm1/1: Undefined procedure: foo:cm1/1
ERROR: Finite trees: received error: plunit_cross_module_call:c_cm1/1: Undefined procedure: foo:cm1/1
ERROR: /tmp/s1/lib/swipl/test/Tests/core/test_call.pl:162:
        test c_cm2 is subject to occurs check (STO):
ERROR: Finite trees (error checking): received error: plunit_cross_module_call:c_cm2/1: Undefined procedure: foo:cm2/1
ERROR: Rational trees: received error: plunit_cross_module_call:c_cm2/1: Undefined procedure: foo:cm2/1
ERROR: Finite trees: received error: plunit_cross_module_call:c_cm2/1: Undefined procedure: foo:cm2/1
```

I can't figure out how this error is happening, because I have the plunit `STO` flag set to `false` for the core/main tests (proved by the fact that it runs OK from the command-line, but not from the top-level).
Could you take a look to see what the problem is?

JanWielemaker:
> Workng

Thanks a lot.  I'll do the next round reviewing and see whether I can find more about the above issue.  It will probably take a few days though if I look at my agenda for today and tomorrow.

JanWielemaker:
I had to rush after pushing the merges.   You found them anyway :)   Your patches worked as is.  Some of the modifications to the packages were not needed.  I guess the result of older attempts.

> So probably the best thing to do is to add a COMMAND argument to the `test_lib` function so that it can be recorded in the `cmake_pkg_tests.db` file. What do you think?

Makes sense to me.

I do see there is far too much data in the `test` dir, including even the `.git`.  I'll have a look at that.  If we can get it small we could for example include it with the Windows distribution as the binary runs a wide variety of Windows versions, but possibly not always as it should ...

JanWielemaker:
Pushed a series of commits that restrict installation of test files for just the needed files.    Please use this as the basis for further patches.   You might consider just adding a `if(INSTALL_TESTS)` section to the JPL CMakeLists.txt file.  That might be easier to understand than more logic to test_libs().  I decided for this route for some packages that needed files and directories from weird places.

It is still a little big to add to e.g. the Windows installers (6.5 Mb), but we could add a .zip (960Kb).

erlanger:
>  You might consider just adding a `if(INSTALL_TESTS)` section to the JPL CMakeLists.txt file.

This won't work for the JPL test that I was mentioning, because the test consists in running a Java VM connecting to swipl, it is not the case of a missing file. 

In any case, this is useless for Android, because  a java VM is not started from the command line.  The only reason why I was concerned about it is for 100% consistency between `test_installation` and  `ctest`. In this case it is 99.99% consistent, only missing that `prolog_in_java` test.


> 
> It is still a little big to add to e.g. the Windows installers (6.5 Mb), but we could add a .zip (960Kb).

Would not the windows installer zip the contents automatically?

JanWielemaker:
> In this case it is 99.99% consistent, only missing that `prolog_in_java` test.

That is fine with me, surely for now.   As a related question: is this Termux version just that or is it a promising step to create Android apps that contain Prolog?  I'd assume this requires us to be able to embed Prolog in Java, no?

> Would not the windows installer zip the contents automatically?

Yes.  Adding 10% to both the installer and the installed system seems a bit of an overkill for this though.   I'm keen to avoid SWI-Prolog becoming _bloatware_ ...

As far as I'm concerned we can close this issue.  Please do if you agree.

erlanger:
>  As a related question: is this Termux version just that or is it a promising step to create Android apps that contain Prolog? I'd assume this requires us to be able to embed Prolog in Java, no?
> 

Yes, it is certainly a very promising step, since termux itself is a Java Android app. 

> Yes. Adding 10% to both the installer and the installed system seems a bit of an overkill for this though. I'm keen to avoid SWI-Prolog becoming _bloatware_ ...

Yes, I agree. It will be used mostly by developers.

PR #367, opened 2018-11-22
PORT: Link packages to libswipl.so and libm.so on Android

erlanger:
The Android dynamic linker does not make symbols globally visible unless the symbols come from the main file or from a library loaded by `LD_PRELOAD`.
So we need to link each package with `libswipl.so`
As discussed in #358 and #366.


erlanger:
Great, thanks for merging!

PR #368, opened 2018-11-22
Add shell prolog flag

erlanger:
Add prolog flag to point to the current shell.
As discussed in #366

### DIFF
--- a/library/qsave.pl
+++ b/library/qsave.pl
@@ -194,7 +194,7 @@ make_header(Out, SaveClass, _Options) :-
     current_prolog_flag(unix, true),
     !,
     current_prolog_flag(executable, Executable),
-    current_prolog_flag(shell, Shell),
+    current_prolog_flag(posix_shell, Shell),
     format(Out, '#!~w~n', [Shell]),
     format(Out, '# SWI-Prolog saved state~n', []),
     (   SaveClass == runtime
--- a/library/qsave.pl
+++ b/library/qsave.pl
@@ -194,7 +194,8 @@ make_header(Out, SaveClass, _Options) :-
     current_prolog_flag(unix, true),
     !,
     current_prolog_flag(executable, Executable),
-    format(Out, '#!/bin/sh~n', []),
+    current_prolog_flag(shell, Shell),
+    format(Out, '#!~w~n', [Shell]),
     format(Out, '# SWI-Prolog saved state~n', []),
     (   SaveClass == runtime
     ->  ArgSep = ' -- '
--- a/library/shell.pl
+++ b/library/shell.pl
@@ -66,19 +66,16 @@ the amount of available memory.
 %!  shell
 %
 %   Execute an interactive shell. The executed   shell is defined by
-%   the environment =SHELL= or =comspec=   (Windows).  If neither is
-%   defined, =|/bin/sh|= is used.
+%   the environment variable =SHELL= or =comspec=  for (Windows).
+%   If neither is defined, the prolog flag =|shell|= is used.
 
-shell :-
-    getenv('SHELL', Shell),        % Unix, also Cygwin
-    !,
-    shell(Shell).
 shell :-
     getenv(comspec, ComSpec),      % Windows
     !,
     shell(ComSpec).
 shell :-
-    shell('/bin/sh').
+    current_prolog_flag(posix_shell,Shell),
+    shell(Shell).
 
 %!  cd.
 %!  cd(Dir).
--- a/src/os/pl-prologflag.c
+++ b/src/os/pl-prologflag.c
@@ -1455,6 +1455,15 @@ initPrologFlags(void)
   setPrologFlag("mitigate_spectre", FT_BOOL, FALSE, PLFLAG_MITIGATE_SPECTRE);
 #endif
 
+#if defined(__unix__)
+  const char* env_shell = getenv("SHELL");
+  if (env_shell != NULL) {
+     setPrologFlag("posix_shell", FT_ATOM, env_shell);
+  } else {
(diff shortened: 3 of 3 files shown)


JanWielemaker:
I am a little worried about this after seeing the Windows hack.   I have my doubt about the hard coded path to `cmd.exe`.  One option might be to get that through `%comspec%` and the other might be not to set this variable for Windows at all because you cannot pass it the same things anyway.   In that case we should rename the variable to `unix_shell` or possibly better `posix_shell`.   What do you think?

I'm also a bit curious about the interaction with the environment variable in shell/0.  Wouldn't it be better to initialise the flag from the `SHELL` variable if this is set and only set it to UNIX_SHELL if the environment
variable does not exists?

erlanger:
> not to set this variable for Windows at all because you cannot pass it the same things anyway.

I think this is the best thing to do; I was not going to the windows setting at all, but I thought we would discuss it.

>  Wouldn't it be better to initialise the flag from the SHELL variable if this is set and only set it to UNIX_SHELL if the environmentvariable does not exists?

Yes, I think this is much better. For some reason I thought that it would be more flexible to use SHELL at runtime, but this doesn't make sense.
Patch coming

JanWielemaker:
Thanks.  Pushed two patches that make System() and process_create/3 always use execve()

erlanger:
> Thanks. Pushed two patches that make System() and process_create/3 always use execve()

That's good, but why eliminate the `env(...)` option? Perhaps it was an oversight.

JanWielemaker:
Thanks.  Merged after edit: use for System(), document and some minor issues.

erlanger:
Great! Thanks!

JanWielemaker:
Hmmm ...  We had a user with `SHELL=/bin/tcsh` and that now breaks saved states :worried: 

What to do?   Only using this variable if the basename is `sh` or `bash` seems hacky, but what else?

erlanger:
> Hmmm ... We had a user with `SHELL=/bin/tcsh` and that now breaks saved states worried
> 
> What to do? Only using this variable if the basename is `sh` or `bash` seems hacky, but what else?

Yeah, I think that is too hacky. I think we could do this:

1. Leave the `posix_shell` prolog flag as it is, so that all shell related commands work properly.
2. Do not use the `posix_shell` prolog flag for `library(qsave)`, instead use the `UNIX_SHELL #cmakedefine` which comes from the `-DUNIX_SHELL` cmake variable at compile time. In this way `qsave` will not    depend on the `SHELL` environment variable or on the current `posix_shell` flag, which seems good to me.

This seems reasonable, because the requirement comes from saved states needing a sh-compatible variable which is set at compile time.

What do you think?

JanWielemaker:
Yeah.  I see that `system()` is hard-wired to use `/bin/sh` according to the Linux docs.  `posix_shell`
however suggests we have a POSIX compliant shell and `tcsh` is not.  So, I think the way out is to 

  1. Rename the current `posix_shell` flag to `shell`
  2. Have a new `posix_shell` that is instantiated from UNIX_SHELL (better renamed into
      POSIX_SHELL) and use that for saved states and possibly a few other places.

Will you make a patch today or shall I do such?  I'll try to release 7.7.24 in the afternoon.

erlanger:
> 
> Will you make a patch today or shall I do such? I'll try to release 7.7.24 in the afternoon.

Could you do it?  That will help me out.

JanWielemaker:
Ok.

erlanger:
> Yeah. I see that `system()` is hard-wired to use `/bin/sh` according to the Linux docs. `posix_shell`
> however suggests we have a POSIX compliant shell and `tcsh` is not. So, I think the way out is to
> 
>     1. Rename the current `posix_shell` flag to `shell`
> 
>     2. Have a new `posix_shell` that is instantiated from UNIX_SHELL (better renamed into
>        POSIX_SHELL) and use that for saved states and possibly a few other places.
> 
> 

Perhaps `shell` is not needed at all, which predicate will use it? `shell/1` works like `system` in linux, so it should probably use `posix_shell`.  Am I missing something?

JanWielemaker:
> Perhaps `shell` is not needed at all

I'm a little in doubt about that, indeed.  In theory `shell/0` should use it, but that is one of the few exceptions I can think of.

erlanger:
> > Perhaps `shell` is not needed at all
> 
> I'm a little in doubt about that, indeed. In theory `shell/0` should use it, but that is one of the few exceptions I can think of.

Yes, `shell/0` should probably use it; I was thinking of that also but I thought it was not worth the trouble since having a posix shell when you run `shell/0` is not that bad given the fact that it will be used interactively and you could just run whatever shell you wanted within the posix shell.

JanWielemaker:
... Being the Prolog flag `posix_shell`.  Yes.  That is ok.  So I instantiate `posix_shell` from CMake
`POSIX_SHELL` (renamed from `UNIX_SHELL`) and change shell/0 as above and that is it, no?

erlanger:
> ... Being the Prolog flag `posix_shell`. Yes. That is ok. So I instantiate `posix_shell` from CMake
> `POSIX_SHELL` (renamed from `UNIX_SHELL`) and change shell/0 as above and that is it, no?

Yup,  I think this works.

erlanger:
> > ... Being the Prolog flag `posix_shell`. Yes. That is ok. So I instantiate `posix_shell` from CMake
> > `POSIX_SHELL` (renamed from `UNIX_SHELL`) and change shell/0 as above and that is it, no?
> 
> Yup, I think this works.

But the order for `shell/0` should be: (1) Try environment, (2) `posix_shell` prolog flag and (3) `POSIX_SHELL #define`

The order for `shell/1`  and `library(qsave)` should be: (1) `posix_shell` prolog flag, (2) `POSIX_SHELL #define`

PR #369, opened 2018-11-22
PORT: Add ability to define default tmp dir

erlanger:
Android doesn't have `/tmp` directory. So we need a cmake variable to specify it at build time. This PR adds the SWIPL_DEFAULT_TMP_DIR cmake variable that determines the default temp dir to be used if the TMP or TEMP environment variable is not defined.

Part of #358.

Let me know if you think this works!


src/os/pl-prologflag.c
```diff
@@ -1323,6 +1323,7 @@ initPrologFlags(void)
   setPrologFlag("c_libplso", FT_ATOM, C_LIBPLSO);
   setPrologFlag("c_ldflags", FT_ATOM, C_LDFLAGS);
   setPrologFlag("c_cflags",  FT_ATOM, C_CFLAGS);
+  setPrologFlag("default_tmp_dir", FT_ATOM|FF_READONLY, SWIPL_DEFAULT_TMP_DIR);
```
JanWielemaker: I'd be more in favor of `tmp_dir` and initialize it the same way as the system decides on its tmp dir.   Next, make the places where this is used use the Prolog flag.  Can you manage that (bit like `posix_shell`)?

erlanger: Sure, sounds great!

erlanger: Patch coming a little later.

JanWielemaker:
Great!  Merged after some more edits.   All nice improvements to the overall portability!

erlanger:
Great! Thanks!

PR #370, opened 2018-11-23
PORT: Various minor fixes for the Android port

erlanger:
* Use the termux provided `gmp` if compiling under termux
* Set the android architecture properly
* Link to android's `liblog.so` so that symbols are visible for packages that print to syslog
* Add symlink to `libswipl.so` since `rpath` is broken in many android versions

As always, I welcome changes/comments. 
Part of #358

### DIFF
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -32,7 +32,7 @@ FUNCTION(ILINK from to)
   get_filename_component(LNTDIR ${to} DIRECTORY)
   get_filename_component(LNTNAME ${from} NAME)
   file(RELATIVE_PATH LNLNK ${LNTDIR} ${from})
-  install(CODE "EXECUTE_PROCESS(COMMAND ln -sf ./${LNLNK} ./${LNTNAME}
+  install(CODE "EXECUTE_PROCESS(COMMAND ln -sf ${LNLNK} ./${LNTNAME}
 				WORKING_DIRECTORY \$ENV{DESTDIR}${LNTDIR})")
 ENDFUNCTION(ILINK)
 
--- a/cmake/Utils.cmake
+++ b/cmake/Utils.cmake
@@ -32,7 +32,7 @@ FUNCTION(ILINK from to)
   get_filename_component(LNTDIR ${to} DIRECTORY)
   get_filename_component(LNTNAME ${from} NAME)
   file(RELATIVE_PATH LNLNK ${LNTDIR} ${from})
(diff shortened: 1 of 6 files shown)


cmake/Utils.cmake
```diff
@@ -32,7 +32,7 @@ FUNCTION(ILINK from to)
   get_filename_component(LNTDIR ${to} DIRECTORY)
   get_filename_component(LNTNAME ${from} NAME)
   file(RELATIVE_PATH LNLNK ${LNTDIR} ${from})
-  install(CODE "EXECUTE_PROCESS(COMMAND ln -sf ${LNLNK} ./${LNTNAME}
+  install(CODE "EXECUTE_PROCESS(COMMAND ln -sf ./${LNLNK} ./${LNTNAME}
```
JanWielemaker: Why is this?

src/CMakeLists.txt
```diff
@@ -534,3 +534,11 @@ ilink(${CMAKE_INSTALL_PREFIX}/${SWIPL_INSTALL_ARCH_EXE}/swipl-ld
 endif()
 
 endif(WIN32)
+
+# Create symbolic link to lib directory, since rpath is broken
+# in many android versions
+if(ANDROID)
```
JanWielemaker: I guess ideally we have an option to install libswipl in `${CMAKE_INSTALL_PREFIX}/lib` _instead_ of in the local tree and forget about the RPATH (I think that is automatic)  We could than activate this for Android, but the Debian distributions also prefer installations without `rpath`.   Is that a plan?

cmake/Utils.cmake
erlanger: This is because ${LNLNK} turns out to be something like `swipl/lib/<arch>...` if we are linking under the same parent directory; and of course the symbolic link is broken then, since we need `./swipl/lib/<arch>`.

src/CMakeLists.txt
erlanger: Good, should it be caled SWIPL_INSTALL_IN_LIB?

cmake/Utils.cmake
erlanger: hmmm you're right, confused it with another error.

JanWielemaker:
Ok.  This one is done.  Moved some stuff around, merged the two last commits, etc.  Also started a now mostly empty cmake/ports/Android.cmake where as much as possible the conditional Android stuff should go.

erlanger:
Great! Thanks!

PR #375, opened 2018-11-24
INSTALL: create tmp_dir if it does not exist

erlanger:
Make sure the SWIPL_TMP_DIR directory exists.
Part of #358.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -464,6 +464,11 @@ else()
   set(LIBSWIPL_DIR ${SWIPL_INSTALL_ARCH_LIB})
 endif()
 
+#Make sure tmp directory exists
+if(NOT IS_DIRECTORY ${SWIPL_TMP_DIR})
+install(DIRECTORY DESTINATION ${SWIPL_TMP_DIR})
+endif()
+
 install(TARGETS swipl libswipl
 	RUNTIME DESTINATION ${SWIPL_INSTALL_ARCH_EXE}
         LIBRARY DESTINATION ${LIBSWIPL_DIR}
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -464,6 +464,11 @@ else()
   set(LIBSWIPL_DIR ${SWIPL_INSTALL_ARCH_LIB})
 endif()
 
+#Make sure tmp directory exists
+if(NOT IS_DIRECTORY ${SWIPL_TMP_DIR})
+install(DIRECTORY DESTINATION ${SWIPL_TMP_DIR})
```
JanWielemaker: I understand this normally creates `/tmp` or something similar in the termux tree?  Is creating in line with Android or Termux policies?  If this is created it should probably be created with the proper mode, e.g., `01777`?  Note there is AFAIK, no built-in feature that uses `/tmp`, so we can also make a general comment that an OS may not provide a default configured tmp directory.

erlanger: Android has a very different security model. Every application is its own user, so there is not a general `/tmp` directory. What this does in the current build setup is to create a `tmp` directory for SWI-Prolog within the termux user directory structure. This is totally okay, since the termux user often creates his own directories to save scripts, etc.

I will update the permissions, but on Android it won't have any effect, because the directory is not shared with other users.

JanWielemaker: Ok.  In that case the patch should be fine as is.  If you want to do anything to it, just comment it properly so other people will not ask the same question.

erlanger: Ok, improved comments.

JanWielemaker:
Merged the two commits and now only does this on Android.  I don't think it is a good idea to do this blindly.   Updated the comment to reflect this.  Lot of work for a small thingy ...

Issue #382, opened 2018-12-04
Error cross-compiling for aarch64


erlanger:
I am getting the following error when cross-compiling for Android **aarch64**:

```
-- Build files have been written to: /home/builder/.termux-build/swi-prolog/build
[194/830] Building C object src/CMakeFiles/libswipl.dir/pl-wam.c.o
FAILED: src/CMakeFiles/libswipl.dir/pl-wam.c.o
/home/builder/.termux-build/_cache/18-aarch64-21-v1/bin/clang --target=aarch64-none-linux-android --gcc-toolchain=/home/builder/.termux-build/_cache/18-aarch64-21-v1 --sysroot=/home/builder/.termux-build/_cache/18-aarch64-21-v1/sysroot -Dlibswipl_EXPORTS -Isrc -I/home
/builder/.termux-build/swi-prolog/src/src -I/data/data/com.termux/files/usr/include -Oz -I/data/data/com.termux/files/usr/include -Os -DNDEBUG -fPIC   -Wall -MD -MT src/CMakeFiles/libswipl.dir/pl-wam.c.o -MF src/CMakeFiles/libswipl.dir/pl-wam.c.o.d -o src/CMakeFiles/l
ibswipl.dir/pl-wam.c.o   -c /home/builder/.termux-build/swi-prolog/src/src/pl-wam.c
In file included from /home/builder/.termux-build/swi-prolog/src/src/pl-wam.c:225:
/home/builder/.termux-build/swi-prolog/src/src/pl-alloc.c:1044:2: error: "Unsupported int64_t alignment conversion"
#error "Unsupported int64_t alignment conversion"
 ^
1 error generated.
[201/830] Building C object src/CMakeFiles/libswipl.dir/pl-comp.c.o
ninja: build stopped: subcommand failed.
```

Any idea why this could be happening?

erlanger:
Strangely I get the same error if I cross-compile for Android **x86_64**:

```
-- Build files have been written to: /home/builder/.termux-build/swi-prolog/build
[141/830] Building C object src/CMakeFiles/libswipl.dir/pl-wam.c.o
FAILED: src/CMakeFiles/libswipl.dir/pl-wam.c.o
[...]

PR #383, opened 2018-12-06
PORT: calculate alignment at compile time

erlanger:
Fix for #382 
It calculates the alignment of `void*`, `double` and `int64_t` at compile time. Which means it can be used when cross-compiling without the need for an emulator.

### DIFF
--- a/cmake/AlignOf.cmake
+++ b/cmake/AlignOf.cmake
@@ -1,51 +1,42 @@
-MACRO(ALIGNOF TYPE LANG NAME)
+# Define variables for alignment of int64, void*, and double
 
-  IF(NOT ${NAME})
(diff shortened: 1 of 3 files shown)


JanWielemaker:
Thanks!  Works well, but needs to be cached.  Will you do so or should I?

erlanger:
> Thanks! Works well, but needs to be cached. Will you do so or should I?

Done!
Enjoy

JanWielemaker:
Thanks.  Merged after rebase.  Will you send a notification if SWI-Prolog is available as Termux package?

erlanger:
Sure, but the people at termux seem to be very slow, I've seen a PR there sitting for months or even a year.

PR #394, opened 2018-12-24
Fix pthread_setname_np() usage on NetBSD.

iamleot:
When compiling on NetBSD:

    .../swipl-devel/src/pl-thread.c: In function ‘set_os_thread_name_from_charp’:
    .../swipl-devel/src/pl-thread.c:1642:27: warning: passing argument 1 of ‘pthread_setname_np’ from incompatible pointer type [-Wincompatible-pointer-types]
       if ( pthread_setname_np(name) == 0 )
                               ^~~~
    In file included from .../swipl-devel/src/pl-mutex.h:74:0,
                     from .../swipl-devel/src/pl-builtin.h:67,
                     from .../swipl-devel/src/pl-incl.h:86,
                     from .../swipl-devel/src/pl-thread.c:52:
    /usr/include/pthread.h:160:5: note: expected ‘pthread_t {aka struct __pthread_st *}’ but argument is of type ‘char *’
     int pthread_setname_np(pthread_t, const char *, void *);
         ^~~~~~~~~~~~~~~~~~
    .../swipl-devel/src/pl-thread.c:1642:8: error: too few arguments to function ‘pthread_setname_np’
       if ( pthread_setname_np(name) == 0 )
            ^~~~~~~~~~~~~~~~~~
    In file included from .../swipl-devel/src/pl-mutex.h:74:0,
                     from .../swipl-devel/src/pl-builtin.h:67,
                     from .../swipl-devel/src/pl-incl.h:86,
                     from .../swipl-devel/src/pl-thread.c:52:
    /usr/include/pthread.h:160:5: note: declared here
     int pthread_setname_np(pthread_t, const char *, void *);
         ^~~~~~~~~~~~~~~~~~

In NetBSD pthread_setname_np(3) has also a third argument, the 2nd
argument `name` is a printf(3) format string, the 3rd argument
`arg` is a printf(3) argument.

This pull request add a check for it in `cmake/Config.cmake`, update
`src/config.h.cmake` and adjust `src/pl-thread.c`. While here
`HAVE_PTHREAD_SETNAME_NP_WITHOUT_TID` was removed and
`HAVE_PTHREAD_SETNAME_NP_WITH_TID` and
`HAVE_PTHREAD_SETNAME_NP_WITH_TID_AND_ARG` macro are used instead (the fallback
case is always `pthread_setname_np(name)`).


JanWielemaker:
Thanks.  Merged after editing the pthread_setname_np tests.  These must use `-Werror` otherwise they all pass

PR #396, opened 2018-12-24
Define {fopen,ftello,fseeko}64 functions for NetBSD

iamleot:
Similar to FreeBSD and OpenBSD, NetBSD does not have
{fopen,ftello,fseeko}64 functions.


JanWielemaker:
Applied, but github doesn't see it after the rebase.

PR #397, opened 2018-12-24
Remove unused QSORT_R_WITH_NESTED_FUNCTIONS sort_r() implementation.

iamleot:
After the migration to cmake QSORT_R_WITH_NESTED_FUNCTIONS was no longer
defined.  Apart that, nested functions are not standard and previously in
operating systems where PaX MPROTECT, W^X or similar security features are
supported SWI-Prolog built successfully but then fail at run time due
that.

While here also update comment describing the current versions.


JanWielemaker:
Thanks for all this!  Will probably review and merge next week.

iamleot:
Hello Jan,

Jan Wielemaker writes:
> Thanks for all this!  Will probably review and merge next week.

No hurry and thank you!


Happy holidays!

JanWielemaker:
Applied, but github doesn't see it after the rebase.  Thanks for all the patches!

PR #410, opened 2019-01-16
Allow building in Haiku

aarroyoc:
This small changes allow SWI Prolog to run on Haiku, which mostly behaves as a POSIX system. I still have problems with some packages, but core functionality works with this PR.

screenshot from 2019-01-16 21-00-44


erlanger:
It will be much cleaner if you do something like:

```c
#ifdef __HAIKU__
#define __unix__
#endif
```

Issue #414, opened 2019-01-20
Cross-running saved state on android - shared library names


erlanger:
What function/file loads the shared libraries when running a saved stated with the `-x` flag for `PL_initialise()` argv? 

I am having a problem running a host-built saved state on the android target device. I build the saved state on the 64-bit development host machine; then I run the apk builder (which I am working on) which produces a packaged installable apk file that includes the saved state along with other needed files. I am able to load and mostly run the saved state, but I have a problem:

The saved state  is unable to load the shared libraries `libpthread.so.0` and `libc.so.6`. The reason is that on android those libraries are called `libc.so` and `libpthread.so` and they are located in `/system/lib`, but I want to run the saved state which was built on the host machine, and _there_ the shared library names are `libc.so.6` and `libpthread.0` and they are in `/usr/lib`. Remember here there is no `swipl` executable running:  I am running directly from JPL  (which loads  `libswipl.so`) and then runs `PL_initialise(...)` within the android app.

I was thinking about modifying the code that loads the shared libraries to make it try library names without the number suffix if the original name fails. Perhaps also try different directory names, if it is needed. So if `libc.so.6` fails try `libc.so`, if that fails try  `/system/lib/libc.so`.... and so on.

Can you point me to the code/function that loads shared libs for the saved state (couldn't find it after a quick search)? 
Or perhaps  you have a better idea/approach on how to solve it? 
Any other gotchas when cross-running a saved state (besides the 32bit/64bit issue, and of course shared library compatibility)?

JanWielemaker:
I don't have time to insert all I know right here and now, but these are some observations

  - Extension (e.g. socket.so) and loaded eventually through dlopen(), called from open_shared_object/2 through `library(shlib)`.
  - If Java loads a shared object it does so normally using `RTLD_LOCAL`, which means that the Prolog core symbols are not added to the global symbol table and subsequent `dlopen()` for extensions go wrong.  I don't know whether Java provides a way around.  For toplevel apps, a way around is to preload Prolog using the `LD_PRELOAD` environment variable, but this won't work here (I fear).   Possibly we can dlopen() libswipl.so ourselfs with `RTLD_GLOBAL`?
  - Dev environments bind to *.so, which is a symlink to *.so.<version>.  The resulting binary binds to
[...]

PR #417, opened 2019-01-23
Cross-architecture shlib support for saved states

erlanger:
Okay, this is a working version supporting multi-architectures for shared libraries in the saved state.
What is missing now is to add some tests to the test suite, but wanted to submit the PR now that you made the intermediate code independent of word size. It is looking good!!

Let me know what you think. This is what is implemented (from the documentation included in the PR):

**foreign(+Action)**
If save, include shared objects (DLLs) for the current architecture into the saved state. See `current_foreign_library/2`, and `current_prolog_flag(arch, Arch)`. If the program strip is available, this is first used to reduce the size of the shared object. If a state is started, `use_foreign_library/1` first tries to locate the foreign resource in the resource database. When found it copies the content of the resource to a temporary file and loads it. If possible (Unix), the temporary object is deleted immediately after opening.

If Action is of the form `arch(ListOfArches)` then the shared objects for the specified architectures are stored in the saved state. On the command line, the list of architectures can be passed as `--foreign=\<CommaSepArchesList\>`. In order to obtain the shared object file for the specified architectures, `qsave_program/2` calls a user defined hook: `qsave:arch_find_shlib(+Arch, +FileSpec, -SoPath)`. This hook needs to unify `SoPath` with the absolute path to the shared object for the specified architecture. `FileSpec` is of the form foreign(Name).

At runtime, SWI-Prolog will try to load the shared library which is compatible with the current architecture, obtained by calling `current_prolog_flag(arch, Arch)`. An architecture is compatible if one of the two following conditions is true (tried in order):

1. There is a shared object in the saved state file which matches the current architecture name (from `current_prolog_flag/2`) exactly.
2. The user definable `qsave:compat_arch(CurrentArch, SoArch)` hook succeeds.

This last one is useful when one wants to produce one shared object file that works for multiple architectures, usually compiling for the lowest common denominator of a certain CPU type. For example, it is common to compile for armv7 if even if the code will be running on newer arm CPUs. It is also useful to provide highly-optimized shared objects for particular architectures.

Discussion in #414

### DIFF
--- a/library/qsave.pl
+++ b/library/qsave.pl
@@ -54,6 +54,14 @@ also used by the commandline sequence below.
 :- meta_predicate
     qsave_program(+, :).
 
+:- multifile error:has_type/2.
+error:has_type(qsave_foreign_option, Term) :-
+    is_of_type(oneof([save, no_save]), Term),
+    !.
+error:has_type(qsave_foreign_option, arch(Archs)) :-
+    is_list(Archs),
+    !.
+
 save_option(stack_limit, integer,
             "Stack limit (bytes)").
 save_option(goal,        callable,
@@ -74,7 +82,7 @@ save_option(stand_alone, boolean,
             "Add emulator at start").
 save_option(emulator,    ground,
             "Emulator to use").
-save_option(foreign,     oneof([save,no_save]),
+save_option(foreign,     qsave_foreign_option,
             "Include foreign code in state").
 save_option(obfuscate,   boolean,
             "Obfuscate identifiers").
@@ -137,11 +145,15 @@ qsave_program(FileBase, Options0) :-
     save_resources(RC, SaveClass),
     lock_files(SaveClass),
     save_program(RC, SaveClass, Options),
-    save_foreign_libraries(RC, Options),
-    zip_close(RC, [comment("SWI-Prolog saved state")]),
-    '$mark_executable'(File),
-    close_map,
-    cleanup.
+    setup_call_cleanup(
+        true,
+        save_foreign_libraries(RC, Options),
+        ( zip_close(RC, [comment("SWI-Prolog saved state")]),
+          '$mark_executable'(File),
+          close_map,
+          cleanup
+        )
+    ).
 
 cleanup :-
     retractall(saved_resource_file(_)).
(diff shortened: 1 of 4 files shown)


JanWielemaker:
Looks pretty close.   Hope to find time tomorrow or Friday to check the details and test.  Thanks!

erlanger:
Thanks!

library/qsave.pl
```diff
@@ -870,30 +882,50 @@
...
-    absolute_file_name(FileSpec,
-                       [ file_type(executable),
-                         access(read),
-                         file_errors(fail)
-                       ], File),
-    !,
+find_foreign_library(Arch, FileSpec, shlib(Arch,Name), SharedObject, Time) :-
+    FileSpec = foreign(Name),
+    % Call user hook to obtain shared library for architecture
+    (   ( catch( arch_find_shlib(Arch, FileSpec, File),
```
erlanger: I thought you renamed this to `arch_find_shlib`?

JanWielemaker: Only the user hook.  I don't care too much about the internal name to change it (as it also needs to be different).

erlanger: Where do you call `arch_shlib` (the user hook)?

JanWielemaker: Line 999

erlanger:
Thanks! Appreciate the tips also, they help me to improve!

library/qsave.pl
erlanger: ahh I forgot to mention, `runtime.doc` still has the old name `arch_find_shlib`

JanWielemaker:
You are doing pretty well.  Many of my comments are personal preferences rather than _this is wrong_.
I try to keep the sources as consistent as possible when it comes to style though.

erlanger:
ahh I forgot to mention, `runtime.doc` still has the old name `arch_find_shlib` ...
 
As for the tips, the more you put the better!!  It really helps me.

erlanger:
should I sumbit a PR with the runtime.doc fixes for the hook name or would you do it?

JanWielemaker:
Good you ask.  I was a bit afraid we'll do the same job.  I pushed a patch.  I think we are done.

erlanger:
Thanks!

PR #428, opened 2019-01-28
fix strip_failed printed even if strip succeeds

erlanger:
The `strip_failed` warning was printed even when strip succeeded.


erlanger:
Thanks!

PR #430, opened 2019-02-12
FIXED: return exit code if swipl -c fails

erlanger:
I separated this PR from the other, since it is a separate bug.

### DIFF
--- a/boot/init.pl
+++ b/boot/init.pl
@@ -1871,6 +1871,7 @@ load_files(Module:Files, Options) :-
     '$must_be'(list, Options),
     '$load_files'(Files, Module, Options).
 
+:- dynamic '$load_file_exception'/1.
 '$load_files'(X, _, _) :-
     var(X),
     !,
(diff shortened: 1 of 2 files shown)


JanWielemaker:
I have merged the patch to toplevel.pl (after adding a message for failure).  I don't like load_files/2 to load all files and then give an exception.   A correct implementation would also require `thread_local` rather than `dynamic` to avoid a thread to raising the exception caused by another thread ...  

I do think that the overall strategy for dealing with compiler errors requires rethinking.  Many people have asked for that over the years.  What we need is a flag that controls what happens with compilation errors, probably depending on the severity (warning vs error) and that controls whether the message is printed or propagated as an exception, aborting the compilation.  A possible third option would be to print them and raise a generic exception when the outermost `load_files/2` completes.

Unfortunately, this also requires a serious rewrite of the compiler frontend to use `setup_call_cleanup/3` to avoid messing up the global state and leaking resources.  This would be a great improvement for more reasons though.    Should you get excited, please make sure the arguments to `setup_call_cleanup/3` are as much as possible calls to a single predicate.  Using hugely complex control structures makes debugging typically a lot harder and doesn't really add to the readability.   Of course, there are exceptions ....

I _think_ this closes this one, no?

erlanger:
Good, thanks!  Yes, I should have thought about thread_local. I thought you wanted load files to
load all files, because even if one load failed you would want the others loaded (i.e. the loading should not depend on the order of the arguments). 

Yes, it does look like some major rewrite is due, but that will be for another time 

Closing issue

PR #445, opened 2019-03-11
PORT: Android minizip USE_FILE32API

erlanger:
Fix 'fopen64' not found in termux.

### DIFF
--- a/cmake/AndroidTargets.cmake
+++ b/cmake/AndroidTargets.cmake
@@ -1,6 +1,4 @@
 # Set target options for Android
 
-# Minizip needs USE_FILE32API set in 32-bit architectures
-if(CMAKE_SIZEOF_VOID_P EQUAL 4)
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Looks a little weird to me, but as you are in control of Android I assume you know what you are doing.

erlanger:
> Looks a little weird to me, but as you are in control of Android I assume you know what you are doing.

Some 64bit versions of Android don't provide 'fopen64' in libc.so, this fixes swi-prolog for those users. 

`fopen64` is not needed because In all 64bit platforms `off_t` is 64 bits  and `fopen` supports large files. So `fopen64` is not needed. In addition, `O_LARGEFILE` has been always inserted silently by android.

32 bit versions always use `fopen`.

 See [here](https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md) if you want more information.

PR #451, opened 2019-03-27
Add xref_called/5 which includes caller's line number

jamesnvc:
In order to facilitate finding the exact line number that a given predicate was called from via `xref_called`, include the `current_line` in the `called` functor.

This gives the line of the particular clause of the calling predicate, which then makes it possible for callers to walk the body term to find the call exactly. Obviously it would be preferable for `xref_called` to return that particular line number exactly, but that seems to require keeping subterm positions through the entire sequence of `process_*` predicates, which seems like it would be a substantial change.

This also will change the number of solutions reported by `xref_called/3` and `xref_called/4`, since the same caller can now be repeated multiple times.

example of the behaviour change; given the below code (in `test.pl`):

```prolog
 1 :- module(test, []).
 2 
 3 foo(X) :-
 4   bar(X).
 5 foo(X) :-
 6   true,
 7   bar(X).
 8 
 9 bar(X).

```

Previously:

```prolog
?- Src = '/home/james/tmp/test.pl',
    xref_source(Src), 
    findall(By, xref_called(Src, bar(_), By, _), Callers).
Src = '/home/james/tmp/test.pl',
Callers = [foo(_1834)].
```

New behaviour: 

```prolog
?- Src = '/home/james/tmp/test.pl',
    xref_source(Src), 
    findall(By-Loc, xref_called(Src, bar(_), By, _, Loc), Callers).
Src = '/home/james/tmp/test.pl',
Callers = [foo(_2498)-3, foo(_2482)-5].
```

### DIFF
--- a/library/prolog_xref.pl
+++ b/library/prolog_xref.pl
@@ -39,6 +39,7 @@
             xref_source/2,              % +Source, +Options
             xref_called/3,              % ?Source, ?Callable, ?By
             xref_called/4,              % ?Source, ?Callable, ?By, ?Cond
+            xref_called/5,              % ?Source, ?Callable, ?By, ?Cond, ?Line
             xref_defined/3,             % ?Source. ?Callable, -How
             xref_definition_line/2,     % +How, -Line
             xref_exported/2,            % ?Source, ?Callable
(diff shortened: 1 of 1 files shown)


JanWielemaker:

https://swi-prolog.discourse.group/t/get-line-numbers-from-xref-called-3/448/6

JanWielemaker:
Nice improvement!  Merged after updating comment, commit message, adding distinct to xref_called/4 to remain compatible and fix two missing locations for the additional argument.

PR #457, opened 2019-04-05
FreeBSD fixes

mifigiel:
I built the system on FreeBSD 12.0-RELEASE and there were few minor problems:
 * missing headers with definititions of CPU_ZERO(); cpu_set_t vs cpuset_t (fixed in ec92f5c)
 * including malloc.h on FreeBSD doesn't work, so I removed the check defining HAVE_MALLOC_H for that platform (43a0705). It's probably not a good idea to fix it only for one platform as I did...I've not checked all the code, but I assume that it's possible to replace malloc.h with stdlib.h everywhere.
 * the timezone from time.h is not a variable on FreeBSD - it's a function; I replaced the check_symbol_exists with check_c_source_compiles, that should avoid any false positives (243d228) 

There are a few compiler warnings concerning the packages (xpce, ssl, readline), but nothing stopping the build. 

One test fails: zlib:zlib gzip_utf8 and I've got no idea why (well, not yet :-))


JanWielemaker:
Thanks.  I'm in favor of checking for `sys/cpuset.h` and `pthread_np.h` and possibly the type.  Could you do that, or shall I git it a try (with a round trip to check whether it works)?

mifigiel:
I can do that, but it has to wait until tomorrow.

JanWielemaker:
Thanks.  I'm not in a big hurry.

mifigiel:
I  built  on FreeBSD 12.0-RELEASE and  Fedora Linux 29  without any problems.

JanWielemaker:
Perfect!  The conflict is because I rebased locally.  I prefer to manage PRs locally before pushing them to master.  The patches are applied without modification.  Thanks.

PR #461, opened 2019-04-08
PORT: fixes 91f03daf6, FreeBSD headers

mifigiel:
I made a mistake:
contrary to my claim in 91f03daf6, pthread.h doesn't pull in the files
necessary to consume sys/cpuset.h, but sys/param.h does.
I'm sorry! I still don't know how could the test build on FreeBSD be successfull...


JanWielemaker:
> I'm sorry!

No worries.  I never make mistakes :smiley:

PR #471, opened 2019-05-07
Fix incorrect argument name

pjhenning:
Flag -DNATIVE_FRIEND is not used by the cmake build, there is only -DSWIPL_NATIVE_FRIEND.


JanWielemaker:
Thanks.  There was one more.  Amended the patch, which isn't picked up by Github.

Issue #507, opened 2019-09-20
Android: [FATAL ERROR: Could not find system resources]


erlanger:
Some people are having a strange problem in some android devices, 
see this issue: https://github.com/erlanger/swipl-termux/issues/5

The error is  [FATAL ERROR: Could not find system resources]
but `boot.prc`  file is found as the strace shows:

```
openat(AT_FDCWD, "/data/data/com.termux/files/usr/lib/swipl/boot.prc", O_RDONLY|O_LARGEFILE) = 3
fstat64(3, {st_mode=S_IFREG|0600, st_size=85604, ...}) = 0
close(3) = 0
gettimeofday({tv_sec=1568916005, tv_usec=857728}, NULL) = 0
openat(AT_FDCWD, "/dev/properties/u:object_r:system_prop:s0", O_RDONLY|O_LARGEFILE|O_NOFOLLOW|O_CLOEXEC) = 3
fstat64(3, {st_mode=S_IFREG|0444, st_size=131072, ...}) = 0
mmap2(NULL, 131072, PROT_READ, MAP_SHARED, 3, 0) = 0xb0937000
close(3) = 0
openat(AT_FDCWD, "/data//misc/zoneinfo/current/tzdata", O_RDONLY|O_LARGEFILE) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/system//usr/share/zoneinfo/tzdata", O_RDONLY|O_LARGEFILE) = 3
read(3, "tzdata2016g\0\0\0\0\30\0\0w\274\0\7I\270", 24) = 24
lseek(3, 24, SEEK_SET) = 24
read(3, "Africa/Abidjan\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0"..., 30628) = 30628
lseek(3, 395734, SEEK_SET) = 395734
read(3, "TZif2\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\7\0\0\0\7\0\0\0\0"..., 3661) = 3661
close(3) = 0
openat(AT_FDCWD, "/data//misc/zoneinfo/current/tzdata", O_RDONLY|O_LARGEFILE) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/system//usr/share/zoneinfo/tzdata", O_RDONLY|O_LARGEFILE) = 3
read(3, "tzdata2016g\0\0\0\0\30\0\0w\274\0\7I\270", 24) = 24
[...]

PR #508, opened 2019-09-21
PORT: Android minizip to use consistent file api

erlanger:
Hopeful Fix for  #507; System resources not found error. This error occurs only on some very specific android systems. So far only on android 7.1.1, armv7l, 32bit.

### DIFF
--- a/cmake/AndroidTargets.cmake
+++ b/cmake/AndroidTargets.cmake
@@ -1,4 +1,4 @@
 # Set target options for Android
 
-# Minizip needs USE_FILE32API on Android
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Good catch!  Thanks

erlanger:
@Jan, 
do you think you can cut a release? This bug seems to have affected more versions of android than what I thought. So if it is possible to cut a release, I can ask the termux people to update.

JanWielemaker:
Most likely in a few days.

erlanger:
Thanks!

PR #511, opened 2019-10-04
Add snap package

LyzardKing:
First attempt at a working snap.
I haven't looked at the PGO build yet, but I'll test it soon


snap/snapcraft.yaml
```diff
@@ -0,0 +1,106 @@
+name: swipl
+base: core18
+adopt-info: swi-prolog
+license: 'BSD-2-Clause'
+summary: ISO/Edinburgh-style Prolog interpreter
```
JanWielemaker: Could you change _interpreter_ that to _environment_ or _implementation_, _system_, ...?

LyzardKing: Sure, I just copied this from the debian package in apt...

snap/snapcraft.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - libarchive13
+      - libgmp10
+      - libossp-uuid16
+      - libssl1.1
+      - ca-certificates
+      - libdb5.3
+      - libpcre3
+      - libedit2
+      - libgeos-c1v5
+      - libspatialindex4v5
```
JanWielemaker: These are only required for the spatial package, which is no longer bundled.   You could consider adding some of the packages as is also done for the Docker image.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - unixodbc
+      - odbc-postgresql
+      - tdsodbc
```
JanWielemaker: What is this?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - libmariadbclient18
```
JanWielemaker: This isn't the ODBC driver,  no?  We do not have a native mariadb interface.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - libsqlite3-0
+      - libserd-0-0
+      - libraptor2-0
```
JanWielemaker: Also these are only for packages (plsqlite and hdt)

snap/snapcraft.yaml
LyzardKing: Possibly some other mariadb dependency..I'll remove them.
What's needed to interface with a database?

JanWielemaker:
Added some comments.  Mostly lacking is probably a description on what you can do with this file.   Could you add e.g., a file SNAP.md?

snap/snapcraft.yaml
LyzardKing: Same issue as tdsodbc, probably debian dependency for database integration..
I can remove it.

snap/snapcraft.yaml
LyzardKing: and gone..

snap/snapcraft.yaml
JanWielemaker: Should be changed there too.  I don't really like _interpreter_.  Seems a term from the past century.   SWI-Prolog (and virtually all Prolog's) depend on a virtual machine.

snap/snapcraft.yaml
LyzardKing: ODBC driver for FreeTDS

snap/snapcraft.yaml
JanWielemaker: The actual list of dependencies is on https://www.swi-prolog.org/build/Debian.txt

JanWielemaker: Oh, the more odbc drivers, the better I assume.  The link only gives what is needed to compile the system (with ODBC), but without a driver the ODBC package is pretty much meaningless.

snap/snapcraft.yaml
LyzardKing: I agree. I personally use implementation, but system seems to be more future proof looking at all the non standard or simply new additions

snap/snapcraft.yaml
LyzardKing: I started from that list, but I had to add a few, so I ended up looking at the deb package..
(probably not a great idea in retrospect)

LyzardKing:
How does swipl output audio? does it work with pulseaudio? or does it need something different?

JanWielemaker:
> How does swipl output audio

Not by itself.  There is a pack interfacing to OSC by Samer Abdallah.   Has been recently discussed on Discourse.  I know little about sound.

JanWielemaker:
Please let me know when you think you are ready; I lost track a little ...

LyzardKing:
I'm looking at the PGO build. That breaks the snapcraft defaults a bit..
I think I'm on the right track, but I'm testing things now.

JanWielemaker:
If you happen to be a CMake expert you might find a clean way to do it in CMake and get rid of this ugly script.  I'd welcome that!

JanWielemaker:
> Not really unfortunately

Bad luck :(

Yes, you need `libyaml-dev`.  `qtbase5-dev` is used by the Qt front end.  I think you can use

    -DSWIPL_PACKAGES_QT=OFF

to stop looking for these packages.

LyzardKing:
I wonder if we should add the frontend, it would increase the size of the snap, but it would give a default ide.
What do you think? It can always be imported later on I guess..

JanWielemaker:
The nice thing is that you can create a desktop icon people can click and get an interface with some menus to get at the tools.   The problem is that with some experience, they either want to run SWI-Prolog in their IDE (Eclipse, Emacs, ...) or on a plain terminal.

So, I don't know what the right balance is.  You can try and ask on Discourse.  The Qt tool is also used for the MacOS binaries.

LyzardKing:
One does not exclude the other..
My personal opinion is that if the size doesn't change too much then it's a nice convenience and inclusive measure to have the cli and gui program, and leave the choice of which one to launch to the user

LyzardKing:
Do I have to explicitly build the qt gui or is it in the main cmake build?

JanWielemaker:
It is in the main CMake build.  You only need Qt5.  I only wonder whether the list of Qt5 packages that is requires is written down somewhere.  Can't find it directly.

LyzardKing:
The snap with the qt gui is up to 95MB total.
Still good I think.
Most of it is taken up by some locale data and qt libraries, but it seems to still work well.
The thing that's missing now is to set the architecturet to build for.
The supported ones on build.snapcraft.io are: 

- [ ] amd64
- [ ] arm64
- [ ] armhf
- [ ] i386
- [ ] ppc64el
- [ ] s390x

JanWielemaker:
> Are there some that are not supported?

I think they all work fine.  Not all are very actively maintained though, so sometimes a little work is needed.  Well, this might be a good opportunity to keep them all alive.

LyzardKing:
`error while loading shared libraries: libswipl.so.8: cannot open shared object file: No such file or directory`
Unfortunately something's not working...
Is there a reason why the swipl folder structure is different from the linux one?
usually the libraries are in {/lib, /usr/lib, /usr/local/lib}/x86_64-linux-gnu
Then the paths can be rebuilt like in these helper functions https://github.com/ubuntu/snapcraft-desktop-helpers/blob/38e6ba960534ac8876bcbcf22ca769d60bdce01b/common/init#L47

LyzardKing:
How does swipl determine the architecture?

LyzardKing:
The java integration might be a bit complicated...
Not having used it I'm not sure.. If I compile the support with java8, do I need to use it only with a java8 jdk later on?
It  might be easier in a classic non confined snap.
On an unrelated matter, did the PL_initialise api method change since 8.0.3? it's used in pyswipl, but now it prints the --help message instead at that point

JanWielemaker:
> `error while loading shared libraries: libswipl.so.8: cannot open shared object file: No such file or directory`
> Unfortunately something's not working...

Use CMake option `DSWIPL_INSTALL_IN_LIB=ON` to avoid that.  I forgot about this, but there are two types of .so files in there.  There is libswipl.so that contains the system and is linked into `swipl` and thus must be in LD_LIBRARY_PATH.  That is the only one.  The others are plugins that are loaded using dlopen() when needed and located by Prolog itself.   The above option installs libswipl.so in the normal OS location.    That should allow you to stop playing with `arch`.

What I do not understand is that the normal build use RPATH on the executables to make them find libswipl.so.  Why doesn't that work?

LyzardKing:
It works with the DSWIPL_INSTALL_IN_LIB=ON option!
Is it worth it to add DSWIPL_INSTALL_DIR? The debian build adds it to set a folder in /usr/lib

JanWielemaker:
> DSWIPL_INSTALL_DIR

I don't think so.  It is mostly to work around packaging differences as far as I recall.

LyzardKing:
I added a desktop launcher for the qt ui, using the icon from the windows swipl-win
The issue with pyswip I mentioned earlier was the -nosignal option, that is now changed to --nosignals

JanWielemaker:
Trying to create a snap.  Install `snapcraft` from snap (version 3.8) on Ubuntu 19.04.  Got your branch and ran `snapcraft` in the Prolog top dir.  Is that right?   Anyhow, after resizing my root partition ..., I get:

```
Pulling swi-prolog 
Cloning into '/root/parts/swi-prolog/src'...
done.
Submodule 'bench' (/root/project/bench.git) registered for path 'bench'
...
fatal: repository '/root/project/bench.git' does not exist
fatal: clone of '/root/project/bench.git' into submodule path '/root/parts/swi-prolog/src/bench' failed
```

I guess this is because it is trying to make a recursive clone from the local git repo and this doesn't work due to the relative paths.   You must have worked around this.  How?

LyzardKing:
I thought it would work once upstream...if not the "tweak" is to change the line
`source: .`
to
`source: https://github.com/SWI-Prolog/swipl-devel.git`

LyzardKing:
It works on my local clone because I initialized the submodules

JanWielemaker:
If we can overrule the git clone --recursive it might work.
> I initialized the submodules

How?

LyzardKing:
In theory the git part of snapcraft (source-type: git) should do that automatically..
I'm not sure why it doesn't work in this case.
I just ran git clone --recurse-submodules I believe (I can't really remember...)

LyzardKing:
On the automated build system it should work automatically...
if not we can try and add the override, it should be:
`git submodule update --init --recursive --force`

JanWielemaker:
> git clone --recurse-submodules

That doesn't work for the same reasons.  After the initial clone the relative URLs for the submodules in `.gitmodules` must be updated before initializing the submodules.  If we can run something else than git clone --recursive we can do that.

> --force

I don't that will help.  You cannot force the wrong source location ...

JanWielemaker:
B.t.w. using the `https` it worked and I have a snap.  I wonder whether this is not the right thing to do in the first place?

JanWielemaker:
It runs!  96Mb may be a bit large, but let us see what people think.   Only, the executable for swipl-win is called `swipl.swipl-win`

LyzardKing:
Yes, the url is the simplest solution...the issue is simply that you have to pull the swipl code that you already downloaded...but it's not an issue.
96MB is good for me with the qt ui, but there might be some other things that can be removed..
Once it's up on the store you can ask  for an official alias so it'll be swipl-win, by default they have to be all associated with the snap name

LyzardKing:
that's why I called the snap swipl in the first place, if the snap was called swi-prolog then the swipl binary would have been swi-prolog.swipl. (again, fixable on the store with an alias)
It's up to you what name you want.
You also need to register a user on the snap store and the snap name

JanWielemaker:
Ok.  It is time to do some cooking for me.  I'll run some tests and try to get it uploaded tomorrow.   No promise, but surely in a few days.

LyzardKing:
:tada:
Let me know if there's anything I can do!

JanWielemaker:
I think I'm fine for now.   You have been of great help.  Ask me for a place on the [contributors](https://www.swi-prolog.org/Contributors.html) list if you want to be on there.

LyzardKing:
I can set the aliases in the snapcraft.yaml now, so once it's pushed to the store it'll be manually reviewed by someone from canonical.
And in the meantime the aliases can be manually enabled.
So for consistency I could set the snap name to swi-prolog, and the aliases to swipl and swipl-win, what do you think?
My contribution is extremely minimal to be added to a contributors list, but it would be nice! Thanks!

JanWielemaker:
> So for consistency I could set the snap name to swi-prolog, and the aliases to swipl and swipl-win, what do you think?

Yes, please!

LyzardKing:
Another thing that could be done is this:
Add all the dependencies that are needed by the non qt binary in the snap.
Then whoever wants the qt ui can install the shared qt/kde base (used by the kde snaps).
That snap is a lot larger, but it is shared by multiple others, and the "simple" swi-prolog snap in itself would be smaller.
I'm also trying to write a small alert window in xpce to tell the user that they have to install the base snap if they try to launch the .desktop file. But I know nothing about xpce...

JanWielemaker:
Do you say that the snap doesn't (only) run against the base it was configured with?    Wouldn't it be easier to use qt/kde as base anyway in hat case?   That keeps things really easy and snap users probably get it sooner or later anyway (most likely).

P.s. I found a serious issue: `/snap/bin/swipl -g emacs_serve` crashes with a segv.  Basic xpce seems to work, but trying `?- emacs.` hangs for a while and than crashes (same segv).  

P.s. Considering init/cache, this seems to be a good moment to consider moving to the XDG conventions.

LyzardKing:
If we remove everything that's in the content snap (the kde snap) the swi-prolog snap falls to 20 MB.
If we assume that people might not want to install the content snap, then I guess it'll still be roughly at 45MB?
The content snap itself is 273MB, but again, it'll likely be installed automatically if you use kde snaps.

LyzardKing:
I tried `?- emacs.` and it works
Maybe something changed in my snap..
emacs_serve is an unknown procedure?

JanWielemaker:
This gives the size of the (unpacked) PPA binaries:

```
du -sh /usr/lib/swi-prolog/
34M	/usr/lib/swi-prolog/
```

JanWielemaker:
> I tried `?- emacs.` and it works

Not sure that makes me more or less happy ...

> Maybe something changed in my snap..
> emacs_serve is an unknown procedure?

Sorry, `emacs_server` (forgot final `r`)

LyzardKing:
ok, so
20MB withouth the libraries contained in the kde content snap
58 MB with the libraries needed for the cli version
The numbers are a bit off because it's squashfs compressed stuff...not really comparable to a real directory..

LyzardKing:
emacs_server just drops me in a swipl instance, is that correct?

JanWielemaker:
> emacs_server just drops me in a swipl instance, is that correct?

Yes.  It creates a unix domain socket and then there is a little shell script that allows you to open files in the editor.  Much like GNU emacs server mode,

LyzardKing:
Ah..ok. So that might need to be exposed in the snapcraft yaml... which script is it?
Any thoughts about the size difference and qt dependency?

JanWielemaker:
> Ah..ok. So that might need to be exposed in the snapcraft yaml... which script is it?

I'd not put that on the shortlist.  I guess I'm the only user :)

>  Any thoughts about the size difference and qt dependency?

I think I vote for using qt/kde as base.  Easy and largest reuse.

LyzardKing:
So qt/kde base for everything? even the swipl command.
Ok. I'm not sure if it can be automatically downloaded, but the first time it's run on the terminal it'll print this:
```sh
snap install kde-frameworks-5-core18
snap connect swi-prolog:kde-frameworks-5-plug kde-frameworks-5-core18:kde-frameworks-5-core18-slot
```

JanWielemaker:
> So qt/kde base for everything

Ideally you'd know how many snap users have this already.   I have little clue.  I'm a Gnome user, but typically I get some KDE application on the system soon enough.   Notably kcachegrind and heaptrack are indispensable.  From snap I now only use spotify.

LyzardKing:
It's impossible to get the real numbers, but it seems to me that the content snap is fairly used:
https://snapcraft.io/kde-frameworks-5-core18
We can always remove it and switch to the mixed solution if there are complaints in the future

JanWielemaker:
>  We can always remove it and switch to the mixed solution if there are complaints in the future

Agree

JanWielemaker:
PR has been squash-merged.   Further work will go on in a new PR.   Now going to try and build the snap!

JanWielemaker:
Hi,

I'm back from holiday.  I've been approved ownership of the snap name `swi-prolog`.  I've lost track a little though.   I recall there were issues with the kde-frameworks.  Has anything been improved there?

LyzardKing:
Hi! welcome back :)
I hope you enjoyed your holiday!
The kde framework issues should be solved soon, with the release of the new snapcraft version.
With that the kde-neon "content snap" will be autoconnected, so the user will not have to install/run any extra commands. It's not released yet, but the difference from the eisting implementation is minimal.
The locales should be fine, even if the command `locale -a` doesn't find them, the correct ones for the system should be configured when the snap starts (according to the forum, but to be sure I asked about the command).
I still can't get the PLLBIR variable to point to where libswipl.so is, but that's not a snap problem I suppose... it points to the dir with the other .so files

JanWielemaker:
Ok.  I guess you keep track of kde-neon and ping me when we are ready to move?

> I still can't get the PLLBIR variable to point to where libswipl.so is, but that's not a snap problem I suppose... it points to the dir with the other .so files

I understand you want the location of libswipl.so to load it directly into Python?   I think the schema is like this, assuming Python will eventually pass the file name you provide to `ldopen()`:

  - If `swipl --dump-runtime-variables`  provides `$PLLIBDIR`, this is `$PLLIBDIR/libswipl.$PLSOEXT`
  - Otherwise just use `libswipl.$PLSOEXT` as `ldopen()` searches in the standard OS locations and the file is installed in a standard OS location.

PR #513, opened 2019-10-07
Add kde-neon content snap

LyzardKing:
This adds the content snap extension
It'll become available in the next version of snapcraft, due to come out soon.
It seems to solve the font/locale issues
and takes care of the heavy liifting (opengl/extra libraries)

### DIFF
--- a/snap/local/snap.md
+++ b/snap/local/snap.md
@@ -4,17 +4,17 @@ Snap that runs the latest version of swipl and swipl-win.
 
 The snap is strictly contained, so it has limited access to the host system via the following plugs:
 
-| Plug              | Description               |
-| ----------------- | ------------------------- |
-| home              | $HOME (no hidden folders) |
-| network           | network access            |
-| x11               | I/O and graphics output   |
-| desktop           | common desktop elements   |
-| desktop-legacy    | legacy desktop elements   |
-| removable-media   | removable media           |
-| opengl            | opengl/gpu hardware       |
-| audio-playback    | audio playback devices    |
-| audio-record      | audio record devices      |
+| Plug              | Description               | Auto-connected |
+| ----------------- | ------------------------- | -------------- |
+| home              | $HOME (no hidden folders) | yes            |
+| network           | network access            | yes            |
+| x11               | I/O and graphics output   | yes            |
+| desktop           | common desktop elements   | yes            |
+| desktop-legacy    | legacy desktop elements   | yes            |
+| removable-media   | removable media           | no             |
+| opengl            | opengl/gpu hardware       | yes            |
+| audio-playback    | audio playback devices    | yes            |
+| audio-record      | audio record devices      | no             |
 
 If other connections are needed they can be added from this [list](https://snapcraft.io/docs/supported-interfaces).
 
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.  Now snapcraft says this:

```
Issues while validating snapcraft.yaml: The 'apps/swipl/extensions[0]' property does not match the required schema: 'kde-neon' is not one of ['gnome-3-28']
```
Is there something that should installed first?

LyzardKing:
yes, at the moment you need the unreleased version of snapcraft to build it..so:
`snap refresh --edge snapcraft`
then to revert back to stable
`snap refresh --stable snapcraft`

JanWielemaker:
Thanks.  After a `snapcraft clean` and rebuilding from scratch with `edge` I now get

```
Failed to generate snap metadata: Specified command 'swipl-win' was not found.
Verify the command is correct and for a more deterministic outcome, specify the relative path to the command from the prime directory.
```

LyzardKing:
try replacing the command swipl-win with $SNAP/usr/bin/swipl-win
and the command swipl with $SNAP/usr/bin/swipl

don't run snapcraft clean, it should not rebuild everything but replace only the commands

JanWielemaker:
I found out.  I get a warning during the build as below.  

```
Could NOT find Qt5Widgets (missing: Qt5Widgets_DIR)
```

Now the question is why.  Didn't see any warnings when trying to get KDE!?

LyzardKing:
mm..strange.
I just ran a snapcraft build succesfully, and installed with the qt ui working..

LyzardKing:
all the qt libraries needed by the build should be satisfied by
`build-snaps: [kde-frameworks-5-core18-sdk]`

JanWielemaker:
That was what I expected.  The I removed `swipl-win` from `apps` to see what happens.  Now it builds, but after installing I get

```
swi-prolog.swipl
realpath: '': No such file or directory
realpath: '': No such file or directory
realpath: '': No such file or directory
realpath: '': No such file or directory
realpath: '': No such file or directory
realpath: '': No such file or directory
realpath: '': No such file or directory
realpath: '': No such file or directory
```
and then it hangs.   I'm starting to doubt about all this.  Would it be an idea to revert to the stable core stuff and ditch swipl-win until snap and KDE have a good marriage?   We can leave this around in a branch.

LyzardKing:
I think we can keep this pr around (or close it and reopen once the new stable release is out).
In the meantime the issues you were having:
the fontconfig should not really be an issue, because it's using the user fonts, so the warnings can be ignored.
the locale issue: did you connect to the kde image? the locale is there

JanWielemaker:
> the fontconfig should not really be an issue, because it's using the user fonts, so the warnings can be ignored.

Not sure what you mean by _user fonts_.  xpce is actively using fontconfig.  In any case, seeing this many warnings when starting something is not acceptable.

> the locale issue: did you connect to the kde image? the locale is there

I could also see the locale files in the image, but `?- shell('locale -a').` only lists the C and POSIX locales.   Not working Xlib locale is probably fatal to xpce Unicode input (I don't have anything but an ASCII keyboard, so it takes a lot of effort to test).

LyzardKing:
> > the fontconfig should not really be an issue, because it's using the user fonts, so the warnings can be ignored.
> 
> Not sure what you mean by _user fonts_. xpce is actively using fontconfig. In any case, seeing this many warnings when starting something is not acceptable.
> 
in the snap xpce uses the version of fontconfig included in the snap, but that can see the fonts available in the host.
So if I install a particular font I can use it in my snaps.
I agree it's not acceptable, that's why the extension fixes it.


> > the locale issue: did you connect to the kde image? the locale is there
> 
> I could also see the locale files in the image, but `?- shell('locale -a').` only lists the C and POSIX locales. Not working Xlib locale is probably fatal to xpce Unicode input (I don't have anything but an ASCII keyboard, so it takes a lot of effort to test).

That's strange..I'll look into that. I agree it's an issue!

LyzardKing:
Apparently the `locale -a` command will not work in the snap, but the localization will.
https://forum.snapcraft.io/t/localisation-of-snaps/8425/6
Also, with the content snap the Xlib errors are gone, as are the fontconfig errors.
Once the new release of snapcraft is out the build will work properly.
I think the issues you were having were from building with multipass.
I find installing lxd and running with that to work better.
It might be an issue with the fact that it's still unreleased..but it looks like it can become better!

JanWielemaker:
Great!  I propose to wait till after my travels.  Hopefully these problems have been resolved by then   and I own `swi-prolog` as snap application name.   Thanks for all the work dine so far!

LyzardKing:
@JanWielemaker I rebased on master in another branch.
The readme will be changed also in the separate branch.
I'll close this and point to #569

PR #528, opened 2019-12-18
remove deprecated use_module directives

kamahen:


kamahen:
At line 242 (after the paragraph starting "The mode declation scheme is equivalent to XSB with partial"), there's some layout that I don't understand. 
```
Var
+
index
``` 

and later a "-" on a line by itself.

I haven't done anything with this because I don't know what it means.

JanWielemaker:
There is a missing `\nodescription`.  They are three alternatives to achieve the same: a normal tabled argument.   Added the \nodescription.

Thanks for the fixes.

PR #529, opened 2019-12-23
Add '-f none' to invocations of PROG_SWIPL, to avoid interference from init.pl

cmsmcq:
This was done for man/CMakeLists.txt on 8 October 2019, but the  "-f none" is not present in packages/cmake/PackageDoc.cmake (and caught me yesterday).

Note that there is still one invocation of PROG_SWIPL in cmake/QLF.cmake, which I have not changed since I am confident I don't know whether it needs changing.

This pull request is also an experiment to see if I can follow the instructions for submitting patches in usable form.

### DIFF
--- a/packages/cmake/PackageDoc.cmake
+++ b/packages/cmake/PackageDoc.cmake
@@ -4,7 +4,7 @@ function(doc2tex file)
   string(REPLACE ".doc" ".tex" tex ${file})
   add_custom_command(
       OUTPUT ${tex}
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Just to make sure, I've put the lines below in my `init.pl` and confirm all builds and tests cleanly now.

```
:- format(user_error, 'ERROR: lacking -f none~n', []).
:- halt(1).
```

PR #530, opened 2019-12-24
Copy edits

cmsmcq:
Some typo corrections and light copy edits to the documentation for built-ins, including adding leading zeroes in things that looked as if intended to be ISO 8601 dates, some braces in pl.bib to protect capitalization.

The package also includes a couple of bogus changes that result from my not knowing how to work correctly with submodules, in particular how to get all the updates before rebasing a branch to make it easier to apply.  If I knew how to exclude the commits that re-synched with the debian and packages/jpl submodules, I would.

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -2,7 +2,7 @@
 
 \section{Notation of Predicate Descriptions}	\label{sec:preddesc}
 
-We have tried to keep the predicat descriptions clear and concise.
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.  Merged.  Please try to avoid submodule updated in the PR.  As luckily you put them in separate commits it wasn't hard to fix this time.   It does cause the PR now marked as conflicting  though.   Just ignore that.

PR #531, opened 2019-12-25
High order tests

borisvassilev:
These are some _preliminary_ tests for library(dcg/high_order). I tried to test the most obvious behaviors.

There are a few things that were problematic for me.

First, I might be completely misunderstanding the idea of "nondet elements".

The test "sequence//2 order of solutions" is exactly as documented, but I am not completely clear about how this is useful.

The test "sequence//3 det element trailing sep" is the troublesome one.

The test "sequence//3 nondet element", well, I don't know what should happen. The current behavior is that it is nondet only for the last element, after the last separator. This is kinda documented but maybe a bit surprising?

Finally, should it be explicitly documented that sequence//5 cuts any choice points left by the element?

### DIFF
--- a/src/Tests/library/test_dcg_high_order.pl
+++ b/src/Tests/library/test_dcg_high_order.pl
@@ -17,6 +17,8 @@ start --> "(".
 
 end --> ")".
 
+:- use_module(library(dcg/basics), [string//1]).
+
 test('sequence//2 ground list',
      forall(member(t(Codes, Result, Rest),
                    [t(``, [], []),
@@ -67,7 +69,7 @@ test('sequence//3 det element',
     L == Result,
     R == Rest.
 
-test('sequence//3 det element trailing sep', fixme('Sep should not be consumed')) :-
+test('sequence//3 det element trailing sep', fail) :-
     phrase(sequence(a, sep, L), `a,`, R),
     L == [a],
     R == `,`.
@@ -80,8 +82,10 @@ test('sequence//3 det element trailing sep consumed silently', fail) :-
 test('sequence//3 separator only', fail) :-
     phrase(sequence(a, sep, _L), `,`).
 
-test('sequence//3 nondet element', fixme('Not sure what should happen')) :-
-    phrase(sequence(b, sep, _L), `b,b`, _R).
+test('sequence//3 nondet element', nondet) :-
+    phrase(sequence(string, sep, L), `b,b`, R),
+    L == [`b`, `b`],
+    R == [].
 
 test('sequence//5 det element',
      [forall(member(t(Codes, Result, Rest),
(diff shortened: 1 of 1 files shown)


JanWielemaker:
> First, I might be completely misunderstanding the idea of "nondet elements".

It means that the element is committed by the following separator or the terminating sequence (if  provided).  That allows for example for ``string(X)`` as element, where  `string//1` is from the dcg basics library and matches `.*` in a non-greedy way.

borisvassilev:
> > First, I might be completely misunderstanding the idea of "nondet elements".
> 
> It means that the element is committed by the following separator or the terminating sequence (if provided). That allows for example for `string(X)` as element, where `string//1` is from the dcg basics library and matches `.*` in a non-greedy way.

OK, this clears up some of my confusion. So basically all the "b" element tests are bogus at the moment?

JanWielemaker:
``sequence//3 det element trailing sep`` is hard to implement when combined with the idea that a separator commits the separator and preceding element.   The only way would be to match the separator twice, i.e., doing something like

    Element, \+ \+ Sep, !, (Sep -> [] ; [])

I'm not sure whether or not we should want that.    What is the use case?

``sequence//3 nondet element`` should produce either [b1,b1] or [b1,b2], no?

JanWielemaker:
> So basically all the "b" element tests are bogus at the moment?

It doesn't really cover the intend.  In that sense is might be better to use string//1 or something similar that matches a variable length string in a non-greedy way.

borisvassilev:
> `sequence//3 nondet element` should produce either [b1,b1] or [b1,b2], no?

This is indeed what it does at the moment. But this test case does not seem relevant. I'd have to clean up the tests. Using string//1 instead of my made up `b//1`.

borisvassilev:
> `sequence//3 det element trailing sep` is hard to implement when combined with the idea that a separator commits the separator and preceding element. The only way would be to match the separator twice, i.e., doing something like
> 
> ```
> Element, \+ \+ Sep, !, (Sep -> [] ; [])
> ```
> 
> I'm not sure whether or not we should want that. What is the use case?

I don't have an obvious use case :-) so I can change this from "fixme" to "fail"?

JanWielemaker:
> so I can change this from "fixme" to "fail"?

That is my proposal for now.   Al least this makes this case clear.

borisvassilev:
> > so I can change this from "fixme" to "fail"?
> 
> That is my proposal for now. Al least this makes this case clear.

OK, I pushed a commit and now all tests succeed. Please feel free to fix anything else you think needs fixing, including commit messages.

JanWielemaker:
Thanks.  Merged after adding a header and squashing the commits.   This at least makes it clearer what it should do and helps dealing with proposals to change the behavior.

PR #541, opened 2020-02-03
Fix stability issues for max(), min_list/2 and max_list/2

borisvassilev:


JanWielemaker:
This doesn't fix the preferred relation `min(1,1.0) == min(1.0,1)`.  If you fix that, all the rest can remain the same.

borisvassilev:
To clarify my original intent:

This has nothing to do with the types of the two arguments to min() and max(), this is only about their position. This way, a "minimum" of a two "equal" (with `=:=`) numbers is the one that comes first, and the maximum is the one that comes second. Similarly, the min of a list is the first equivalent element and the max is the last.

This is really not so useful, I know.

borisvassilev:
PS: such considerations would only be useful if sorting on a key that is a "number". In this case, it seems to me that 1 and 1.0 shouldn't be swapped, for example. Similarly, 3/4 and 0.75 shouldn't swap places either, no matter in what order they appeared.

JanWielemaker:
My intend is to define min/max to do numerical comparison.  If this is equal, do standard order comparison.  If that is also equal, take the first/last (not sure it makes a difference and whether or not min should e.g. take the left and max the right).   The latter situation is only relevant for two floats that have not the same bits, but are numerically equivalent.  I don't recall if/how standard order deals with this ...

I think that fixes stability between standard order (sort) and min/max, which is what we were after, right?

borisvassilev:
:-) yes, making standard order consistent with min() and max() is a worthy goal. I will re-think the pull request, do not close it yet.

borisvassilev:
So, from what I understand, IEEE 754 does define a total ordering. I will have to look at the code and experiment a bit to see if the current implementation follows this (why wouldn't it?). At least -0.0 and 0.0 do what you'd expect at the moment:

```
?- -0.0 =:= 0.0.
true.

?- sort([-0.0, 0.0], Sorted).
Sorted = [-0.0, 0.0].

?- sort([0.0, -0.0], Sorted).
Sorted = [-0.0, 0.0].
```

I will try to find time for this but don't hold your breath.

borisvassilev:
PS: ... and at the same time, both min() and max() are broken:

```
?- X is min(-0.0, 0.0).
X = -0.0.

?- X is min(0.0, -0.0).
X = 0.0.

?- X is max(0.0, -0.0).
X = 0.0.

?- X is max(-0.0, 0.0).
X = -0.0.
```

No one has ever complained so this cannot be high priority on anyone's list :-) but at least I have a reason to try and fix it.

PR #542, opened 2020-02-11
clearHTable() does nothing if free_symbol is not set. Most importantl…

thetrime:
…y, it doesnt remove any items from the table itself

### DIFF
--- a/src/os/pl-table.c
+++ b/src/os/pl-table.c
@@ -610,10 +610,6 @@ clearHTable(Table ht)
   DEBUG(MSG_HASH_TABLE_API,
         Sdprintf("ClearHTable(). ht: %p, kvs: %p\n", ht, kvs));
 
-  if ( !ht->free_symbol )
-  { return;
-  }
-
   while ( idx < kvs->len )
   {
     n = htable_name(kvs, idx);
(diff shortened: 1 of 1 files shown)


thetrime:
In general, we would want clearHTable to have free_symbol set otherwise memory will be leaking. But if you put term references in the htable then there's nothing to free - but you'd surely still expect the entries to be removed from the table

JanWielemaker:
I wonder what Keri was thinking (I assume it is his code).  Alteratively we could use an assert() in `pl-table.c` and set a dummy function on the export table.   That seems safer, no?

thetrime:
Indeed. it looks like before the table stuff was made lockless, it did essentially what my patch proposes - everything that normally happens, but only calling free_symbol if it's non-null. Keri seemed surprised that clearHTable was called from outside of deleteHTable. I might wait for him to comment directly here.

keriharris:
This definitely looks like a bug; clearHTable() shouldn't be bypassing removing items if there is no free_symbol function set. Since HTables are created by default without a free_symbol function (see newHTable()) and there is no API call for setting said function, it makes sense to handle a NULL free_symbol function.

JanWielemaker:
I understand we al agree to apply this patch as is?

keriharris:
Yes, I'm happy for it to be applied as it.

PR #544, opened 2020-02-14
Support rational exponents in ar_pow

ridgeworks:
1. Added `if` block to `ar_pow` for rational exponents.
2. Expanded rational unit tests to cover various cases (different numeric base types, signs, etc.)

Built and tested except that the test cases `test_rational.pl` have only been run in isolation, not in a pl_unit test environment.

### DIFF
--- a/src/Tests/core/test_rational.pl
+++ b/src/Tests/core/test_rational.pl
@@ -51,11 +51,17 @@ test_rational :-
 set_prefer_rationals(Old, New) :-
     current_prolog_flag(prefer_rationals, Old),
     set_prolog_flag(prefer_rationals, New).
+    
+set_float_undefined(Old, New) :-
+    current_prolog_flag(float_undefined, Old),
+    set_prolog_flag(float_undefined, New).
+
+
 
 :- begin_tests(rational,
 	       [ condition(current_prolog_flag(bounded, false)),
-                 setup(set_prefer_rationals(Old, true)),
-                 cleanup(set_prefer_rationals(_, Old))
+                 setup((set_prefer_rationals(Old, true), set_float_undefined(OldU, nan))),
+                 cleanup((set_prefer_rationals(_, Old), set_float_undefined(_, OldU)))
 	       ]).
 :- set_prolog_flag(rational_syntax, compatibility).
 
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks!   This looks more complicated than I feared :)   I've synced the layout, removed some (IMO) too trivial comments and added a few more.  Also squashed the commits.   That is why Github reports this as a conflict, but it is merged.

P.s. if you have no remaining local patches, the simplest way to get in sync again is `git fetch` and then `git reset --hard origin/master`.

ridgeworks:
Yes, it does get a bit complicated - a lot of differents cases with signs and odd/even denominators, but it condenses down into a single "if" block so I am happy that it's pretty self-contained. 

I tend to insert comments as I go as placeholders and reminders, and usually leave them there as crude documentation, but it's just clutter if they're not helpful. I'll leave the "de-cluttering" to you. I can send you "draft" copies ahead of the pull request if that helps the process in any way.

I'm starting the IEEE unit tests. A lot of them are currently failing because the C math libraries are returning negative NaN's. According to the available quotes of the standard, the sign is not to be treated as significant. That suggests that everybody's life would be simpler if the only NaN produced by Prolog arithmetic is the standard one as defined by the static double const_nan in pl_arith. So something like:

	if (isnan(r->value.f)) r->value.f = const_nan; 

should go after the calls to the C library functions where NaN's can be produced. I think the unit tests should identify all the possibilities. (There's a whole murky underworld in NaN values, but I don't think it should impact Prolog arithmetic.)

I'm probably going to try out some fixes on my clone, assuming that there's not a lot of activity involving pl_arith other than what I'm doing. I'm not sure I trust my git command line skills enough to do things the "simplest" way, but we'll see.

Rick

JanWielemaker:
As is, the process is satisfactory to me.   I don't mind doing some final edits as I need to read it carefully anyway.  Some people do not like to see there PR being edited.

You're switching topic.   Please start a new issue/PR for the float stuff.  Yip, most likely a lot of the arithmetic goes wrong.   A hard nut is what to do with NaN.  There is not a single NaN in float world and I think you are suggesting different float operations lead to different NaNs?   Seems you are suggesting to normalize all NaNs, which implies we can at least use = and == on them?  I have little clue what the implication of that might be.   Another or additional solution could be to add isnan/1 and/or float_class/2 based on C99 fpclassify(), producing `nan`, `infinite`, `zero`, `subnormal` or `normal`.

Don't worry about conflicts in pl-arith.c.   They are not that likely and I can deal with that.   I close this PR.

PR #546, opened 2020-02-17
Fix duplicate resource error under MSVC

pd-robgee:
#### Do not include manifest when building with MSVC, since it causes a duplicate resource error.

This fixes the "some Windows resource handling issue" referenced in https://swi-prolog.discourse.group/t/build-swi-prolog-using-microsoft-visual-studio/1086.
MSVC Generates this manifest on its own, so including it (with an id of 1) causes a duplicate resource error.

I did not completely remove the manifest, since I don't know that it is not required when cross-compiling under Linux.
If MinGW under Linux can also generate the manifest I would suggest completely removing it from here (especially since it's quite unclear that this binary blob is actually an XML).


JanWielemaker:
Hmm.  Some things may be wrong here.  The idea is that Windows code depends on `__WINDOWS__`, MinGW code on `__MINGW32__` and MSVC code on `_MSC_VER`.   Once upon a time that was correct ...

I don't know the MinGW status wrt adding a manifest itself.

pd-robgee:
I suppose this is a discrepancy between the C and the resource compilers. I'll check if `__MINGW32__` is also defined by the MinGW resource compiler. That should match the  intent more clearly.

pd-robgee:
I did some further research, there is a package (for MinGW-w64 at least) "mingw-w64-windows-default-manifest". However it is not installed by default (at least on my system) so I thought it was better to keep the manual manifest when building under MinGW.

JanWielemaker:
> I thought it was better to keep the manual manifest when building under MinGW.

Agree.  And this looks like a step in the right direction.  On other stuff now, but I'll check this and merge unless weird things happen.   Thanks!

pd-robgee:
Thanks and no problem!

BTW, I verified on:
* Debian 9 with x86_64-w64-mingw32-gcc 6.3.0, using a native build for the Prolog build steps
* Visual Studio Professional 2019 16.4.5

JanWielemaker:
Thanks,  Don't worry about the conflict.   Squashed and changed commit message.  Will look at the others tomorrow.

Issue #545, opened 2020-02-16
GMP rounding mode for mpX_get_d() functions


ridgeworks:
In informal testing of  the rational tripwire in 8.1.22 and  I discovered that the GMP `mpX_get_d()` functions do not heed the hardware rounding mode, instead they always truncate (round to zero). This is a bug IMO.

SWIP can workaround the bug by wrapping calls to these functions in a call to something like:

```
double mpX_round(double d) {
  switch (fegetround()) {
    case FE_UPWARD  : return (d>=0) ? nexttoward(d, INFINITY) : d;
    case FE_DOWNWARD: return (d<=0) ? nexttoward(d,-INFINITY) : d;
    default: return d;
  }
}
```

For example, replace `mpz_get_d(r->value.mpz)` with `mpX_round(mpz_get_d(r->value.mpz))` .

There are 4 such calls in pl_arith.c and probably a few more elsewhere (promoteToFloatNumber() ?).

jburse:
I dont know whether nexttoward solves the problem. I am expecting 
that promotion uses "nearest". So that in the end we have for example:

```
/* should be true */
?- 3.0E23 =:= 3*10^23.
false.

?- 3.0E22 =:= 3*10^22.
true.
```

ridgeworks:
I think it does solve the problem for `to_positive` and `to_negative`modes. In these cases, depending on the sign of the value, you know whether the rounding was incorrect or not. The function has no effect if the mode is `nearest` since you don't know whether the rounding was incorrect, as your example demonstrates.

JanWielemaker:
I'll happily handle pull requests :)

jburse:
Yes. But I doubt that `nexttoward` would do something useful, like
for example converting "towards_zero" into "away_zero". Try:

    ?- 123 =:= 123.0
    true

As you observed yourself, you would need to know whether 
rounding was exact, and only act up, when it wasn't exact.
What you could do for "nearest", add a float ULP/2 to the 

number before doing "toward_zero". You would possibly get
"half_down". For "half_even" you could adopt this code,
which inspects the limbs of the number and does not

really perform the float ULP/2 addition:

**OpenJDK - BigInteger**
[https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/math/BigInteger.java#l3929](https://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/math/BigInteger.java#l3929)

[...]

PR #549, opened 2020-02-17
Gmp rounding (#545)

ridgeworks:


JanWielemaker:
Ok.  Was integrating, but I think I should wait until you are done ...  Please send a message when you think you are done.    Looks great!

A few small things:  please indent clause bodies with 4 spaces for consistency (thanks for trying to align operators and terms).   Please define local C functions as `static` to avoid polluting the name space and allow the compiler to perform local optimizations (static functions used only once are typically inlined and otherwise the compiler can choose the best calling conventions rather than the compatible one).

ridgeworks:
Sorry for any confusion.

From my end, I submitted a pull request for the test_IEEE754.pl unit test case file. That should be ready for merging.

I'm also working on a fix for the GMP rounding issue. That's almost ready to go, but I've only just done what I thought was a local merge. Are you seeing that? I assumed you wouldn't until I did a pull request back to your main repository.

I have just added a couple of functions mpX_round() which is exposed through the pl_gmp.h which I assume will not be static. And a local support function (used once) for used in ap_pow which I will make static before issuing the pull request.

I've extended the rounding mode test cases, so I'll try to fix the clause body indent at the same time. (I use tab (=4 spaces) for my indents, so that habit will be a bit hard to break.)

Rick

JanWielemaker:
> Are you seeing that?

Yes.  Once you create a pull request, anything else that you push to that branch gets added to the PR.   Might be wise to merge the whole thing in one go?  

Note that to run the test, just do `swipl test_XXX.pl` an then run `?- test_XXX.`  Note that the file base name (`test_XXX`), the module name and the toplevel predicate should all have the same name.    I started turning everything into `ieee` (lower case) as upper case only gives problems on case insensitive file systems.

ridgeworks:
OK. I'll try and cleanup the test file and generate another pull request. Will let you know.

ridgeworks:
I think I cleaned everything up except for test file name. I'll keep my fingers off it until you merge. It says it's mergeable but let me know if I need to fix anything.

JanWielemaker:
Thanks a lot!

Ok.  All done.  I added `blocked(fail)` to tests that failed and `blocked(crash)` to one that actually crashes Prolog (FPE).   I assume you will continue fixing this issues and enable the failing tests?

Please discard this branch and create a new PR from `master` (I think you already got that,
but it happens often that people continue on modified merged branches which doesn't make the processing easier).

We should sync `nearest` `to_nearest` with ECLiPSe.   I'm in favour of `to_nearest`, but recall reading `nearest` in the proposal.   Did I miss something?

ridgeworks:
We should go with to_nearest for compatibility. I'm not sure where nearest came from, hope it wasn't me.

I will undertake to keep submitting fixes to clear the test case issues, but I may need some guidance from time to time where the scope of fixes is larger than just pl_arith.c  or I just don't understand the existing code well enough to make an informed decision.

I agree it's good practice to regularly discard modified branches that have been merged.


A digression:

While I agree that there are valid reasons for Eclipse compatibility, I don't think it's the best model for Prolog arithmetic. It isn't as big a deal for Eclipse users because breals (another name for intervals) are a first class numeric type. Support for IEEE continuation values through flags is fine, and actually entails the most work (as demonstrated by the test case results). But I would like to add better alternatives for rounding control and next value (nexttoward). 

Rounding is an admission that something is not quite right with the whole floating point concept - no mathematician has rounding in his vocabulary. To the standards body's credit rounding has been part of the standard from the beginning, so they recognized the problem. I believe they had two choices for rounding, either a global mode or a per operation setting (as in a couple of bits in the relevant machine fp instructions). I think they went with the global mode because it was easier to implement and didn't require compiler support, but it's not clear to me it was the best choice. And it also has a performance impact because the fp pipeline has to be flushed every time the mode changes, but there's not much we can do about that.

Programming language designers have historically not known what to do about rounding control. For years they ignored it. I remember when in-line assembler was required; error prone and very un-portable. Eventually they added a thin veneer over the hardware global mode in the form of a couple library functions; easy enough, and hasn't changed since. Other language designers just followed suit. But a global flag which changes the result of an evaluation without any visible change in the expression is borderline unsafe IMO. It's analogous to a compiler optimizing flag that's broken, i.e., the same program produces different results depending on the flag setting.

A better model IMO is the instruction level control, but applied in Prolog at the subexpression level. This would entail adding arithmetic functions which scope rounding to the subexpression, so to_positive(Exp), to_negative(Exp), and to_zero(Exp). The default is to_nearest so no need have an explicit function for that. Now the rounding control is explicit in the expression and not implicit in some global state. Given this programmers should be discouraged from using the global flag - it's purely there for Eclipse compatibility reasons. And the bonus is that it's considerable more efficient than using the flags, as discussed before.

The other thing I would like to add is a bounded_number predicate, so bounded_number(Num,Lo,Hi) or  bounded_number(Num,[Lo,Hi])as a better solution to the problem addressed by the nexttoward arithmetic function. It's a full blown relation so it can be used in various ways and it can provide both upper and lower bounds in a single primitive call, so more efficient (for some purposes). This feels much more Prolog-like (logical) than a function, which isn't fully compatible with Eclipse in any case. I'm a little less opinionated about this one, because it's easy to roll your own in Prolog (and I will if necessary), but it feels like the right thing to do and users should be encouraged to think this way.


These additions look to me like a trivial amount of work and fully backwards compatible. I certainly don't expect anyone else to do it, including any documentation if you point me at it. But this is the right time, rather than waiting for alternative practices to become entrenched. There doesn't seem to be a lot of disucssion on discourse so maybe nodbody else cares (not syntax, I guess), but I did want to fly these specific proposals  past you in case you have any philosophical objections.


Rick

JanWielemaker:
> to_nearest

Yes.  No clue from where I got `nearest` :(   Please change.

> A digression

I think I agree with all you say.   I would like to take it a step at a time though.   A to_X(Expr) function looks attractive, but there is a problem: expressions are evaluated bottom-up.   This is probably all doable, but we have to take care of two cases: interpreted is/2 and compiled is/2.   Both come with their problems.  The interpreted version isn't a simple recursive function as this tends to exhaust the C stack, causing crashes, so it is a loop managing its own stack.

> documentation

Docs are in the `man` subdir, for arithmetic mostly in `builtin.doc`.  These .doc files are in fact LaTeX files, but they are pre processed a little.   If you are used to LaTeX it should all be pretty simple.   If you (by now) have build the entire system you can include documentation into the build and check what you have done using `?- help(mypred)`.

P.s.   Would the Slack clone Mattermost be more suitable for these things?  I have an instance running for SWI-Prolog and can invite you.   I'm not too happy with long discussions in Github as they tend to get spread over issues, PRs, etc.

ridgeworks:
See below:

> On Feb 18, 2020, at 10:43 AM, Jan Wielemaker <notifications@github.com> wrote:
> 
> to_nearest
> 
> Yes. No clue from where I got nearest :( Please change.
> 
> 
Will do.
> A digression
> 
> I think I agree with all you say. I would like to take it a step at a time though. A to_X(Expr) function looks attractive, but there is a problem: expressions are evaluated bottom-up. This is probably all doable, but we have to take care of two cases: interpreted is/2 and compiled is/2. Both come with their problems. The interpreted version isn't a simple recursive function as this tends to exhaust the C stack, causing crashes, so it is a loop managing its own stack.
> 
> 
I was using eval as my model. So I'd basically wrap a call to it (or cpNumber) with some calls to fesetround. Is there a reason that won't work?

I'm pretty clueless about compiled arithmetic. Can you give me a short description of what's going on there?


> documentation
> 
> Docs are in the man subdir, for arithmetic mostly in builtin.doc. These .doc files are in fact LaTeX files, but they are pre processed a little. If you are used to LaTeX it should all be pretty simple. If you (by now) have build the entire system you can include documentation into the build and check what you have done using ?- help(mypred).
> 
> 
I have been avoiding building the documentation to this point (just a core build for testing purposes), so I'd have to take this next step.
> P.s. Would the Slack clone Mattermost be more suitable for these things? I have an instance running for SWI-Prolog and can invite you. I'm not too happy with long discussions in Github as they tend to get spread over issues, PRs, etc.
> 
> 
I'll try to do whatever works best for you; just tell me what that is and what I'd have to do differently.

JanWielemaker:
> Is there a reason that won't work?

If you look at valueExpression(), you'll see it is not a simple recursive function, so you have to push/pop the required rounding reset operations onto the stack that is being maintained there.  Well, I can help.

> compiled arithmetic

If -O is used, arithmetic expressions create a stack engine in the VM.  That avoids creating terms and subsequently analyse these.   So, `A is B+C` is translated into something like push(B), push(C), func(+), is(A).  The stack consists of the number structures you now know about.  You can see what happens by writing a small predicate and run `?- vm_list(mypred).`  So, this needs to include rounding instructions.  Not a very big deal, but some work.

I'll invite you on Mattermost.  We'll see whether we like it.  May be tomorrow, I've got to run ...

ridgeworks:
I've looked at valueExpression and think I understand what you're saying. Not quite sure how to fix it so I'll await your input.

JanWielemaker:
I've pushed a branch `fround` for inspection.  This allows for

```
101 ?- A is roundtoward(sqrt(2), to_negative).
A = 1.414213562373095.

102 ?- A is roundtoward(sqrt(2), to_positive).
A = 1.4142135623730951.
```
It should allow for nesting and ensure restoring the mode in case of an exception.

This branch doesn't yet implement this function for optimized compilation.   Did you get a mattermost invitation?

ridgeworks:
Just to make sure I understand the code, the steps are:

1. In valueExpression(), if the functor is roundtoward, set the rounding mode according to the second argument. set_roundtoward() updates it's second argument to the current rounding mode (n2).

2. Evaluate the first argument (n1).

3. Then call ar_rountoward(n1,n2,R) which sets r to n1  and restores the previous rounding mode from n2. Done.

4. In case of an error, valueExpression saves the current rounding mode at the beginning and restores it if the error exit is taken.

That looks good to me. It's a bit unfortunate that there's a special case in valueExpression just for this purpose, but I guess this is a special case.


I've not received a mattermost invitation; I assume that will come by email.


I've implemented a number of fixes affecting files ATOMS, pl_arith.c and pl_gmp.c and the IEEE tests. I'm going away for a week starting tomorrow (Thursday), so won't get back to fixing anything before I return, but I'll push the changes before I leave so things don't get too out of sync. I should be in email contact if that is helpful.


Rick

JanWielemaker:
Yes.  That is it.  I kind of copied the conventions from nexttoward/2 which I consider related.    I'll pick up your PRs.  All this is not on any critical path as long as we do not break anything existing AFAIK, so don't worry.

Could the invitation  be in your spam box?

ridgeworks:
Just generated a pull request for my latest updates; good progress but a few more to go.

Checked the spam box; no invite. Will keep looking.

PR #550, opened 2020-02-18
Bugfix/static library

pd-robgee:
This PR fixes building libswipl statically under MSVC: MSVC does not automatically prepend 'lib' to the library name, furthermore it would generate swipl.lib in the same directory as swipl.exe, causing a linking failure: "LNK1114 cannot overwrite the original file 'build/src/Debug/swipl.lib'; error code 5"
Reference: https://github.com/MicrosoftDocs/cpp-docs/issues/741

Furthermore building libswipl statically without multithreading caused the CMake variable `C_CFLAGS` to be empty, and therefore `#undef`ed in `parms.h.cmake` on all platforms. Since it is unconditionally used in code, this caused a compilation error.


JanWielemaker:
Thanks.  Somehow Github doesn't pick up the offline merge, but it is merged ...

PR #553, opened 2020-02-19
FIXED: Debug statement in pl-tabling.c which depended on multi-thread…

pd-robgee:
…ing being enabled without checking if multi-threading is enabled.

This makes the build with `-DCMAKE_BUILD_TYPE=Debug -DMULTI_THREADED=OFF` pass again and I presume this logic is solid, but please check if the added debug statement makes sense for single-threaded builds.


JanWielemaker:
Thanks.  Merged after removing the assertion from the non-mt branch.   The whole distinction doesn't exist, so it can't be wrong either.

PR #554, opened 2020-02-19
IEEE fixes - batch 1

ridgeworks:
1. `float_rounding` flag value, change `nearest` to `to_nearest`
2. Prolog arithmetic now produces a single signed NaN (same value as `nan`).
3. Unit test cases `ieee_cmp`,`ieee_sign`,`ieee_log`,`ieee_log10`,`ieee_lgamma`,`ieee_max` and `ieee_min` now pass. (Still `ieee_div`, `ieee_pow`, `ieee_ceil`, `ieee_floor`, `ieee_round` and `ieee_trunc` to fix.)
4. Arithmetic functions `pi` and `e` now produce correct value under `to_positive` rounding mode.


JanWielemaker:
Thanks for all that!   On `master` you find everything merged as well as the new `roundtoward/2` function.  I made some changes, notably:

  - Moved the single nan into check_float() and as a result change the argument to be a number
    rather than a float.
  - Disabled the `AR_UNDEFINED_IF` and `AR_DIV_ZERO_IF` macros.  They are from the old days
    where most of these functions raised SIGFPE.
  - Changed the lgamma/1 test for lgamma(-0.0).  On my machine that produces `Inf`, which I think
    is correct as it is "the absolute value of ...".   If they fails on the Mac we might have to hack around
    that.
  - Changed the rounding tests near the end to be all =< or >=.  On my machine the cos(-pi) test
    produces equal values for all roundings, which I think is fine.

Please check whether you agree and if not, start a new PR from the current master.   You probably also want to use the new roundtoward/2 function :)

ridgeworks:
See below,

Rick

> On Feb 20, 2020, at 6:10 AM, Jan Wielemaker <notifications@github.com> wrote:
> 
> Thanks for all that! On master you find everything merged as well as the new roundtoward/2 function. I made some changes, notably:
> 
> Moved the single nan into check_float() and as a result change the argument to be a number
> rather than a float.

I wondered about the best place to put this check. Finally decided on the success exit from valueExpression() since that was sure to catch everything, and minimized the number of checks during evaluation so maybe (very) slightly more efficient. But I have no problem you moving it to where you see fit. If some function fails to call checkFloat when it should, the unit tests should catch it.
> Disabled the AR_UNDEFINED_IF and AR_DIV_ZERO_IF macros. They are from the old days
> where most of these functions raised SIGFPE.
OK. I didn't really know what these did, so I wasn't about to change them.
 
> Changed the lgamma/1 test for lgamma(-0.0). On my machine that produces Inf, which I think
> is correct as it is "the absolute value of ...". If they fails on the Mac we might have to hack around
> that.
I also think +Inf is the correct value but the C11 spec is a bit vague. The Mac seems to produce -Inf but I tested on a code snippet on an online C programming website and it produced +Inf so I wasn't sure what was going on. 

We should probably change to a custom ar_lgamma that changes everything to +Inf so this discrepancy doesn't escape to the Prolog level. 
> Changed the rounding tests near the end to be all =< or >=. On my machine the cos(-pi) test
> produces equal values for all roundings, which I think is fine.
I think this is OK with possibly one exception, round to_negative should always be less than round to_positive. But let me think about this a bit (and do some more testing). 
> Please check whether you agree and if not, start a new PR from the current master. You probably also want to use the new roundtoward/2 function :)
> 
> 
When I get back to this, I'll start with a fresh copy of the master. I'll add a version of the rounding tests that use your new roundtoward function in the next batch of fixes. Thanks for doing that, you probably saved both of us a lot of time.

JanWielemaker:
>  We should probably change to a custom ar_lgamma that changes everything to +Inf so this discrepancy doesn't escape to the Prolog level.

The tests pass on my Mac.  I agree that if there are portability issues we can simply roll a more dedicated wrapper.

> I think this is OK with possibly one exception, round to_negative should always be less than round to_positive. But let me think about this a bit (and do some more testing).

This is a theoretical exception though.  cos(-pi) is `-1` and therefore no rounding is needed.   I doubt the libraries recognise this as `pi` is of course also not the real thing.   What happens though if the internal 80 bits (I think) representation is exactly the same as a representable double?

ridgeworks:
> On Feb 21, 2020, at 4:09 AM, Jan Wielemaker <notifications@github.com> wrote:
> 
> I think this is OK with possibly one exception, round to_negative should always be less than round to_positive. But let me think about this a bit (and do some more testing).
> 
> This is a theoretical exception though. cos(-pi) is -1 and therefore no rounding is needed. I doubt the libraries recognise this as pi is of course also not the real thing. What happens though if the internal 80 bits (I think) representation is exactly the same as a representable double?
> 
> 
You're right and this was a poor choice on my part. This is supposed to be a test of the rounding modes, not the underlying math library. Will be addressed in next batch of fixes.

JanWielemaker:
It does hint at a possibly more fundamental issue for you though: what does float rounding actually mean?  Is the `to_negative` rounded function result always strictly =< the true (mathematical real) value?   I doubt this is the case (as few things one would think must be true about floats are actually always true).

ridgeworks:
Yes, I need to more investigation. Currently I'm force rounding all calculations using epsilon, which is conservative.

My take is that whenever an IEEE rounding operation occurs, e.g., to convert values internal to the hardware (with additional guard bits), the rounding will occur in the designated direction. To me that means that a to_negative value is always =< the to_postive value.  And they'll only be the same when the float value is precise. Similarly for the to_zero, depending on the sign of the value. The only unpredictable one is to_nearest, but again it should be >= to_negative and =< to_positive. I'll pay more attention to this when I redo the unit test, but I'm prepared to be surprised.

Mathematical correctness is a different issue. I think for standards compliance, the result has to be within 1 ULP (unit of least precision) for most operations, which should be fine for my purposes.. But transcendentals may be an issue. I know of interval arithmetic implementations that have replicated hardware functionality (e.g., log, exp, transcendentals) in software (Taylor series) because they didn't trust the results. (I'm *not* about to do that.)

At some point I'm going to run some identity functions (Interval =:= exp(log(Interval)) , Interval =:= sin(asin(Interval), etc.) on a bunch of random values to satisfy myself that there's nothing obviously amiss.

PR #556, opened 2020-02-20
FIXED: MemoryBarrier under MSVC for 32-bit targets is an inline funct…

pd-robgee:
…ion instead of a #define

### DIFF
--- a/src/pl-atom.c
+++ b/src/pl-atom.c
@@ -617,7 +617,7 @@ redo:
         PL_free(a->name);
       a->type = ATOM_TYPE_INVALID;
       a->name = "<race>";
-      MemoryBarrier();
+      _MemoryBarrier();
       a->references = 0;
       goto redo;
     }
@@ -921,7 +921,7 @@ destroyAtom(Atom a, Atom **buckets)
 
   a->name = "<reclaimed>";
   a->type = ATOM_TYPE_INVALID;
-  MemoryBarrier();
+  _MemoryBarrier();
   a->references = 0;
 
   index = indexAtom(a->atom);
--- a/src/pl-atom.c
+++ b/src/pl-atom.c
@@ -617,7 +617,7 @@ redo:
         PL_free(a->name);
       a->type = ATOM_TYPE_INVALID;
       a->name = "<race>";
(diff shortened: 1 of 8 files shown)


JanWielemaker:
Can't we keep this limited to pl-inline.h using a #define MemoryBarrier() _MemoryBarrier()?

pd-robgee:
I changed the name before including `windows.h`, it was necessary then because otherwise the macro would be applied in `winnt.h` for compilation units that include `windows.h` later than `pl-inline.h`. This would cause syntax errors in `winnt.h` because the name of the function definition would be replaced by `(void)0`.

However now that `windows.h` is included in `pl-inline.h`, this should not be necessary anymore, let me check if my logic is correct there.

JanWielemaker:
I see.   What about changing throughout the code all calls to `MEMORY_BARRIER()` and map that appropriately dependent on the OS/C compiler in pl-inline.h?   My experience with #defines and #undef around system include files to fix conflicts is not very positive.

pd-robgee:
As I see it, not including `windows.h` in `pl-inline.h` was an error in the first place, since we depend on that file to provide `MemoryBarrier` (either as a function or as a define), and then define it ourselves when `windows.h` does not. The whole `_MemoryBarrier` was before I realized that we did not yet include `windows.h`.

Also I see that my comment also got mangled by my renames, which might have caused some confusion.

I have pushed a new commit, which tidies it all up. However we might still consider changing the calls to `MEMORY_BARRIER()` to indicate that we might override the bare function/macro defined by the compiler. However I'll leave that decision to you.

JanWielemaker:
Ok.  Merged the two commits and then added one renaming to MEMORY_BARIER, which is defined to MemoryBarrier for MSVC, and to __sync_synchronize() if CMake detects this built-in (GCC and CLang) and otherwise `(void)0`.   Tested the build on Linux/gcc and Linux/MinGW for Windows.

I trust that if you see I did anything wrong, I'll hear it :)   Thanks!

pd-robgee:
Thanks, looks good! I just verified, building under MSVC also works.

PR #557, opened 2020-02-21
Create test_utf8.pl

EricGT:
Initial test cases for utf8_codes//1

### DIFF
--- a/src/Tests/library/test_utf8.pl
+++ b/src/Tests/library/test_utf8.pl
@@ -1,0 +1,118 @@
+/*  Part of SWI-Prolog
+    Author:        Jan Wielemaker, Eric G. Taucher
+    E-mail:        J.Wielemaker@vu.nl
+    WWW:           www.swi-prolog.org
+    Copyright (c)  2006-2020, University of Amsterdam
+                              VU University Amsterdam
+    All rights reserved.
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Unfortunately lines 100 and 101 contain non-ascii data that is garbled in the process somehow.  Files should be strict US-ASCII except for comments to avoid misinterpretation.

There is also no point in checking both `"\x61"` and `"\u0061"`.    The first is an old Prolog mistake that means nothing because it is undefined how the number is to be interpreted.

If the tests are ok I can reconstruct the input :)

JanWielemaker:
Ok.  Fixed the input.   I've left in all the different Prolog representations for strings and lists, although 
they do not test the UTF-8 library, but the Prolog parser.   That is useful too :)   Merged.

EricGT:
> There is also no point in checking both "\x61" and "\u0061". The first is an old Prolog mistake that means nothing because it is undefined how the number is to be interpreted.

I did those extra variations because when I need to understand how code works I often look at test cases. Sometimes seeing the extra variations in the examples helps me through a problem. So I just put them there in case someone else might need to see that those variations also work. There were many that didn't but I didn't include those. :)

PR #560, opened 2020-02-28
ieee fix 2 

ridgeworks:
1. `pl-arith.c` - added custom lgamma for consistent answers across platforms when arg <=0.

2. `test_ieee754.pl` Fixes and expanded tests:
  - rounding tests use imprecise expressions to test effect of mode
  - `roundtoward` tests added
  - exception tests added
  - power tests fixed. Require parentheses for negative IEEE special values due to operators `-` and `**` precedence.


JanWielemaker:
Merged after layout and combining some commits.

PR #561, opened 2020-03-02
New bounded_number predicate

ridgeworks:
1. pl-arith.c - implementation - simplicity/efficiency tradeoff biased towards simplicity.
2. test_ieee754.pl - test group added.
3. builtins.doc - added doc after float_class/2 in Arithmetic, but should be easy to move. Untested doc since I don't have a full install environment.

### DIFF
--- a/README.mingw
+++ b/README.mingw
@@ -9,20 +9,67 @@
 # Bagnara (initial port of the Prolog   kernel), Abramo Bagnara and Keri
 # Harris.
 #
+# The current releases are  configured  and   build  using  CMake.  As a
+# result, this README only applies for   building the PREREQUISITES. See
+# CMAKE.md and/or scripts/make-distribution for configuring and building
+# SWI-Prolog.
+#
 # USAGE
 #
 # The remainder of the  file  is   an  executable  specification  of the
-# installation process. To use it, copy this   file to a directory above
-# the location where you want to   build all components. I.e., typically
-# the directory above this one and source it into the bash shell:
+# installation process. The setup is designed to create ~/mingw32 and/or
+# ~/mingw64 for building and installing   the prerequisites. We describe
+# the setup for Win64. Simply replace  64   with  32 to build for 32-bit
+# Windows. First, install the requires tools from the Linux packages:
 #
-#   % . README.mingw win32|win64
+#    % sudo apt-get install wine wine-binfmt mingw-w64
+#    % sudo apt-get install cmake ninja-build nsis
 #
-# Next, execute the functions for the parts of the process you want. The
-# function build_all does everything.
+# Next, create and prepare the prerequisite location (the script changes
+# your prompt to `W64`
 #
-# This script has been tested for   cross-compilation  on a Linux system
-# with the mingw cross-compiler and wine installed.
+#    % cd ~
+#    % mkdir -p mingw64/src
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks and sorry to have missed this.  Merged after some edits:

  - bounded_number/3 was wrong on int64_t overflow/underflow.   Now uses ar_add_ui()
  - Got rid of the special_ieee flag.  Use isfinite() to check for both inf and nan.
  - Use promoteToFloatNumber() rather then dedicated call.  This uses the mpq_to_double(),
    which I think is better, but I'm not sure.   Possibly needs to set the rounding mode to
    _nearest_ for the promotion?
  - Changed round tests for -1/3 to use -1.0/3 as it otherwise fails of `prefer_rationals` is set
    and we want float arithmetic here.
  - Re-formulated the rounding tests a little to make it easier to see which numbers are produced.

Except for the rounding stuff (notably the mpq_to_double() to be resolved), I have one open issue.  We now have between/3 and bounded_number/3 that seem too close IMO to justify different argument order.   If you agree, please change the argument order of bounded_number/3.

ridgeworks:
Only issue with the changes is the rounding behaviour of `mpq_to_double()` under various modes. It appears that the only GMP function `mpz_fdiv()` uses is `mpz_get_d_2exp()` which rounds to zero (truncates):
```
    l = mpz_get_d_2exp(&le, li);
    s = mpz_get_d_2exp(&se, si);
    if ( swapped )
      res = (l/s) * pow(2.0, le-se);
    else
      res = (s/l) * pow(2.0, se-le);
```
So sometimes (not always) the numerator will be truncated, and the same for the denominator. And since GMP seems to be oblivious to the actual current rounding mode, the mode shouldn't affect the `mpz_get_d_2exp()` result but can affect the subsequent divide/multiply/power operations. At this point I have no clue as to what the range of outputs could be. If it's within 1ULP all is well.

On balance, I think I prefer the dedicated call to `mpq_get_d()` because it is predictable, more efficient (although that's not a primary consideration),  and known to be within 1 ULP. (My original code also used `mpX_round()` which probably isn't necessary for this purpose.) 

Also note that if programmers are using `roundtoward()` (as they should) rather than the global flag, the rounding mode when this predicate is called is always `to_nearest` which does reduce the uncertainty somewhat.

Regarding the argument order, it does make sense to use the same ordering as `between/3`. If we do that, perhaps we should consider changing the predicate name to just `bounded`. `bounded_number` only makes more sense (to me) when the first argument is the number.

JanWielemaker:
Hard to say.  I guess the problem only starts to play when the numerator and/or denominator are greater than the the max representable precise integer for a double.   I think the claim behind mpq_to_double() is to be more efficient, but that might apply to older versions of GMP only.    I have some preference to have only a single place to do MPQ -> double promotion to keep things tractable.  Can we postpone this one for now?

I am in favor of flipping the arguments, but certainly not for a `bounded/3`.   That, to me, is a far too general name for a bit odd built-in.   `bounded_number` is fine as well: the target (number) is a the end and the _bounds_ are the first arguments.   You could go for `bounds_number`?

ridgeworks:
Ok. Will leave implementation and name as is. But will re-order the arguments.

I like `bounded` as a name (analogous to `between`), but understand the issue.

PR #563, opened 2020-03-06
bounded_number/3 arguments re-ordered

ridgeworks:
Argument ordering now matches between/3.
Affects implementation, unit tests, and doc.


JanWielemaker:
Thanks!  Merged after squash.

PR #564, opened 2020-03-09
ar_pow fixes

ridgeworks:
1. Factored out special case code for `n1` (-1,0,1), which apply to any type of `n2`, and placed it first.
2. Fixed rational exponent cases where a float results, e.g., -3**1r3. Added a few unit tests in test_rational and test_ieee754 (for rounding power expression).


JanWielemaker:
Thanks.  Merged after some layout.   Remaining issues:
  - Two failing tests:

```
ERROR: /ufs/wielemak/src/swipl-devel/src/Tests/core/test_ieee754.pl:506:
	test ieee_rmode: assertion failed
	Assertion: plunit_ieee754:test_rounding(-2.0**1/3)
A
ERROR: /ufs/wielemak/src/swipl-devel/src/Tests/core/test_ieee754.pl:515:
	test ieee_rndto: assertion failed
	Assertion: plunit_ieee754:test_roundto(-2.0**1/3)
A. done
```

  - What should happen to `0**nan`?  Now says `0`, but this seems dubious to me.

ridgeworks:
Were the failing rounding test fails on the same platforms as the "dubious rounding" fails or general across all platforms.

0**nan definitely wrong; fix required. Are there any ** unit tests where this should be added?

PR #565, opened 2020-03-11
ar_pow C11 compliance and relaxed rounding test

ridgeworks:
1. Fixes for C11 compliance on integer special cases (incl. `0**nan`).
2. IEEE test cases: relaxed rounding check for round to_zero. No longer expects equivalence with Rn/Rp.

### DIFF
--- a/library/writef.pl
+++ b/library/writef.pl
@@ -233,11 +233,11 @@ swritef(String, Format) :-
     name(Atom, String),
     '$padout'(Atom, Size, Just).
 '$padout'(Term, Size, Just) :-
-    format(string(Atom), Term, Atom),
-    atom_length(Atom, Length),
+    format(string(String), '~w', [Term]),
+    atom_length(String, Length),
     '$padout'(Just, Size, Length, Left, Right),
     tab(Left),
-    write(Atom),
+    write(String),
     tab(Right).
 
 '$string'(0) :- !, fail.


JanWielemaker:
Thanks.  Merged after some cleanup.  There was a serious mistake in your new set_rnum() which left the type to be V_MPQ pr V_MPZ if the first argument was of that type.  Luckily caused a crash in test_arith,pl.

Getting close?

ridgeworks:
Oops, my bad. 

That `set_rnum()` thing was a possibly misguided attempt to preserve an argument type in the result, but I'm not really sure that's right. The results of these special cases are precise, so there's probably no reason why they aren't just integers. 

Other than integer divide by zero discussion (TBC offline), I think the end is in sight.

JanWielemaker:
> The results of these special cases are precise, so there's probably no reason why they aren't just integers.

That is indeed a good point.  Related are issues as X*0, which can be integer 0 regardless of X (not to be confused with X*0.0; that must be a float.

ridgeworks:
Or `X/inf`? Why not 0 for any finite number X?

JanWielemaker:
I see GitHub interpreted my markdown a little different from what I expected.  Not sure where `X/Inf` refers to.   I was reffering to anything times (integer) 0.  Surely for any finite number that is precisely the integer 0.   Not sure about Inf/NaN.

P.s.  Joachim has played with the idea to allow for the rational 1/0 as a thing denoting rational infinity.  He didn't like all the consequences it has to the code.  It is something to keep in mind.

ridgeworks:
Sorry for the confusion. `X/inf` was just another example of a precise result from any `X`, like `X*0`.

Not saying it's a good idea, but how would one store the rational `1r0` internally? Wouldn't GMP get upset if it got hold of one? Or maybe that was your point.

JanWielemaker:
Ok.  `X/Inf` is only 0 if we have a reliable infinity, not merely a float overflow.  I think it is easy to make GMP represent 1/0.  The only issue is that you have to treat this special almost anywhere in the code.  That is probably also what bothered Joachim.   Alternatively we could have our own rational infinity and just print it as `1r0`.   No need to rush.

ridgeworks:
The only way of prohibiting an infinity from float overflow via the `float_overflow` flag. If that kind of infinity is unacceptable in this case, it's probably unacceptable everywhere so just set the flag to `error`.

For now, I'm happy to leave it at X/Inf --> 0.0. Seems much easier than inventing a new form of infinity which propagates through all the integer/rational arithmetic, and for the vast majority of users, this is a non-issue.

PR #567, opened 2020-03-19
Uniform divide_by_zero and cleanups

ridgeworks:
Arithmetic function updates:

1. Divide by 0 semantically equivalent to divide by 0.0. Generates an infinity or divide by zero error depending on state of flag `float_div_zero`. No divide by zero case should generate `float_overflow` error.
2. Special cases in `/` and `**` can return integers where appropriate, e.g., `X**0 --> 1, 1**X --> 1, 0**X -> 0 for X>0, -1**X -> 1 for even_integer(X), -1**X --> -1 for odd_integer(X), X/inf --> 0 for integer(X)`, etc.
3. Added unit tests to `test_ieee754.pl` and `test_rational.pl` for updated functionality. These all pass but arithmetic tests outside the these specific modules have not been executed.


JanWielemaker:
Thanks.  Merged after squash, commit message and a few comments.    Seems we are getting really close!

PR #569, opened 2020-04-06
snap: use kde-neon extension

LyzardKing:
I managed to find some time to look a the snap integration recently.
This PR adds the (not yet autoconnected) kde-neon content snap.
This is used for the qt interface to swipl.

The snap is confined, thus it will not integrate with other programs or libraries installed via other means.
It will support packages installed in swipl itself.
If this is not acceptable the other solution is to build a classic unconfined snap.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,7 +1,7 @@
 name: swi-prolog
 base: core18
 adopt-info: swi-prolog
-icon: packages/swipl-win/swipl.png
+#icon: packages/swipl-win/swipl.png
 license: 'BSD-2-Clause'
 summary: ISO/Edinburgh-style Prolog implementation
 description: |
@@ -22,25 +22,27 @@ description: |
 
 apps:
   swipl:
-    command: kf5-launch $SNAP/usr/bin/swipl
+    command: swipl
+    extensions:
+      - kde-neon
+    plugs:
+      - home
+      - network
+      - removable-media
+      - opengl
+      - audio-playback
+      - audio-record
   swipl-win:
-    command: kf5-launch $SNAP/usr/bin/swipl-win
-
-plugs:
-  home:
-  network:
-  x11:
-  desktop:
-  desktop-legacy:
-  removable-media:
-  opengl:
-  audio-playback:
-  audio-record:
-  kde-frameworks-5-plug:
-    content: kde-frameworks-5-core18-all
-    interface: content
-    default-provider: kde-frameworks-5-core18
-    target: kf5
+    command: swipl-win
+    extensions:
+      - kde-neon
+    plugs:
+      - home
+      - network
+      - removable-media
+      - opengl
+      - audio-playback
+      - audio-record
 
 grade: stable
 confinement: strict
@@ -57,7 +59,12 @@ parts:
       # Custom build process to enable PGO build
       mkdir -p $SNAPCRAFT_PART_SRC/build
       cd $SNAPCRAFT_PART_SRC/build
-      cmake -DCMAKE_BUILD_TYPE=Release -DSWIPL_PACKAGES_JAVA=OFF -DCMAKE_INSTALL_PREFIX=/usr -DSWIPL_INSTALL_IN_LIB=ON -G Ninja ..
+      cmake -DCMAKE_BUILD_TYPE=Release \
+      -DSWIPL_PACKAGES_JAVA=OFF \
+      -DCMAKE_INSTALL_PREFIX=/usr \
+      -DSWIPL_INSTALL_IN_LIB=ON \
+      -DCMAKE_FIND_ROOT_PATH=/snap/kde-frameworks-5-core18-sdk/current \
+      -G Ninja ..
       ../scripts/pgo-compile.sh
       ninja
       DESTDIR=$SNAPCRAFT_PART_INSTALL ninja install
@@ -107,6 +114,4 @@ parts:
       - unixodbc-dev
       - zlib1g-dev
       - libyaml-dev
-  kde-frameworks-5-env:
-    plugin: dump
-    source: https://github.com/apachelogger/kf5-snap-env.git
+      - libglvnd-dev


JanWielemaker:
Sorry for the late response.  Building the snap seems to work, but the result does not work with x11.   Running `?- emacs.` on either `swi-prolog.swipl` or `swi-prolog.swipl-win` results in a core dump.  Also the font of `swi-prolog.swipl-win` is not a fixed font, which it is when you build swipl-win from the source.  Finally I'm wondering about the XDG file locations.  It doesn't seem to pick up ``~/.config/swi-prolog``.  

Getting there ...  Would be nice to be able to release the upcoming 8.2 as a snap.

LyzardKing:
It doesn't pick up ~/.config/swi-prolog because the confined snap has it's own "home" directory in `~/snap/swi-prolog/current`
So the xdg locations should be in there..
`~/.config/swi-prolog` -> `~/snap/swi-prolog/current/.config/swi-prolog`
(Internally swipl sees it as the normal `~/.config/swi-prolog` folder, so no adjustments should be necessary in prolog codebases)

JanWielemaker:
> Regarding emacs...I'm not sure..could it be that a dependency is missing in the snap?

Good question.   It is clearly related to text.  See below.   Creating and displaying a simple empty window works fine, so there is X11 connection.   Trying to display a text goes wrong.

```
?- [library(pce)].
true.

?- new(W, window).
W = @11862827323233/window.

?- send($W, open).
W = @11862827323233/window.

?- send($W, display, text("Hello world")).
Segmentation fault (core dumped)
```

I see local/snap.md is a bit old and talks about an `x11` plug, but this doesn't appear in the current yaml file?

LyzardKing:
The x11 plug should be connected by the kde-neon extension.
To see the active connections/plugs you can run `snap connections swi-prolog`

JanWielemaker:
Getting snappy :)  Tried

    snap run --gdb swi-prolog.swipl

and get the stack.  This looks like below.   This happens if even the most default font is not accessible: it opens a font.  On failure it replaces this by a simple default font to show at least something.  But this too fails and then we end up with a loop.   Although it probably makes sense to abort when opening the default font fails, we are still faced with the situation that there are apparently no fonts :cry: 

```
#3  0x00007ff8d8429191 in FcFontMatch () from target:/snap/swi-prolog/x1/usr/lib/x86_64-linux-gnu/libfontconfig.so.1
#4  0x00007ff8d865d35b in XftFontMatch () from target:/snap/swi-prolog/x1/usr/lib/x86_64-linux-gnu/libXft.so.2
#5  0x00007ff8d91c0531 in ws_create_font () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#6  0x00007ff8d90e91cc in XopenFont () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#7  0x00007ff8d9120990 in pceExecuteGoal () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#8  0x00007ff8d9120ee8 in vm_send () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#9  0x00007ff8d9121a7c in sendPCE () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#10 0x00007ff8d9134ad3 in getXrefObject () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#11 0x00007ff8d90e9137 in replaceFont () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#12 0x00007ff8d91c05b5 in ws_create_font () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#13 0x00007ff8d90e91cc in XopenFont () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#14 0x00007ff8d9120990 in pceExecuteGoal () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#15 0x00007ff8d9120ee8 in vm_send () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#16 0x00007ff8d9121a7c in sendPCE () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#17 0x00007ff8d9134ad3 in getXrefObject () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
#18 0x00007ff8d90e9137 in replaceFont () from target:/snap/swi-prolog/x1/usr/lib/swipl/lib/x86_64-linux/pl2xpce.so
```

JanWielemaker:
Hmmm.  Fonts seem a common problem.  We find

  - https://forum.manjaro.org/t/system-fonts-on-snap-applications/134451

It only doesn't really explain the issue to me.  It seems strange to remove/rebuild the system font cache if snap uses its own fonts?

LyzardKing:
Out of curiosity I tried building a classic snap, with full access to the system.
The cli and qt ui start, but I haven't tested them yet..
If I try to run emacs in swipl I get this error:

```
?- emacs.
[PCE fatal: @helvetica_roman_12/font: Xopen failed on @display/display
	in: 	<No exception goal>
]
	[13] M @helvetica_roman_12/font ->_xopen(@display/display)
	[12] M @helvetica_roman_12/font ->_xopen(@display/display)
	[11] M @17525740126718/reporter <-height()
	[10] M @17525740126980/point ->initialise(0, (nil))
	[ 9] V @17525740126718/reporter ->reference((nil))
	[ 8] M @17525740126718/reporter ->initialise()
	[ 7] M @17525740087830/dialog ->append((nil), (nil))
	[ 6] M @emacs_mark_list/emacs_bookmark_editor ->fill_dialog()
	[ 5] M @emacs_mark_list/emacs_bookmark_editor ->initialise()
	[ 4] 
	[ 3] M @emacs/emacs ->initialise(@emacs_buffers/dict)
	[ 2] 
	[ 1] M @prolog/host ->call(trap_ref, emacs)
Host stack:
 [21] pce_principal:send(@17525740126718/reporter, reference(point(0, ?(@17525740126718/reporter, height))))
 [19] '$c_call_prolog'
 [18] pce_principal:send(@17525740087830/dialog, append(new(reporter), right))
 [17] Send-method on @emacs_mark_list/emacs_bookmark_editor: emacs_bookmark_editor->fill_dialog
 [16] '$c_call_prolog'

```

LyzardKing:
Ignore the previous comment..
I didn't reinstall properly.
The classic snap seems to work as intended with emacs.
Still testing the qt interface...

JanWielemaker:
As a PR, this is all handled AFAIK.  Remaining is the snap fontconfig issue, but that should not stop us closing this PR.

PR #578, opened 2020-05-10
Add CMake export config to SWIPL install.

sei-mwd:
This adds files in lib/cmake/swipl which can be used by CMake in an
external project to automatically find the swipl program or library as
needed, including dependencies such as include paths.

This exports two targets: `swipl::swipl` and `swipl::libswipl`.  The latter, if linked with using target_add_libary, will automatically handle the location of the library and adding the correct include directories.  I also have it export SWIPL_EXECUTABLE, which should be a value that should be useable for `argv[0]` in `PL_initialise()`.

I do not have the ability right now to test this feature for Windows or on OS X.


JanWielemaker:
Thanks!  Wanted that for a long time, but didn't find the time to figure out how to do this properly.  Merged after rebase and edit the commit message to start with `[A-Z]+:`, so it gets added to the auto generated ChangeLog.    That upsets github, but it is merged ...

PR #579, opened 2020-05-13
Bugfix/swipl install cmake config dir windows

pd-robgee:
I noticed that the CMake variables `SWIPL_CMAKE_NAMESPACE` and `SWIPL_INSTALL_CMAKE_CONFIG_DIR` were only set when not building for Windows.
This led to an error:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.16/Modules/CMakePackageConfigHelpers.cmake:240 (message):
  No INSTALL_DESTINATION given to CONFIGURE_PACKAGE_CONFIG_FILE()
Call Stack (most recent call first):
  src/CMakeLists.txt:506 (configure_package_config_file)
```

This probably happened because these two lines were accidentally included inside a non-indented `if()`, so I indented that block while I was at it.


pd-robgee:
Don't mind the long list of commits that GitHub decides to show, they're mainly from merges.
Only one file changed.
The actual bugfix is: https://github.com/SWI-Prolog/swipl-devel/pull/579/commits/c78c46e743d8d7e7bf6e46055c496e060482530f

JanWielemaker:
Thanks.  Cherry picked the right commit.  Please make PR requests relative to the central master.   Typically I keep `master` in sync with the upstream and use some branch to maintain my personal
version.   For master, simply use `git reset --hard origin/master` to properly sync (make sure all changes you want to keep are elsewhere).

pd-robgee:
No problem.

I thought I'd make a clean PR by merging `SWI-Prolog:master` into `pd-robgee:master` (in https://github.com/SWI-Prolog/swipl-devel/pull/579/commits/4102671f5f546686424387d2e6d78e33143a62a6) before opening.
The diff between `pd-robgee:bugfix/swipl-install-cmake-config-dir-windows` and `SWI-Prolog:master` was actually exactly the content I wanted in the PR, only the list of commits that GitHub shows isn't what I expected.
(So I *think* merging without cherry-picking would've been fine).
Apparently GitHub doesn't consider the two in sync, even after I did that `SWI-Prolog:master` =>`pd-robgee:master` merge.

I'll try `git reset --hard origin/master` next time to see if that avoids the issue.
Thanks for the tip!

Anyway, maybe you'd like to cherry-pick https://github.com/SWI-Prolog/swipl-devel/pull/579/commits/9c621cdaff282d02a9b6d5b92a51e1bb1880822a as well?
I just highlighted the other commit, since it kind of blends into the fixing of the indentation.

image

JanWielemaker:
(cherry picked the indentation patch.  thanks)

I'm a little confused.  It seems to me that the safest way to get large file support on 32-bit systems (assuming it is always there on 64-bit systems) is to use the *64() functions, i.e., have `IOAPI_NO_64` not defined.  The backup is to rely on ftello/fseeko() (I think we only need to check one, it would be pretty strange for an OS to have the one and not the other). 

What is the root of this problem?   Do I understand CMake with MSVC claims fopen64() exists while in fact it does not?   Or does this platform provide fopen64() but not ftello64()?

To some extend this is all a bit theoretical.   I very much doubt anyone on a 32-bit system would like to have a saved state of more than 2Gb, although you might want to put videos in there :smile:  The system should compile though ...

P.s. On the MinGW32 build I have

```
janw (build.win32; master) 7_> grep HAVE_FOPEN64 CMakeCache.txt 
HAVE_FOPEN64:INTERNAL=1
```

(re-opening until we agree)

pd-robgee:
Ah, so the actual issue is probably that `src/minizip/ioapi.c` shouldn't have been touched in https://github.com/SWI-Prolog/swipl-devel/commit/bb2b7ad89db0976028d28f6b7c97ca3712db0b98.
See below for what I was about to comment, before I found that out.
The clue is in `src/minizip/ioapi.h`:
```C
#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(__APPLE__)) && (!defined(__ANDROID__))

  // Linux needs this to support file operation on files larger then 4+GB
  // But might need better if/def to select just the platforms that needs them.

        #ifndef __USE_FILE_OFFSET64
                #define __USE_FILE_OFFSET64
        #endif
        #ifndef __USE_LARGEFILE64
                #define __USE_LARGEFILE64
        #endif
        #ifndef _LARGEFILE64_SOURCE
                #define _LARGEFILE64_SOURCE
        #endif
        #ifndef _FILE_OFFSET_BIT
                #define _FILE_OFFSET_BIT 64
        #endif

#endif

#include <stdio.h>
#include <stdlib.h>
#include "zlib.h"

#if defined(USE_FILE32API)
#define fopen64 fopen
#define ftello64 ftell
#define fseeko64 fseek
#else
#if defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) \
	|| defined(__HAIKU__) || defined(__ANDROID__)
#define fopen64 fopen
#define ftello64 ftello
#define fseeko64 fseeko
#endif
#ifdef _MSC_VER
 #define fopen64 fopen
 #if (_MSC_VER >= 1400) && (!(defined(NO_MSCVER_FILE64_FUNC)))
  #define ftello64 _ftelli64
  #define fseeko64 _fseeki64
 #else // old MSC
  #define ftello64 ftell
  #define fseeko64 fseek
 #endif
#endif
#endif
```

It seems that minizip already tries it best to enable large file support on all platforms it supports.
I *suppose* `IOAPI_NO_64` is an override that can be defined externally to disable LFS if unwanted (for whatever reason...)

This explains why CMake doesn't find `fopen64` and relatives on Windows, they're defines within the minizip project.

Actually, what was the reason behind https://github.com/SWI-Prolog/swipl-devel/commit/bb2b7ad89db0976028d28f6b7c97ca3712db0b98 ?
Because I suggest reverting that commit, as that seems to be the cleanest way to guarantee LFS on all platforms that support it.


#### This is what I was about to comment, just leaving it here for posterity:

> I'm a little confused. It seems to me that the safest way to get large file support on 32-bit systems (assuming it is always there on 64-bit systems) is to use the *64() functions

Ah, yes, I was under the impression that not having `ftello64`/`fseeko64` meant that the non-64 functions had large file support.
I was wrong, so relying on the non-64 functions *only as backup* seems like the better choice.

> What is the root of this problem?

The problem is that CMake does *not* find `fopen64` when making a Visual Studio project, while the *64() functions *are* available.
If I see correctly, before https://github.com/SWI-Prolog/swipl-devel/commit/bb2b7ad89db0976028d28f6b7c97ca3712db0b98, `IOAPI_NO_64` was never set, and so the 864() functions were used under MSVC.

FYI, these are the check results under MSVC:
```
-- Looking for fopen - found
-- Looking for ftello - not found
-- Looking for fseeko - not found
-- Looking for fopen64 - not found
-- Looking for ftello64 - not found
-- Looking for fseeko64 - not found
```
Of which I believe the last three to be incorrect.

And under GCC:
```
-- Looking for fopen - found
-- Looking for ftello - found
-- Looking for fseeko - found
-- Looking for fopen64 - found
-- Looking for ftello64 - found
-- Looking for fseeko64 - found
```

> To some extend this is all a bit theoretical.

Doesn't mean we shouldn't try to make it correct :smiley:



Both http://coding.derkeiler.com/Archive/C_CPP/comp.lang.c/2006-12/msg03560.html and https://stackoverflow.com/a/32924097 seem to indicate that under Windows `_ftelli64` and `_fseeki64` should be used instead, with a regular `fopen`.
Both those functions are detected by CMake.
I suppose `ftello64` and `fseeko64` are `#define`s under MSVC, so that they work, but are not detected by CMake, but that's just a theory.
Anyway, I suppose we do something like this instead:
```C
#if defined(HAVE_FOPEN64)
  #define FOPEN_FUNC(filename, mode) fopen64(filename, mode)
  #define FTELLO_FUNC(stream) ftello64(stream)
  #define FSEEKO_FUNC(stream, offset, origin) fseeko64(stream, offset, origin)
#else
  #define FOPEN_FUNC(filename, mode) fopen(filename, mode)
  #if defined(HAVE__FTELLI64)
    #define FTELLO_FUNC(stream) _ftelli64(stream)
    #define FSEEKO_FUNC(stream, offset, origin) _fseeki64(stream, offset, origin)
  #else
    #define FTELLO_FUNC(stream) ftello(stream)
    #define FSEEKO_FUNC(stream, offset, origin) fseeko(stream, offset, origin)
  #endif
#endif
```

JanWielemaker:
Thanks for looking into this.   bb2b7ad89db0976028d28f6b7c97ca3712db0b98 was the result of yet another platform for which we had to make some exception.   The minizip code is copied from the zlib project and is basically a not that well maintained example program.   It has been modified heavily to get its data from the SWI-Prolog context and allow for multi-threaded access to the Prolog resources that are represented using it.

Porting using if WIN32 || ANDROID && !...  is just asking for trouble.   So, I want to move the configuration to use more CMake detects rather than adding more conditions.

The core CMake already does detect how large file support is to be enabled that is already used in the rest of the code.  I think we should also use that for the minizip files.

As for detecting of functions that are in fact macros, CMake can handle this fine as it just creates a small C program, compiles and links it.   If I recall well one needs to set some CMake variable with the header files to include for testing the availability of functions.

How does this change your opinion?   What can/do you want to/... do to help?  I'm happy to give it a try, but I only test on Linux, MinGW and MacOS.

pd-robgee:
Thanks for the explanation.

> Porting using if WIN32 || ANDROID && !... is just asking for trouble. So, I want to move the configuration to use more CMake detects rather than adding more conditions.

Sure, probably stems from minizip not using CMake. So I agree that using CMake detects is inherently better since we're using CMake. It however means that `src/minizip/ioapi.h` needs to be updated as well in order to remain in sync with `src/minizip/ioapi.c`.

> The core CMake already does detect how large file support is to be enabled that is already used in the rest of the code. I think we should also use that for the minizip files.

I assume you are talking about `cmake\TestLargeFiles.cmake`?
Using that would certainly make it more consistent, but the Windows tests are commented out.
Other than that I think it should provide all defines necessary to convert `src/minizip/ioapi.h` (as well as `src/minizip/ioapi.c`).
Any idea why the Windows tests are commented out?

> As for detecting of functions that are in fact macros, ...

Ok, so this is a fallback option, but if `cmake\TestLargeFiles.cmake` can be used, this shouldn't be necessary.

> How does this change your opinion? What can/do you want to/... do to help? I'm happy to give it a try, but I only test on Linux, MinGW and MacOS.

If there is an easy way to re-enable the Windows tests in `cmake\TestLargeFiles.cmake`, I'd be happy to propose a patch for the ioapi files.
I can test it under MSVC and GCC, if you can then test it under MinGW and MacOS, we should have most bases covered.

JanWielemaker:
> It however means that `src/minizip/ioapi.h` needs to be updated as well in order to remain in sync with `src/minizip/ioapi.c`.

Agree.   My old patch was too simpleminded.

> Any idea why the Windows tests are commented out?

No.  I did copy the used file to the cmake dir.  If you pull and uncomment the Windows test you can see whether it works.  I'd assume it can't do much harm.

> I'd be happy to propose a patch for the ioapi files.

Thanks a lot.  I propose to simply uncomment the tests and see how far you get.

pd-robgee:
I have updated this PR with https://github.com/SWI-Prolog/swipl-devel/pull/579/commits/b6550ffc979fd1696d371e41a0c79de8acb21d65 (I also noticed that we're working on pd-robgee:bugfix/swipl-install-cmake-config-dir-windows, instead of on pd-robgee:bugfix/fopen64-on-windows 😅 ).

It builds on MSVC and GCC, I haven't run the test suite.
Could you verify if you agree with the logic, and check if it builds (and tests) under MinGW and on MacOS?

pd-robgee:
> But this commit removes HAVE_FOPEN64 from src/config.h.cmake. Is that intended?

That was intended, but I notice I still use it instead of `OPJ_HAVE_FSEEKO`, will update...

pd-robgee:
Or is it maybe better to put that check back in and continue to use `HAVE_FOPEN64`?
It'll then not come from `TestLargeFiles.cmake`, but it matches the intended purpose better.
I'm not quite sure, what's your opinion?

JanWielemaker:
> I'm not quite sure, what's your opinion?

I don't really know.  The minizip code is really old.  Some of the stuff they try to support is really dead.  I think following what the CMake test suggests is the best way to go (and hope for the best).

JanWielemaker:
Just noticed the Linux error and your new patch.   From the Linux docs one gets the impression that using ftello(), etc. is the preferred way of doing things.    Shouldn't we first check for that?

JanWielemaker:
A bit more browsing seems to suggest that on Windows we need _ftelli64() and on any system that has off_t we can probably quite safely assume we can use ftello().   Shall I take it from here?

pd-robgee:
> Just noticed the Linux error and your new patch. From the Linux docs one gets the impression that using ftello(), etc. is the preferred way of doing things. Shouldn't we first check for that?

I'm not quite sure about this, I would've thought that using *64() when available would be the safest bet.
But https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html seems to indicate that with `_FILE_OFFSET_BITS` set to 64 also the non-64 functions support large files.
If `_FILE_OFFSET_BITS` is always available when the *64() functions are and vice versa, I don't think we even need to check for the *64() functions.

> A bit more browsing seems to suggest that on Windows we need _ftelli64()

That's what I found out as well, and I have incorporated that in this PR

> and on any system that has off_t we can probably quite safely assume we can use ftello().

I *think* that is a safe bet indeed.

> Shall I take it from here?

That would be great, if you need any help, please just ask. 😄

JanWielemaker:
> That would be great,

Please check https://github.com/SWI-Prolog/swipl-devel/commit/130d268ae0253625d8af00b83bdbb6ce0ee49459

Works for me on Linux, MacOS and MinGW 32/64 bits.

Any other commits I should cherry pick?  I lost track :(  Sorry ...

pd-robgee:
Looks good, except `SIZEOF_OFF_T` is set to empty instead of `0` when `off_t` is not present.
I think it is better to do `#if defined(HAVE_SIZEOF_OFF_T)` instead.
See also: https://stackoverflow.com/a/50156365 and https://cmake.org/cmake/help/latest/module/CheckTypeSize.html

> Any other commits I should cherry pick? I lost track :( Sorry ...

No problem! You haven't missed any commits.

pd-robgee:
And it still builds on both platforms at my side as well! 😄

JanWielemaker:
> I think it is better to do `#if defined(HAVE_SIZEOF_OFF_T)` instead.

Its not that simple because an empty string is considered defined :cry:  Updated the commit and pushed to master.  If all your problems are resolved :smile:, please reset your master using `git reset --hard origin/master`, so all subsequent PRs are nice and clean again.

Thanks for taking all the trouble.  Sometimes simple things turn out more complicated than hoped for ...

pd-robgee:
> Its not that simple because an empty string is considered defined

I think you misunderstood, I suggest to use a different CMake variable (*with* a `HAVE_` prefix) that is also automatically generated by the CHECK_TYPE_SIZE CMake test.
`SIZEOF_OFF_T == 0` actually means "type has arch-dependent size" instead of "type does not exist", see the CMake documentation https://cmake.org/cmake/help/latest/module/CheckTypeSize.html

Although the check should probably be `#if HAVE_SIZEOF_OFF_T == TRUE`

> please reset your master using `git reset --hard origin/master`, so all subsequent PRs are nice and clean again.

Will do 😉 

> Sometimes simple things turn out more complicated than hoped for ...

Definitely :)

JanWielemaker:
You are right.  Changed to `#if defined(HAVE_SIZEOF_OFF_T)`, which I think is good enough and a bit easier to read ...  Seems we are done (for now).

P.s. How much actually works using MSVC?  Just the core?

pd-robgee:
Great, thanks!

> Seems we are done (for now).

Agreed 😄 

> P.s. How much actually works using MSVC? Just the core?

Pretty much, for reference, this is what I build with:
```
cmake
  -DCMAKE_CROSSCOMPILING_EMULATOR=
  -DCMAKE_INSTALL_PREFIX=../install-swipl
  -DMINGW_ROOT=<path to zlib>
  -DMULTI_THREADED=OFF
  -DUSE_SIGNALS=ON
  -DUSE_GMP=OFF
  -DSWIPL_SHARED_LIB=OFF
  -DSWIPL_VERSIONED_DIR=OFF
  -DSWIPL_INSTALL_IN_LIB=OFF
  -DSWIPL_M32=OFF
  -DSWIPL_PACKAGES=OFF
  -DSWIPL_PACKAGE_LIST=cpp
  -DINSTALL_DOCUMENTATION=OFF
  -DBUILD_PDF_DOCUMENTATION=OFF
  -DBUILD_MACOS_BUNDLE=OFF
  -DBUILD_TESTING=OFF
  -DBUILD_SWIPL_LD=OFF
  -DINSTALL_TESTS=OFF
  ../swipl-devel/
```

JanWielemaker:
Thanks.  You surely can get a lot more working.   Notably I think multi threading should work (although it might be broken again since the recent move from old GCC legacy atomic instructions to the more widely supported __atomic_* API, but that should be easy to fix).   GMP works too, provided you either use the MinGW compiled one with some trickery to create a .LIB file or you find a MSVC compiled one (these surely used to exist; I've been using MSVC until about 5 (?) years back).  

I dropped MSVC to avoid the trouble of maintaining two sets of make files (for GNU make and nmake), the (then) lack of C99 support and the fact that the MinGW version was about 30% faster.  CMake resolved the Makefile nightmare.  I have no clue how the performance compares these days.

The main reason for MinGW now is that it allows cross-compiling from Linux which seriously simplifies my workflow as I can quickly compile and test both versions from the same source tree and can create all binaries except for MacOS from one simple script.

pd-robgee:
Thanks for the hints. However I have disabled multi-threading, since it depends on libwinpthread on Windows, and I don't (yet) necessarily need it. Furthermore GMP is mainly disabled due to the licensing, and I also don't need it since I hardly do any arithmetic within Prolog. This configuration was the quickest and easiest way to get up and running.

> I dropped MSVC to avoid the trouble of maintaining two sets of make files

Understandable

> I have no clue how the performance compares these days.

Probably about the same. If with "about 30% faster" you actually meant "runs in about 30% of the time"

> The main reason for MinGW now is that it allows cross-compiling from Linux which seriously simplifies my workflow as I can quickly compile and test both versions from the same source tree and can create all binaries except for MacOS from one simple script.

Understandable, although you might want to look into the concept of Continuous Integration (CI) if you're serious about supporting multiple platforms.
I think it is paid, and I don't have experience with it, but GitHub has so called "GitHub Actions".
Please ignore this piece of "ongevraagd advies" if you're not interested 😃

pd-robgee:
I'd have hoped we'd be done. Unfortunately it seems that MSYS2 MinGW 64-bit acts a bit odd.
It has `SIZEOF_OFF_T == 4`, no `HAVE__FSEEKI64`, no `HAVE_FOPEN64`, `OPJ_HAVE_FSEEKO == OFF` and no `ftello`/`fseeko`.
It basically appears to not have LFS (which I'd thought it should).
I'm now updating it and will investigate the LFS.
But I think it might be best to make the preprocessor block in `ioapi.c` the following, to avoid trying to enable LFS if not available.
Apparently our assumption `HAVE_SIZEOF_OFF_T` => `HAVE_FTELLO` doesn't always hold.
For this to work, an entry will need to be made in `config.h.cmake` for `OPJ_HAVE_FSEEKO`.
```C
#if defined(HAVE__FSEEKI64)
  #define FOPEN_FUNC(filename, mode) fopen(filename, mode)
  #define FTELLO_FUNC(stream) _ftelli64(stream)
  #define FSEEKO_FUNC(stream, offset, origin) _fseeki64(stream, offset, origin)
#elif defined(HAVE_FOPEN64)
  #define FOPEN_FUNC(filename, mode) fopen64(filename, mode)
  #define FTELLO_FUNC(stream) ftello64(stream)
  #define FSEEKO_FUNC(stream, offset, origin) fseeko64(stream, offset, origin)
#elif defined(OPJ_HAVE_FSEEKO)
  #define FOPEN_FUNC(filename, mode) fopen(filename, mode)
  #define FTELLO_FUNC(stream) ftello(stream)
  #define FSEEKO_FUNC(stream, offset, origin) fseeko(stream, offset, origin)
#else
  #define FOPEN_FUNC(filename, mode) fopen(filename, mode)
  #define FTELLO_FUNC(stream) ftell(stream)
  #define FSEEKO_FUNC(stream, offset, origin) fseek(stream, offset, origin)
#endif
```

JanWielemaker:
> I'd have hoped we'd be done. Unfortunately it seems that MSYS2 MinGW 64-bit acts a bit odd.

Got a mail, but can't find it.   Did you delete this post?   My assuption was that using fopen(), ftell() and fseek() is the last resort, not providing LFS (which isn't too bad).

pd-robgee:
Yes, I deleted the comments, because I found out that during CMake configuration it did output:
```
  -- Checking for 64-bit off_t
  -- Checking for 64-bit off_t - present with _fseeki64
  -- Checking for fseeko/ftello
  -- Checking for fseeko/ftello - not found
  -- Large File support - found with _fseeki64
```
But, I can't find `HAVE__FSEEKI64` in `CMakeCache.txt`, so I'm still investigating what's going on exactly.

pd-robgee:
Ah, if I change `set(HAVE__FSEEKI64 1)` in `TestLargeFiles.cmake` to `set(HAVE__FSEEKI64 1 CACHE INTERNAL "")` I do see `HAVE__FSEEKI64` in `CMakeCache.txt` and it is properly defined in `config.h`.
Probably we should add that to all the `set(...)` in `TestLargeFiles.cmake`.
See also: https://stackoverflow.com/questions/37093598/problems-with-cmakedefine-and-set
Why it behaves differently under MSYS2 MinGW 64-bit, I don't know though.

pd-robgee:
No, wait, even without that it *does* get set in `config.h`...
Let me further investigate.

JanWielemaker:
Still, variables that must be used in the end should be cached.   I'll change that.

JanWielemaker:
Pushed

pd-robgee:
Seems like I must have had some stale files somewhere in the project in which I'm embedding Prolog, as I have now done a clean build and Prolog builds fine.
So, sorry for the undue commotion 😅 

> Still, variables that must be used in the end should be cached. I'll change that.

You might want to check the other `set(...)` statements in `TestLargeFiles.cmake` as well, as we're using more of the variables defined there.

JanWielemaker:
> You might want to check the other `set(...)` statements in `TestLargeFiles.cmake` as well, as we're using more of the variables defined there.

Maybe later.  I don't know which ones we actually use and am to busy to figure it out ...   We are warned now :smile:

PR #580, opened 2020-05-13
Bugfix/fopen64 on windows

pd-robgee:
Whether `IOAPI_NO_64` needs to be defined got changed to depend on CMake's `check_function_exists(fopen64 HAVE_FOPEN64)` in https://github.com/SWI-Prolog/swipl-devel/commit
/bb2b7ad89db0976028d28f6b7c97ca3712db0b98.
When building with `cmake  -G "Visual Studio 15 2017" -A Win32` (so a 32-bit target on VS2017) this check returns false.
However `IOAPI_NO_64` needs to be undefined since otherwise linker errors occur since `ftello` and `fseeko` cannot be found.

While debugging this issue I noticed on CMakeLists.txt:179 that the CMake variable `HAVE_LARGE_FILES` is set, but it's never used. Is that correct?


JanWielemaker:
Doesn't that imply we should check for `ftello` and use that?  Unless unavoidable I want tests on detected features rather than _WIN32, etc.

pd-robgee:
Ah, that is indeed a nicer way of going about it (why didn't I think of that...)
> Unless unavoidable I want tests on detected features rather than _WIN32, etc.

Seems like a very sensible protocol. 😄 

I updated the PR, it seems to do what is expected.
Use `fopen`/`fseeko`/`ftello` on Linux and use `fopen64`/`fseeko64`/`ftello64` on a 32-bit Windows target.
Considering https://github.com/SWI-Prolog/swipl-devel/pull/579#issuecomment-628269921, this PR should be able to be merged, if you agree with the new way of checking `fseeko`/`ftello` support.
Or https://github.com/SWI-Prolog/swipl-devel/pull/580/commits/c07b0ea8637893a60c342bd979ee2c5f8ed73bd6 could be cherry-picked instead, if you prefer.

JanWielemaker:
> if you agree with the new way of checking `fseeko`/`ftello` support.

Looks good.   Added an inline comment though about fopen64.

PR #581, opened 2020-05-13
Bugfix/libraries for running build steps use module no single quotes

pd-robgee:
Using single quotes (at least in `QLF.cmake`, I don't build documentation) leads to an error when building on Windows, since the single quotes are passed into the Prolog executable, which then thinks it's a non-compound term with a weird name:
```
-g 'use_module(library(prolog_install))': catch/3: Unknown procedure: 'use_module(library(prolog_install))'/0
```


JanWielemaker:
Thanks.  Think I applied the critical one.

PR #583, opened 2020-05-13
Improved intro text for library(yall) intro

dtonhofer:
I have added more explanations regarding the concept of "closure" as used in Prolog and extended the description of the rest. It feels better and is hopefully clearer to a newcomer, but in the end it's matter of taste. I was unable to justify the text to 73 columns though, I just kept to this as close as possible,

One thing that is missing in markup processing is link-ization of things like maplist/N. I have used maplist/2-5 instead, which linkized maplist/2 properly.


JanWielemaker:
Thanks.  Merged.   Re-formatted the comment, replaced the ≤ by =< as I do not want to have to rely on encodings when avoidable (and thus want sources to be US-ASCII if possible) and modernized a few PlDoc things ``(== --> ```)``.  These things cause github to think it isn't merged, but it is.

In the end it would be better to fix the maplist/N auto linking ...

PR #595, opened 2020-05-19
Using _BitScanReverse/_BitScanReverse64 as according to Microsoft documentation

pd-robgee:
Without this patch I was getting errors about `BitScanReverse(64)` being undefined on a `Windows Server 2012 R2` machine.
```
warning C4013: 'BitScanReverse64' undefined; assuming extern returning int
error LNK2001: unresolved external symbol _BitScanReverse64
```
Funnily enough it was working perfectly on `Windows Server 2016` and `Windows 10`, all using the `Visual Studio 15 2017` CMake generator (sorry, don't know the specific MSVC versions).
According to the Microsoft documentation the intrinsics:
- Start with an underscore
- Require inclusion of `<intrin.h>`
- Require a `#pragma`

This patch makes the usage match the documentation and solves the problem on all platforms.
Documentation: https://docs.microsoft.com/en-us/cpp/intrinsics/bitscanreverse-bitscanreverse64?view=vs-2017

Furthermore `_BitScanReverse64` is only defined on 64-bit targets, so I added a check for that around the definition of `MSB64`.
```
warning C4163: '_BitScanReverse64': not available as an intrinsic function
warning C4013: '_BitScanReverse64' undefined; assuming extern returning int
error LNK2001: unresolved external symbol __BitScanReverse64
```

### DIFF
--- a/src/pl-inline.h
+++ b/src/pl-inline.h
@@ -41,6 +41,11 @@
 #ifdef __WINDOWS__
 #include <windows.h>
 #undef small
+#include <intrin.h>
+#if SIZEOF_VOIDP == 8
+  #pragma intrinsic(_BitScanReverse64)
+#endif
+#pragma intrinsic(_BitScanReverse)
 #endif
 
 		 /*******************************
@@ -67,22 +72,24 @@ MSB(size_t i)
 { unsigned long index;
 #if SIZEOF_VOIDP == 8
   unsigned __int64 mask = i;
-  BitScanReverse64(&index, mask);
+  _BitScanReverse64(&index, mask);
 #else
   unsigned long mask = i;
-  BitScanReverse(&index, mask);
+  _BitScanReverse(&index, mask);
 #endif
 
   return index;
 }
 
+#if SIZEOF_VOIDP == 8
 #define HAVE_MSB64 1
 static inline int
 MSB64(int64_t i)
 { unsigned long index;
-  BitScanReverse64(&index, i);
+  _BitScanReverse64(&index, i);
   return index;
 }
+#endif
 
 #define MEMORY_BARRIER() MemoryBarrier()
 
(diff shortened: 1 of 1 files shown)


JanWielemaker:
On MinGW this results in

     ../src/pl-inline.h:46: warning: ignoring #pragma intrinsic

Is this otherwise ready?  Doesn't merge, but I can fetch your remote and find the right patch ...

pd-robgee:
> On MinGW this results in

Ah, that's too bad, shall I see if there is a macro test to exclude this pragma from MinGW?

> Is this otherwise ready? Doesn't merge, but I can fetch your remote and find the right patch ...

I'm still running the build on Windows 10, just to make sure it's fixed there as well (typed before I remembered to do that 😅 ).
That's why I made it a draft PR. After I mark it as "ready for review" you should be able to simply merge it.

pd-robgee:
Ok, the Windows 10 build indeed also succeeds.
For the MinGW warning, I can put an `#ifndef __MINGW32__` around the pragmas, which should also work for 64-bit MinGW: https://stackoverflow.com/a/17493828
Want me to add that?

JanWielemaker:
> Want me to add that?

Maybe better test for MSVC.  #pragma is AFAIK always compiler specific.

pd-robgee:
> Maybe better test for MSVC. #pragma is AFAIK always compiler specific.

Good thinking, I've added a `#ifdef _MSC_VER`, could you check if that silences the warning on MinGW?

JanWielemaker:
Works.  32-bit has some other issue right now which I should fix before.  I take it from here?

JanWielemaker:
Merged the two commits.  Thanks!

pd-robgee:
Sorry for the late reply, I was having a break.
Thanks for merging!

> 32-bit has some other issue right now which I should fix before.

Not related to this patch, I suppose?

JanWielemaker:
> Not related to this patch, I suppose?

No.  Affecting 32 bit version though, which I do not test often ...

pd-robgee:
Ah, ok.
I'll try putting the new master in production in a moment.
If I encounter any problems I'll probably open a new PR 😉

PR #606, opened 2020-05-22
Fix typos in manual

sddu:
In `man/overview.doc`, line 2239 was a duplicate of the `windows` flag entry at line 2270. Hence it was deleted.


JanWielemaker:
Thanks!   Merged after rebase and moving the flag `wine_version` which was misplaced due to the duplicate `windows` flag you correctly spotted.

PR #610, opened 2020-05-31
DOC: Fix and changes in manual

sddu:
Off-Topic:

`man/windows.doc` is not in the build file.
As a result, the `windows.html` in `plweb-www` is out-of-sync from the current `man/windows.doc`.

However, it's still being referred to in the [download page](https://www.swi-prolog.org/download/stable).

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -6478,7 +6478,7 @@ locale dependent sorting of atoms.
 The predicate format_time/3 can be used to format time and date
 representations, where some of the specifiers are locale dependent.
     \item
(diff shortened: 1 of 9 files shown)


JanWielemaker:
> `man/windows.doc` is not in the build file.
> As a result, the `windows.html` in `plweb-www` is out-of-sync from the current `man/windows.doc`.

Thanks.   Bit hard to sync automatically as these are two unconnected repos :(   Now synced by hand.    Will be on the servers next update.

Thanks again.    This really helps making the docs look more professional!

PR #616, opened 2020-06-17
FIXED: If source(false) is provided as an option, the location of the…

thetrime:
… clause is provided by a file/4 term, but there is no clause of prolog:message_location for file/4, meaning that the prolog:message/2 call for trace_call_to fails, and the message is printed as an unknown message

### DIFF
--- a/library/prolog_codewalk.pl
+++ b/library/prolog_codewalk.pl
@@ -1138,6 +1138,9 @@ prolog:message_location(clause(ClauseRef)) -->
     [ '~w: '-[Name] ].
 prolog:message_location(file_term_position(Path, TermPos)) -->
     message_location_file_term_position(Path, TermPos).
+prolog:message_location(file(Path, Line, _, _)) -->
+    !,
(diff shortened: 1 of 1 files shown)


JanWielemaker:
There is something else going wrong.  The docs say on_trace can produce a location

    file(+File, +Line, -1, _)

And in line 439 of boot/messages we have:

```
swi_location(file(Path, Line, -1, _CharNo)) -->
    !,
    [ '~w:~d: '-[Path, Line] ].
```

That should handle this case.   I guess it is easy enough to figure out what goes wrong, but it needs a test case that you apparently have :)

thetrime:
You don't call swi_location//1 from prolog_codewalk though - you call prolog:message_location//1.

I'm not sure which is the more fundamental thing?

JanWielemaker:
> I'm not sure which is the more fundamental thing?

Its choosing between two bad things.   I've opted for the other, calling $messages:swi_location//1, reasoning that two rules for the same term make it rather unpredictable if one ever decides to change the human message.

PR #630, opened 2020-07-03
DOC: Reviewed "4.10.3 The exception term"

dtonhofer:
I have reviewed the manual chapter "4.10.3 The exception term" once more because the text didn't make me all that happy, it didn't "flow" well. Also removed the commented-out exception details, no need to keep those around (they are in the repo after all).

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -2543,391 +2543,81 @@ generating a full stack trace on errors in the HTTP server library.
 \label{sec:generalformofexceptionterm}
 
 The predicate throw/1 takes a single argument, the \jargon{exception
-term}:
-
-\begin{code}
-throw(+Exception)
-\end{code}
-
-The ISO Standard stipulates that the exception term \arg{Exception} be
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Yes, looks better.  I'm not totally happy with this style of creating error terms (or terms in general).  

```
    Exception = error(Formal, Context), 
    Context   = ... some local convention ...,
    Formal    = type_error(ValidType, Culprit), % for "type error" for example
    ValidType = integer,                        % valid atoms are listed in the ISO Standard
    Culprit   = ... some value ...,
    throw(Exception)
```

More conventional style is like this.   In this example there is some point in naming the fields using the extra variables.  I still have my doubt about promoting unconventional style though.

```
   throw(error(type_error(integer, a),
               context(name/arity, "You shouldn't do this")))
```

dtonhofer:
I can add a footnote saying "this is for illustration purposes, but..."

PR #632, opened 2020-07-06
DOC: Mostly changes to atom_codes/2 et al.

dtonhofer:
Changes to documentation:

- A large part of changes to https://eu.swi-prolog.org/pldoc/man?section=manipatom "Analyzing and constructing atoms". I actually just wanted to change the misnamed "String" parameter of [`atom_codes/2` ](https://eu.swi-prolog.org/pldoc/doc_for?object=atom_codes/2) but ended up revieweing the rest of the chapter too. The diff is not informative, the new text has to be generated for review.
- Slight changes to the exception pages; "ISO Standard" rewritten as "ISO standard" for consistency, and a note has been added to the example code for exception construction saying that one would probably not write it like this when actually coding.
- Added links to papers to the page on "Delimited Continuations".  The preprint for Schrijvers et al., 2013 can be found on the SWI-Prolog server, the one for Desouter et al., 2015 is freely available at Cambridge University Press. Currently people wanting to take a look at those are directed to the bibliography page, where no live links exist. So they have to duckduckgo themselves. Why not fix that, these are good papers.

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -2328,15 +2328,17 @@ The same warning as for call_cleanup/2 applies.
 
 The predicates reset/3 and shift/1 implement \jargon{delimited
 continuations} for Prolog. Delimited continuation for Prolog is
-described in \cite{DBLP:journals/tplp/SchrijversDDW13}. The mechanism
+described in \cite{DBLP:journals/tplp/SchrijversDDW13} 
+(\href{http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf}{preprint PDF}).
+The mechanism
 allows for proper \jargon{coroutines}, two or more routines whose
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Applied after some edit (hence the conflict).  In particular:

  - Avoid git white-space errors (spaces at the end of a line in this case).
  - Re-flow several paragraphs
  - Flags can be linked to using ``\prologflag(flag)``.
  - Removed quite a few footnotes and remarks from the dark ages.  I think they confuse more than they add.
  - Reorganised the first couple of paragraphs.   Fixed claim of codes being only 0..2^16-1: that only holds on Windows (and is now marked as a bug).

Thanks!   Looks a lot more consistent now and a couple more examples should help novices.

PR #635, opened 2020-07-10
ADDED: Support "file" IRI scheme

royratcliffe:
Hooks IRI-style file system access using the "file" scheme. Useful when dragging files in macOS, for example, where Finder drops the file reference using a file://some-file atom. Maps file-scheme hooks directly to their corresponding built-in file-oriented predicates.


JanWielemaker:
Thanks.   Good to have.  Only, I don't like it in library(files), which is mostly a historical library.  I think I'm in favor of adding a subdir to the library, e.g. `iri_scheme` and then add files named after the scheme, so we have 

    :- use_module(library(iri_scheme/file)).

Would you agree?  If so, you can add an additional commit or leave it to me (also requires src/CMakelist.txt changes)

royratcliffe:
Hi Jan, I agree. My original thoughts were to make a boot/file.pl for a '$file' module, mirroring boot/rc.pl, but that involves `boot.prc`. I like your idea of having an `iri_scheme` sub-directory since it opens the door to supporting other schemes. Happy to leave it to you.

JanWielemaker:
Will take care of this.   Might be a couple of days.

royratcliffe:
Many thanks, Jan. No rush.

JanWielemaker:
Merged after moving the code into iri_scheme/file.pl and squashing the commit.   Might inspire others :)  Thanks.

PR #640, opened 2020-07-18
Fix typos and format summary.doc

sddu:
In `man/threads.doc`, there were 2 predicate name typos which I'm not confident of my fixes:

* Line 142, `setlocale/1` -> `set_locale/1` (might be `setlocale/3`?)

* Line 684, `thread/2` -> `thread_wait/2`


JanWielemaker:
Thanks.  Your guesses are correct.  Thanks also for fixing all of summary.doc.   As some point this was automatically reformatted and that was discovered too long afterwards :(

PR #641, opened 2020-07-19
DOC: Mainly extended "Notation of Predicate Descs"

dtonhofer:
1. Reviewed & extended Section 4.1 "Notation of Predicate Descriptions". In particular:
-  Rewrote the text for mode flags
- Give the description of "predicate indicator" and "non-terminal indicator" their own headings
- Add a section for "determinism" (always a mystery to freshmen and this is probably my 6th revision until I could structure this correctly).
2. Moved section "Printing messages" and "Printing from libraries" _up one level_. Currently they are under section "Exceptions", which is obviously wrong.
3. Two minor typo fixes.
4. Added text to exceptions emphasizing that exceptions say "what would be correct", not "what is wrong" 
5. The adjective "free" is used to describe variables in a few places where "fresh"/"uninstantiated" looks more appropriate. Fixed.
6. Changed the explanation for ground/1 a bit and added a note on anonymous and don't care variables (one of the problems of Prolog-for-beginners is that the expression "the term contains a variable" is exceedingly misleading - indeed, terms contain "holes" and variables name parts of terms, which may or may not be "holes" ... but terms don't contain variables at all, except when represented in written code, which is quite different; I tried to make this cleared in ground/1; it's a bit quixotic)

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -6479,9 +6479,11 @@ ERROR: Domain error: `compound_non_zero_arity' expected, found `a()'
 \end{code}
 
     \predicate{compound_name_arity}{3}{?Compound, ?Name, ?Arity}
-Rationalized version of functor/3 that only works for compound terms
-and can examine and create compound terms with zero arguments (e.g,
-\exam{name()}).  See also compound_name_arguments/3.
+Rationalized version of functor/3 that only works for compound terms.
+A \const{type_error} is thrown if, at call time, \arg{Compound} is
+instantiated to something that is not a compound term. The predicate
+can examine and create compound terms with zero arguments 
+(e.g, \exam{name()}). See also compound_name_arguments/3.
 
     \predicate{compound_name_arguments}{3}{?Compound, ?Name, ?Arguments}
 Rationalized version of \predref{=..}{2} that can compose and decompose
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -1678,10 +1678,12 @@ error condition and do not instantiate their argument. See also library
 
 \begin{description}
     \predicate[ISO]{var}{1}{@Term}
-True if \arg{Term} currently is a free variable.
+True if \arg{Term} currently is an uninstantiated (fresh,unbound) 
+variable.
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.   Need to chew a little on this though.   Just checked _free variable_.  This used throughout the ISO standard for Prolog, so we should keep this.   Clarifications could be given in the glossary.    Also looked at ground.  hat you are describing are properties of read/1 more than terms.   Internally there are no anonymous (_), named anonymous (_Var) or named variables, just variables.    At a glance, I think I can agree with most of the other proposed changes.  Have to check the details.  A general remark is not to use one-sentence paragraphs too often.  Especially in the PDF docs that looks rather weird.

dtonhofer:
Thanks Jan. Agree differentiation between variables is not made internally, they are "language-level curlicues" but nonetheless still have their interpretative effects. No stress, I will be programming a bit (at some point I will have to disengage from Prolog to get ahead in statistics again ... )

dtonhofer:
Yes, agree for this being too much. Writing it using something like a "typing notation" might be interesting, but is there some defined way of doing it?

JanWielemaker:
> but is there some defined way of doing it?

Not really.  There are a good deal of type notations around in the Prolog world, most of which are fairly similar.   SWI-Prolog for now accepts any term as a type description for documentation.   For this purpose I would  not worry too much, just create a set of mode descriptions for all built-in and library predicates based on some invented type description that seems to fit the purpose.   The formal meaning can be decided later.  Having a consistent machine readable set makes it fairly easy to automatically modify the descriptions later as well.

JanWielemaker:
I have merged this PR after some significant editing.   Kept _free variable_ consistent and shortened a number of explanations you added.   Some of the explanations were not really correct and others do (in my view) not belong in a reference manual.

I guess we need a better overall way to deal with terminology issues, for example by completing the glossary and adding many more links from the text to the glossary.  Possibly we could do so automatically (not sure how many ambiguous terms there are that could lead to wrong links).

PR #663, opened 2020-09-03
CLEANUP: memberchk/2 was still exported from lists.pl

dtonhofer:
Looks like memberchk/2 was still being exported though not defined. Noticed this by https://eu.swi-prolog.org/pldoc/doc/_SWI_/library/lists.pl which complains about the undocumented memberchk/2.

### DIFF
--- a/library/lists.pl
+++ b/library/lists.pl
@@ -35,7 +35,6 @@
 
 :- module(lists,
         [ member/2,                     % ?X, ?List
-          memberchk/2,                  % ?X, ?List
           append/2,                     % +ListOfLists, -List
           append/3,                     % ?A, ?B, ?AB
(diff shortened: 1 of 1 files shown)


JanWielemaker:
This export was actually added recently :smile:   The reason is portability.  memberchk/2 is built-in, but SWI-Prolog allows exporting predicates that are built-in.   Doing it this way ensures code doing

    :- use_module(library(lists), [memberchk/2]).

runs unmodified on most Quintus derived Prolog systems.

Possibly we can add a documentation record.   Not sure whether that works.  Or at a rule to the undocumented that built-in predicates are never undocumented.

dtonhofer:
Well, that explains it! Adding an documentation record without code doesn't seem to work. Oh well.

PR #676, opened 2020-09-24
added prod_list/2

damianoazzolini:
Since i often do artithmetic operations in Prolog, i've added the predicate prod_list/2 that computes the product of the elements of a list. Its implementation is intuitive. If the list is empty, i've opted to set the result to 1 since 1 is the neutral element of multiplication.

### DIFF
--- a/library/lists.pl
+++ b/library/lists.pl
@@ -62,6 +62,7 @@
 
                                         % Lists of numbers
           sum_list/2,                   % +List, -Sum
+          prod_list/2,                  % +List, -Prod
           max_list/2,                   % +List, -Max
           min_list/2,                   % +List, -Min
           numlist/3,                    % +Low, +High, -List
@@ -551,6 +552,21 @@ sum_list([], Sum, Sum).
 sum_list([X|Xs], Sum0, Sum) :-
     Sum1 is Sum0 + X,
     sum_list(Xs, Sum1, Sum).
+
+%!  prod_list(+List, -Sum) is det.
+%
+%   Prod is the result of multiplying all numbers in List.
+
+prod_list(Xs, Prod) :-
+    prod_list(Xs, 1, Prod).
+
+prod_list([], Prod, Prod).
+prod_list([X|Xs], Prod0, Prod) :-
+    (   X =\= 0 ->
+    	Prod1 is Prod0 * X;
+    	prod_list([],0,Prod)
+    ),
+    prod_list(Xs, Prod1, Prod).
 
 %!  max_list(+List:list(number), -Max:number) is semidet.
 %
--- a/library/lists.pl
+++ b/library/lists.pl
@@ -556,6 +556,8 @@ sum_list([X|Xs], Sum0, Sum) :-
 %!  prod_list(+List, -Prod) is det.
 %
 %   Prod is the result of multiplying all numbers in List.
+%   If List = [], Prod is 1 since it is the neutral 
+%   element of multiplication
 
 prod_list(Xs, Prod) :-
     prod_list(Xs, 1, Prod).
--- a/library/lists.pl
+++ b/library/lists.pl
@@ -553,7 +553,7 @@ sum_list([X|Xs], Sum0, Sum) :-
     Sum1 is Sum0 + X,
     sum_list(Xs, Sum1, Sum).
 
-%!  prod_list(+List, -Sum) is det.
+%!  prod_list(+List, -Prod) is det.
 %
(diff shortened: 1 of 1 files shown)


library/lists.pl
```diff
@@ -552,6 +553,21 @@
     Sum1 is Sum0 + X,
     sum_list(Xs, Sum1, Sum).
 
+%!  prod_list(+List, -Sum) is det.
+%
+%   Prod is the result of multiplying all numbers in List.
+
+prod_list(Xs, Prod) :-
```
kamahen, line 560: I suggest you add some test cases.

For example:
```
?- debug, prod_list([1,2,0,3], Z).
ERROR: Arguments are not sufficiently instantiated
ERROR: In:
ERROR:   [16] _8848 is _8854*3
ERROR:   [15] prod_list([3],_8882,0) at /tmp/test_prod.pl:5
ERROR:   [14] prod_list([0,3],2,0) at /tmp/test_prod.pl:10
ERROR:   [13] prod_list([2,0|...],1,0) at /tmp/test_prod.pl:10
ERROR:   [12] prod_list([1,2|...],1,0) at /tmp/test_prod.pl:10
ERROR:   [11] prod_list([1,2|...],0) at /tmp/test_prod.pl:2
ERROR:   [10] '<meta-call>'(user:user: ...) <foreign>
ERROR:    [9] toplevel_call(user:user: ...) at /usr/lib/swi-prolog/boot/toplevel.pl:1113
```

library/lists.pl
```diff
@@ -552,6 +553,21 @@
     Sum1 is Sum0 + X,
     sum_list(Xs, Sum1, Sum).
 
+%!  prod_list(+List, -Sum) is det.
+%
+%   Prod is the result of multiplying all numbers in List.
```
kamahen, line 558: You should specify the result if List is [].

In essence, you're defining:
```
prod_list(List, Prod) :- foldl(times, List, Prod).
times(X, Y, Prod) :- Prod is X * Y.
```

library/lists.pl
```diff
@@ -552,6 +553,21 @@
     Sum1 is Sum0 + X,
     sum_list(Xs, Sum1, Sum).
 
+%!  prod_list(+List, -Sum) is det.
```
kamahen, line 556: `Prod`, not `Sum`.

library/lists.pl
```diff
@@ -552,6 +553,21 @@
...
+%
+%   Prod is the result of multiplying all numbers in List.
+
+prod_list(Xs, Prod) :-
+    prod_list(Xs, 1, Prod).
+
+prod_list([], Prod, Prod).
+prod_list([X|Xs], Prod0, Prod) :-
+    (   X =\= 0 ->
```
kamahen, line 565: This optimization should be used only if it can be justified by benchmarks. My guess is that it isn't justified.

kamahen (changes requested):

library/lists.pl
damianoazzolini: Thanks, this was a silly mistake

library/lists.pl
damianoazzolini: Yes, the predicate definition is equivalent. However, when a 0 is present in list, we can stop multiplying all the numbers and return 0. This is not easy to implement with `foldl/N`. Moreover, for consistency, my definition is similar to the one of `sum_list/2` (which can as well be implemented as `foldl(my_sum,L,0,R1)` with `my_sum(A,B,C):- C is A + B.`)

library/lists.pl
damianoazzolini: Thanks, fixed

library/lists.pl
damianoazzolini: In the latest commit there is indeed an improvement. With the 0 check (current version)
```
test:-
    length(L,1000),
    maplist(random(0,3),L), !,
    time(prod_list(L,_)).
```
```
?- test. 
15 inferences, 0.000 CPU in 0.000 seconds (113% CPU, 660153 Lips)
```
without
```
2,001 inferences, 0.000 CPU in 0.000 seconds (101% CPU, 4730183 Lips)
```
results are almost the same in case the list does not contains 0 elements (list of 10 000 elements generated where each element is random between 1 and 3).
```
20,001 inferences, 0.009 CPU in 0.009 seconds (100% CPU, 2214337 Lips) (with)
20,001 inferences, 0.009 CPU in 0.009 seconds (100% CPU, 2349666 Lips) (without)
```

PR #678, opened 2020-09-26
Add indent_string and test cases for indent_string, dedent_string derived from Python

kamahen:
This is a "work-in-progress", but I'm pushing it now for initial review. There's also discussion at the Discourse forum.

PR #679, opened 2020-09-26
Add indent_lines and test cases for indent_lines, dedent_lines derived from Python

kamahen:
This is replacement for PR #678

Some test cases that fail with instantiation error have been commented out, pending a fix.

### DIFF
--- a/library/strings.pl
+++ b/library/strings.pl
@@ -33,8 +33,11 @@
 */
 
 :- module(strings,
-          [ dedent_string/3,            % +In,-Out,+Options
+          [ dedent_lines/3,             % +In,-Out,+Options
+            indent_lines/3,             % +Prefix,+In,-Out
+            indent_lines/4,             % :Pred,+Prefix,+In,-Out
             interpolate_string/4,       % +In,-Out,+Map,+Options
+            split_lines/2,              % +In,-Lines
             string/4                    % Quasi quotation support
           ]).
 :- autoload(library(apply), [include/3, foldl/4, maplist/3, maplist/2]).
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.   Looks good.   There are still two things I'm not happy about.

split_lines/2 has some problems.   First of all the name.   It must be(come) a relation and than we cannot have a verb.    I propose string_lines/2.   Second, it either should do as advertised and not have a "" from after the last \n or we should discard the whole thing.   For now I'm probably mostly in favor of not using it internally as it seems split_string/4 does the right thing.  Next implement it simply on top of split_string/4 and atomics_to_string/3.   That is quite easy: remove the final "" from the list or add it.

Second, less important, please use `test(name, Result == <expected>)` rather than an intermediate predicate using assertion/1.   Tests written this way are not more verbose and lead to better error messages.   Assertion/1 was an after thought, in part to not frustrate people from other languages too much, in part to be able to check multiple assertions in a single test.  The latter is typically not ideal, but pragmatic if testing something requires several state transitions.

I'm happy to take care of especially the first point.

kamahen:
I've change split_lines/2 to string_lines/2 and made it reversible. I suggest you look at the test cases to verify that it does what you expect.

I've also improved the handling of "\n" for indent and dedent.
I'm waiting for @JanWielemaker to fix `dedent("", _, [])` before improving dedent's behaviour with empty lines (plus there are some test cases that are commented out because of this bug).

I've removed all uses of assertion/1 from the tests, and have used fixme in one place.

JanWielemaker:
Thanks.   Notably a lot of test cases help :)  I've made several further changes, fixed the `dedent("", _, [])` cases (and some related using only black lines).   Changed string_lines/2 for mode (-,+) to add a final newline and thus be a true relation.  Change the docs for that as well as indent_lines/4.  

Merged the whole stuff.   I'm not claiming this is the final word on this library, but I think it starts to make sense :smile:   Thanks for all the input.

(and ignore the conflict; github doesn't deal well with manually rebased and modified patches).

kamahen:
Thank-you for improving my code and comments.

PR #682, opened 2020-09-28
Remove back quotes from ansi_get_color_/2

fnogatz:
`ansi_get_color_/2` in `library/ansi_term.pl` currently uses back quotes to get the character codes from strings. This is incompatible with SWI-Prolog's `--traditional` command line flag, which currently throws a warning as follows when called:

```sh
> swipl --traditional
ERROR: library/ansi_term.pl:447:13: Syntax error: Operator expected
Welcome to SWI-Prolog (threaded, 64 bits, version 8.3.8)
...
```

The back quotes can be replaced by 0'... character code sequences instead which preserves compatibility with the ISO standard and versions 7 and 8 of SWI-Prolog.


JanWielemaker:
Thanks!  Handled, but a little different.   See 1252e3b123583ba192eb927ffc004df523f20093.
Note that I do not aim to keep libraries working with `--traditional`.  As this library is loaded by the toplevel when connected to a terminal I think the trouble is justified.

fnogatz:
Alright, thank you, Jan! So my [history](https://github.com/SWI-Prolog/swipl-devel/commits?author=fnogatz) remains of commits related only to corrections of spelling mistakes :stuck_out_tongue_closed_eyes:

JanWielemaker:

https://swi-prolog.discourse.group/t/porting-from-sicstus-prolog-double-quoted-strings-and-traditional/3330/1

PR #689, opened 2020-10-10
DOC: Section "Dicts: structures with named arguments" rewritten

dtonhofer:
I have reviewed and in the end essentially rewritten (sorry, I got dragged along) the introduction on dicts.

- Examples reviewed and extended
- Text made more extensive. The original text was hard to understand when not coming from the design team.
- Aspects that were not readily apparent got highlighted: 
   - Functions calls in heads (I just recently noticed this is possible, and then it occurred to me that this is a very peculiar thing which in reality nukes the whole syntax-only unification principle. One could even use it as "attributed variables made apparent")
   - Added a note on "eager evaluation"
   - The syntax with the `:=` verb (which I always forget about) 
   - The fact that dicts are linked to modules via their tags.

... That's about it. 

But there really needs to be a name for these function calls (like "dot-calls") and a "predicate" indicator (like `.foo/n`).

### DIFF
--- a/man/extensions.doc
+++ b/man/extensions.doc
@@ -726,110 +726,196 @@ Below is an example that illustrates the representation of a typical
 \section{Dicts: structures with named arguments}
 \label{sec:bidicts}
 
-SWI-Prolog version~7 introduces dicts as an abstract object with a
-concrete modern syntax and functional notation for accessing members and
-as well as access functions defined by the user. The syntax for a dict is
-illustrated below. \arg{Tag} is either a variable or an atom. As with
-compound terms, there is \textbf{no} space between the tag and the
-opening brace. The keys are either atoms or small integers (up to
-\prologflag{max_tagged_integer}). The values are arbitrary Prolog terms
-which are parsed using the same rules as used for arguments in compound
-terms.
+SWI-Prolog version~7 introduces the \jargon{dict} as an abstract object
+with a concrete modern syntax. It also adds a functional notation \const{dict.key} 
+for accessing dict entries and for calling ``functions associated to
+a dict`` \const{dict.func(x)} which resolve to user-defined predicates.
+
+The syntax for defining a dict (a \jargon{dict literal}) is as follows:
 
 \begin{quote}
 Tag\{Key1:Value1, Key2:Value2, ...\}
 \end{quote}
 
-A dict can \emph{not} hold duplicate keys. The dict is transformed into
+\begin{itemize}
+\item The dict's \arg{tag} is generally either an unbound variable (then
+we have an \jargon{anonymous dict}) or an atom, but arbitrary terms are 
+allowed. Similar to compound terms, there is \textbf{no} space between the
+dict's \jargon{tag} and the subsequent opening brace. 
+\item The dict's \jargon{keys} are either atoms or ``small'' integers.
+\footnote{Integers up to \prologflag{max_tagged_integer} to be precise, 
+which can be quite large. The query 
+\const{?- current_prolog_flag(max_tagged_integer,X).} yields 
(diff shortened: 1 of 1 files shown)


man/extensions.doc
```diff
@@ -894,47 +1068,188 @@ Dicts currently define the following reserved functions:
...
+      house{city:C,street:S,house_number:N,postal_code:X}).
+C = "Jamestown",
+S = "Ash Tree Lane",
+N = "44",
+X = "23081".
+/end{code}
+
+Let's look at the code:
+
+/begin{code}
```
JanWielemaker, line 1199: Should be \begin

JanWielemaker:
Thanks.  I need to read this carefully.  Before doing so I have a general comment though.  Please do not build a story in code blocks using comments.   Instead, use should code blocks and normal running text in between.   That looks a lot better, notably in the PDF version where long code blocks causes poor page layout.

dtonhofer:
Got it. Will re-read and fix.

dtonhofer:
Ok, hopefully better and easier to read and more consistent. Dicts have a lot of implications that are not evident at all when one first sees them.

dtonhofer:
Any news on this one?

JanWielemaker:
This is useful stuff.   I do have some problems with the placement in this section though.   Where the manual is called a _reference manual_, this is more like a tutorial.   That is good to have, especially as we cannot point at some external text book.   At this place though the whole thing becomes rather unbalanced.

I'm inclined to start either a chapter _Tutorials_ and sections in there with for example this, but possibly also the new stuff on the tracer and make references from the reference material to these sections.  Would that work?

Some small points

  - \const is for constants.  \exam is for example code.   They currently render the same, but they are semantically different.
  - I would not stress the tag to allow for arbitrary terms.  Yes, you can unify the variable that is a tag, but that is surely not intended usage and while write/1 writes such terms, read/1 can't read them.   There is merely no sensible way to stop the user from instantiating the tag with an arbitrary term.
  - Expressions such as ``Pt.put([z-3,i-4,j-5,k-6])`` work merely because predicates that expect a dict as argument convert some other common key-value representations (see dict_create/3).   I don't think these should be promoted as good style.
  - I'm reluctant giving too many implementation details.  That invites people to rely on these details, preventing future improvements.
  - Especially for the PDF version, many very short paragraphs read rather poorly.   As my PhD supervisor told me, a paragraph is a collection of sentences :smile:

dtonhofer:
> At this place though the whole thing becomes rather unbalanced.

This is certainly true.

The tutorial pages are the right approach, but it will take a while to set it up and grow it. Meanwhile there is the wiki, too. 
 
I propose to move this text into my GitHub space for general availability and link from the comment section. Once a decision about how to extend the reference manual has been taken, the editor can grab the stuff he/she needs.

JanWielemaker:
> The tutorial pages are the right approach, but it will take a while to set it up and grow it. Meanwhile there is the wiki, too.

True.   We've got tutorial stuff at many places, in various sites.   That is good and bad.   The good part is that apparently various people in the community take the trouble to write tutorials, showing interest and commitment.   The bad part is that there is very little consistency.   It might be wiser to write tutorials in Markdown, as @EricZinda did for the text debugger.   That gives more flexibility (and is more pleasant to type).   I might also move the debugger section to a new tutorials chapter.  Then we already have two and a third on the graphical debugger coming.

dtonhofer:
I have put a copy of the text in 

https://github.com/dtonhofer/prolog_notes/tree/master/swipl_notes/about_dicts/manual_review

We can close this pull request. If there is interest in pieces of the updated text, we can pull copypaste it from teh above.

Issue #690, opened 2020-10-12
WASM:  -s BINARYEN_TRAP_MODE=clamp


jacobfriedman:
BINARYEN_TRAP_MODE=clamp needs to be removed from both web & node builds of EmscriptenTargets.cmake in order for the dockerfile to build the WASM.

JanWielemaker:
Ok.  Removed.  Not tested.   Please close when ok.

jacobfriedman:
Will do with @rla (please pull the latest swipl-wasm and run the docker)

PR #691, opened 2020-10-12
Update EmscriptenTargets.cmake

jacobfriedman:
Please see https://github.com/jacobfriedman/swipl-wasm/tree/master/docker for the latest docker build. It does not work as of now - I am waiting on this clamp mode to be removed.

### DIFF
--- a/cmake/EmscriptenTargets.cmake
+++ b/cmake/EmscriptenTargets.cmake
@@ -2,8 +2,7 @@
 
 set(WASM_NODE_LINK_FLAGS
     -s NODERAWFS=1
-    -s EXIT_RUNTIME=1
-    -s BINARYEN_TRAP_MODE=clamp)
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Can you explain why this clamp mode needs to be removed.   Its been a while ago.  Didn't the clamp mode have something to do with float handling?   Is it just old school?

jacobfriedman:
I'm fairly certain it's old school - at any rate, it doesn't build with that flag. See https://github.com/emscripten-core/emscripten/issues/10365#issuecomment-582930197

"_Oh indeed, you are right. Wasm backend is always -s BINARYEN_TRAP_MODE=clamp (but does not allow one to specify -s BINARYEN_TRAP_MODE=clamp)_"

rla:
Linking to issue #690

PR #739, opened 2020-12-11
DOC: Reviewed Section 5 (extensions) except "Dicts"

dtonhofer:
Reviewed the manual section 5 (SWI-Prolog extensions) BUT NOT the "Dict" part which is still sitting on my HD.

What changed:

----

Moved the sections on how to modify one's code to handle double-quoted string to the back of 5.2, where there are now to two sections concerning "adaptations", where they can be easily found and the headers say what they are about:

5.2.4 Adapting code for double quoted strings
5.2.5 Predicates to support adapting code for double quoted strings 

----

Contrariwise, a chunk of text was moved forward: The section "Representing text: strings, atoms and code lists" sat awkwardly at the end of section 5.2, whereas it is exactly what one needs before the predicates are introduced.

---

Added a little image to illustrate the new model list of SWI-Prolog.

----

Unified usage of markup, as given by the comment:

```
% Note on markup:
%
% After some reflection:
%
% - For the traditional atom [], we use: 
%   Verbatim with quotes inside: \verb$'[]'$ (to make it visibly "quoted")
%
% - For the SWI-Prolog 7 symbol [], we use:
%   \Snil{} without quotes
%   rather than constant without quotes: \const{[]}
%
% For [|] there is \Scons{} but it should be quoted, being an atom 
```

---

To express the fact that text-manipulating predicates accept all manner of types, I have coined the neologism "anytext", which is explained (and then used in the predicate definitions)

---

Big blocks of code with comments has been split into smaller blocks with the comments transformed into separating explainer texts.

---

The header "Syntax Changes", which is not very telling, has been changed to "Syntax changes since SWI-Prolog~7"

---

The PDF looks pretty good, except for page 288, where there is a split code box.

Best regards,

-- David

### DIFF
--- a/man/extensions.doc
+++ b/man/extensions.doc
@@ -2,8 +2,8 @@
 \label{sec:extensions}
 
 This chapter describes extensions to the Prolog language introduced with
-SWI-Prolog version~7. The changes bring more modern syntactical
-conventions to Prolog such as key-value maps, called \jargon{dicts} as
+SWI-Prolog version~7 in 2014. The changes bring more modern syntactical
+conventions to Prolog such as key-value maps, called \jargon{dicts}, as
 primary citizens and a restricted form of \jargon{functional notation}.
 They also extend Prolog basic types with strings, providing a natural
 notation to textual material as opposed to identifiers (atoms) and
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Looks good now.  Merged.  Did fix white space errors (spaces at the end of the line or inside tabs, etc.), so the patch is reported as a conflict now.

PR #746, opened 2020-12-20
Add initial SICStus 4 emulation (split from existing SICStus 3 emulation)

dgelessus:
As discussed [on Discourse](https://swi-prolog.discourse.group/t/expects-dialect-sicstus-how-to-handle-sicstus-3-vs-4-api-conflicts/3381), it's not possible to properly emulate SICStus 3 and 4 as a single dialect, because of some API changes that would lead to name conflicts. This PR adds a separate `sicstus4` dialect that emulates just SICStus 4, and moves the few SICStus 4-only features from the `sicstus` dialect to `sicstus4`.

The SICStus 4 emulation reuses/re-exports many parts of the SICStus 3 emulation. I've made small changes to the term/goal expansions in library/dialect/sicstus.pl so that they take effect in both SICStus 3 and 4 mode. It's not very nice that the `sicstus` (3) dialect now has code in it that's only used by the `sicstus4` dialect, but this way the term/goal expansions don't need to be implemented twice.

The SICStus 4 emulation doesn't have many new features yet - it mostly just re-exports the unchanged predicates from the SICStus 3 emulation. I did add a few simple SICStus 4-only predicates in `library(lists)` and `library(terms)`.

The first few commits in this PR are small fixes and comment updates in the SICStus 3 emulation. Some of them are not directly related to the SICStus 4 emulation and are just small issues that I came across while preparing this PR.

### DIFF
--- a/library/dialect/sicstus.pl
+++ b/library/dialect/sicstus.pl
@@ -60,7 +60,6 @@
 
 	    op(1150, fx, (block)),
 	    op(1150, fx, (mode)),
-	    op(1100, xfy, (do)),
 	    op(900, fy, (spy)),
 	    op(900, fy, (nospy))
 	  ]).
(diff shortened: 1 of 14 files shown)


JanWielemaker:
Thanks.  Merged.   Added an additional patch with some small cleanup in sicstus.pl.    Note that a lot of the stuff you claim missing in the various libraries is there.   Some as built-in predicates, some in other libraries.   Note that you are allowed to export built-ins from modules.   It is pretty meaningless, but does make use_module/2 that explicitly imports some predicate from a specific module work as expected.

I guess we'll see more patches :smile:

PR #753, opened 2020-12-24
sequence//3: commit before separator

pkoch:
From https://swi-prolog.discourse.group/t/confusing-behavior-of-sequence-3-and-sequence-5-from-dcg-high-order/1607

Saw `ctest` pass.

### DIFF
--- a/library/dcg/high_order.pl
+++ b/library/dcg/high_order.pl
@@ -95,8 +95,15 @@ sequence_([], _) -->
 %
 %   See also sequence//5.
 
-sequence(OnElem, OnSep, List) -->
-    sequence_(List, OnElem, OnSep).
+sequence(OnElem, OnSep, L) --> sequence_1(OnElem, OnSep, L), !.
+sequence(_, _, []) --> [].
+
+sequence_1(OnElem, OnSep, [H|T]) -->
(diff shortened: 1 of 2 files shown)


JanWielemaker:

https://swi-prolog.discourse.group/t/confusing-behavior-of-sequence-3-and-sequence-5-from-dcg-high-order/1607/18

JanWielemaker:
I'm not really sure this is going where it should.   This is in part because the main reason why I wanted the high order DCG lib was for serialization.   Even when used for matching though, there are enough cases where the separator is the right place to commit on elements.   Think of the simple case breaking a line on commas:

    sequence(string, ",", List).

A sequence of elements _separated_ by something is different from a sequence of elements _terminated_ by something.   Maybe we need two predicates?  A good alternative, but I fear not so easy to implement, is sequence//3 leaving the final separator on the input if it is not followed by an element.   Might be doable using prolog_cut_to/1 to delay committing.   Would make a bit ugly code, but if it creates an efficient and clearly defined predicate I can live with that.

pkoch:
Could you help me out by writing the test case for that?

pkoch:
This feels stale. Closing until someone shows enthusiasm again to move this forward!

PR #768, opened 2021-01-14
Docs 20210113

dtonhofer:
**Reviewed the text for predicate length/2**

Straightforward. It now looks like this:

Reviewed length two

However, I have added "nondet" to the header so that it now says `[ISO, nondet]` instead of just `[ISO]`. Is this allowed?

**Reviewed the text for =/2**

y adding all the cases where it can fail and which one must collect all over the manual. This is hopefully more accessible (although difficult to explain clearly).

Reviewed Unification

There is one problem with that description which is the original text "Unification acts as if defined by the following fact: =(term, Term)." I just don't understand this. This is either a statement of unification being a reflexive relation or a definition of unification in terms of unification. 

**Moved most of the description of sub_string/5 to sub_atom/5**

In sub_string/5, the text now says "look at sub_atom/5" (but the example specific to sub_string/5 is still there)

sub_atom/5 has been given the text of sub_string/5. The fact that _Before_ increases monotonically on backtracking, formerly unstated, is now mentioned explicitly.

sub_atom
sub_string


JanWielemaker:
Thanks.   Did a partial merge and a few edits.   Notably

  - Please do not add ``% XYZ ---`` comments.  They merely break the uniformity of the docs.
  - Removed the `nondet` for length/2.   Determinism specification is way too complicated for such a simple annotation.   It would only work if we split the modes and have `length(-List, -Length) is nondet`.
  - Modified the sub_string/5 example a bit and make it explicit that these predicates accept either an atom or string as _input_.
  - Discarded the =/2 update.   What unification means is sufficiently described in text books.  Also the details on dif/2, when/2, etc. belong in the section on attributed variables.  A pointer could be useful.   A subsection somewhere on unification seems more appropriate, pointing at occurs check and constraints.   This because unification happens in many places, not just with =/2.

In general I'd like to see more examples being moved to the new example infrastructure.   The new infrastructure allows for more sophisticated display and search of examples.

JanWielemaker:
Handled the current_stream/3 by hand.   Please do not use the ``\Sxxx`` macros directly.   They are inserted automatically from `{xxx}` input.

dtonhofer:
> A subsection somewhere on unification seems more appropriate

Yes. I can't do that thought. Someone with implementation detail knowledge is needed. The unification details were collected empirically...

PR #770, opened 2021-01-14
DOC: 4.1.3 Determinism, 4.5 Type verifications, 4.7 the "\+"

dtonhofer:
More documentation updates (i'm working through my notes here)

In **4.1.3 Predicate behaviour and determinism**

- Extended the list of indicators according to "pldoc" (also, put them into a "table")
- Clarification of vocabulary regarding "leaving no choicepoint" and "well-behavedness"
- Added an example with `member/2`.

Screenshot_20210114_235210

In **4.5 Verify Type of Term**

- Reviewed the whole section

Screenshot_20210114_235420

In **4.7 Control Predicates***

- A reviewed \+

Screenshot_20210115_001605

P.S. 

Is there a way to linkify references to libraries? `\pllib{error}` just generates the text `library(error)` but doesn't transform it into an URL.

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -110,26 +110,66 @@ module declaration, spy/1, and dynamic/1.
 \subsection{Predicate behaviour and determinism} \label{sec:determinism}
 
 \index{predicate behaviour and determinism}%
-To describe the general behaviour of a predicate, the following vocabulary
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Merged after several edits due to conflicts as well as some content refinement.   Sorry for the late action ...

PR #773, opened 2021-01-27
Add SICStus emulation for library(sets), ordsets:ord_member/2, and more library(lists) predicates

dgelessus:
Nothing exciting, just a bunch of simple list predicates and some renames.

### DIFF
--- a/library/dialect/sicstus/lists.pl
+++ b/library/dialect/sicstus/lists.pl
@@ -36,7 +36,9 @@
 	  [ substitute/4,		% +Elem, +List, +NewElem, -List
 	    nth/3,			% ?N, ?List, ?Element
 	    nth/4,			% ?N, ?List, ?Element, ?Rest
+	    same_length/3,		% ?List1, ?List2, ?Length
 	    sublist/2,			% ?Sub, +List
+	    suffix/2,			% ?Suffix, ?List
 
 	    % The following predicates are built-in on SWI.
 	    % We re-export them here to avoid warnings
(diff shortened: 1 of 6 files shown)


JanWielemaker:
Thanks.   I recently added same_length/2; most likely you can remove that again.   For some of the others I'm more inclined to add the missing predicates to the core libraries as well.   We can always do this.

P.s.  The preferred PlDoc lead-in is %! instead of %%.   I started with %% but users convinced be that %! leads to fewer conflicts.

dgelessus:
> I recently added same_length/2; most likely you can remove that again.

same_length/2 is already re-exported from SWI library(lists). Here I added same_length/**3** (with an extra parameter for the length), which SWI doesn't have at the moment.

> For some of the others I'm more inclined to add the missing predicates to the core libraries as well. We can always do this.

Yeah, that may be a good idea for some of the predicates, like same_length/3. Having the full family of \[proper_](prefix/suffix)[_length] predicates might be useful too, but sadly SICStus 4's argument order (`prefix(List, Prefix)`) is different from SWI (and SICStus 3) (`prefix(Prefix, List)`). So even if SWI supported them natively, library(dialect/sicstus4/lists) would still need to emulate them to handle the different argument order.

> P.s. The preferred PlDoc lead-in is %! instead of %%. I started with %% but users convinced be that %! leads to fewer conflicts.

Ah, good to know, I'll use that in my next PRs then. I was following the style used in library(dialect/sicstus), which uses `%%`.

Should I change the comments I already submitted to `%!`, or just leave them as they are?

JanWielemaker:
> but sadly SICStus 4's argument order (`prefix(List, Prefix)`) is different from SWI (and SICStus 3) (`prefix(Prefix, List)`).

:cry: Messing with the argument order of list predicates is asking for hard to debug bugs ...

> to `%!`, or just leave them as they are?

Just leave them as is.   I might at some point change all internal docs to use %! and start actively disabling %%.   A simple `sed` script will do in that case.   Might also never happen.

Just keep up the good work!

Issue #771, opened 2021-01-19
expects_dialect doesn't make emulated builtins visible to subsequently imported modules


dgelessus:
Using SWI-Prolog 8.3.17 built from source (e5ff4a2c4d76ce57d177649b941f38792d876acd). I have code that uses `:- expects_dialect(sicstus).` and then loads other modules, expecting the other modules to be loaded with SICStus emulation as well. This seems to work fine with emulated library predicates, but emulated built-in predicates are visible only to the module that calls `expects_dialect(sicstus)` and not to any modules that it loads afterwards. To reproduce:

inbetween.pl:

```prolog
:- module(inbetween, []).
:- expects_dialect(sicstus).
:- use_module(things).
```

things.pl:

```prolog
:- module(things, []).
:- use_module(library(system)).

% Emulated built-in predicate - doesn't work.
:- trimcore.
% Emulated library predicate - works.
:- environ('TERM', TermType), format("Terminal type: ~w~n", [TermType]).
```

Loading the module:

```prolog
?- use_module(inbetween).
ERROR: .../things.pl:5:
ERROR:    catch/3: Unknown procedure: things:trimcore/0
ERROR:      However, there are definitions for:
ERROR:            inbetween:trimcore/0
ERROR:            sicstus:trimcore/0
Warning: .../things.pl:5:
Warning:    Goal (directive) failed: things:trimcore
Terminal type: xterm-256color
true.
```

For some reason, running `expects_dialect(sicstus).` at the toplevel before loading the module makes the error go away:

```prolog
?- expects_dialect(sicstus).
true.
?- use_module(inbetween).
Terminal type: xterm-256color
true.
```

PR #775, opened 2021-01-29
Add more statistics/2 keys to SICStus emulation, mostly from SICStus 4

dgelessus:
Most of these keys could probably be implemented natively in SWI instead, as "compatibility keys" like the current set of Quintus/SICStus-compatible keys. Some of the keys defined here conflict with their normal SWI meaning though (specifically `garbage_collection` and `atoms`) and would need to stay in the SICStus emulation. Also, SWI's current implementation of statistics/2 seems to assume that all "compatibility keys" return a list value, which is no longer true as of SICStus 4.1, which adds many single integer statistics.

I currently implemented this by defining a custom statistics/2 predicate in library(dialect/sicstus) and library(dialect/sicstus4). This seems to work without issues (except for #771), but based on [this Discourse discussion](https://swi-prolog.discourse.group/t/overriding-shadowing-built-in-predicates-when-is-it-not-allowed/3514/9) it may be safer to implement this using goal expansion and a differently named predicate instead.

### DIFF
--- a/library/dialect/sicstus.pl
+++ b/library/dialect/sicstus.pl
@@ -59,6 +59,8 @@
 	    prolog_flag/3,		% +Flag, -Old, +New
 	    prolog_flag/2,		% +Flag, -Value
 
+	    statistics/2,		% ?Key, ?Value
+
 	    op(1150, fx, (block)),
 	    op(1150, fx, (mode)),
 	    op(900, fy, (spy)),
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.   Not sure whether or not overruling statistics/2 is a good idea.   It may work fine for modular programs, but for none-modular programs you may easily run into conflicts.   I think it would be better to define these as sicstus_statistics/2 and use goal_expansion/2 to map calls to statistics/2 to sicstus_statistics/2 when the sicstus dialect is active.

One scenario that (I think) fails now is if the program uses statistics/2 before expects_dialect(sicstus).

PR #780, opened 2021-02-08
DOC: Correct mode of is/2 from (-,+) to (?,+)

fnogatz:
See ISO-Prolog standard Sec. 8.1.6.2. The point that the first argument should *usually* be a free variable is clarified enough by the documentation, IMO. Related doc page: https://www.swi-prolog.org/pldoc/doc_for?object=(is)/2

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -7642,7 +7642,7 @@ True if expression \arg{Expr1} evaluates to a number non-equal to
 True if expression \arg{Expr1} evaluates to a number equal to \arg{
 Expr2}.
 
-\infixop[ISO]{is}{-Number}{+Expr}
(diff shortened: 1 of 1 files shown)


JanWielemaker:
SWI-Prolog's mode declarations work slightly different.   Anything marked `-` can be instantiated at call time, in which case it is defined to behave the same as e.g. `Tmp is 1+1, Output = Tmp`.   There are only very few predicates that require an uninstantiated argument.  One is open/3 as the stream handle is something that never unifies to any existing term.   Such predicate argument should be marked as `--`.

fnogatz:
So, you prefer to leave it as is, @JanWielemaker? :)

PR #792, opened 2021-02-25
Implement propagation of clp(fd) bitshifts and `div`

### LINKED ISSUE #568: CLP(FD) request: improve bitshifts
I'm writing an assembler in Prolog, and noticed that the bitshift operators (especially right shift), can be improved:

```
?- A #= 4 << 4.
A = 64.

?- 64 #= A << 4.
64#=A<<4.

?- 65 #= A << 4.
65#=A<<4.
```

In the first case, it's behaving as it should be. In the second case, Prolog could determine the exact value of A by bitshifting 64 4 placs to the right. In the third case, Prolog could find that the answer is `false` because the last 4 bits in 65 (0b1000001) aren't zero: the 1 can't appear from thin air.

I would like to improve CLP(FD) with these facts, but don't really know how to get started since the library is huge. I think @triska has written a large part of it, maybe they can help out?

kritixilithos:
This should fix #568. I've also added `div` propagation so that the bitshift propagation code is a bit simpler, and to make it more precise.

What's remaining is to edit the docs to note that the bitshifts are being propagated, and I also want to check my arithmetic again.

### DIFF
--- a/library/clp/clpfd.pl
+++ b/library/clp/clpfd.pl
@@ -1093,6 +1093,17 @@ cis_sign(sup, n(1)).
 cis_sign(inf, n(-1)).
 cis_sign(n(N), n(S)) :- S is sign(N).
 
+cis_slash(sup, Y, Z)  :- ( Y cis_geq n(0) -> Z = sup ; Z = inf ).
+cis_slash(inf, Y, Z)  :- ( Y cis_geq n(0) -> Z = inf ; Z = sup ).
+cis_slash(n(X), Y, Z) :- cis_slash_(Y, X, Z).
+
+cis_slash_(sup, _, n(0)).
+cis_slash_(inf, _, n(0)).
+cis_slash_(n(Y), X, Z) :-
+        (   Y =:= 0 -> (  X >= 0 -> Z = sup ; Z = inf )
+        ;   Z0 is X // Y, Z = n(Z0)
+        ).
+
 cis_div(sup, Y, Z)  :- ( Y cis_geq n(0) -> Z = sup ; Z = inf ).
 cis_div(inf, Y, Z)  :- ( Y cis_geq n(0) -> Z = inf ; Z = sup ).
 cis_div(n(X), Y, Z) :- cis_div_(Y, X, Z).
@@ -1101,16 +1112,8 @@ cis_div_(sup, _, n(0)).
 cis_div_(inf, _, n(0)).
 cis_div_(n(Y), X, Z) :-
         (   Y =:= 0 -> (  X >= 0 -> Z = sup ; Z = inf )
-        ;   Z0 is X // Y, Z = n(Z0)
+        ;   Z0 is X div Y, Z = n(Z0)
         ).
-
-cis_slash(sup, _, sup).
-cis_slash(inf, _, inf).
-cis_slash(n(N), B, S) :- cis_slash_(B, N, S).
-
-cis_slash_(sup, _, n(0)).
-cis_slash_(inf, _, n(0)).
-cis_slash_(n(B), A, n(S)) :- S is A // B.
 
 
 /* - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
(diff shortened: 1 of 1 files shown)


JanWielemaker:
@triska, do you have comments?   A little related, there are no test cases for library(clpfd) in the test.   I'm pretty sure these tests exists.   How can we get these added?

triska:
Thank you for this contribution!

At least https://github.com/SWI-Prolog/swipl-devel/pull/792/commits/971ee503f9fd06268804f84578c1aec4bddf3f35 does not seem correct, though I have not attempted to find a counterexample. Exhaustive enumeration of all domains in a certain range together with suitable constraints where this portion is used (such as multiplication) may yield one.

A few test cases are in the commit messages of `clpfd.pl`, an example is 435209799b58ae01537c82baea6754f8e27a5c70. I hope you find them useful. The best principle for good test cases of CLP(FD) is to enumerate all possible cases, indefinitely. An ideal test case only terminates if a counterexample to a desired property is found, otherwise it keeps running. Mistakes in CLP(FD) propagation often affect only a very small number of cases out of many possibilities, so exhaustive enumeration is very important to detect them.

kritixilithos:
@triska Thanks for the feedback. Running your test without my changes on multiplication gives fewer failed bounds than with my changes (330 vs 4160), so you are indeed right. I wanted a floor_div for some of the constraints, I'll see if I can do without it, or at worst, introduce a `cis_fdiv`. Thank you for the test, it's much better than me manually checking the bounds. I'm not sure how suitable it will be to add this to the Prolog test suite, since even without my changes there still are failed bounds, maybe it can check if the count of failed bounds is less than the previous best, anyways I'll worry about the tests after fixing what I already have.

triska:
I think there are more test cases in the commit history of `clpfd.pl`, please have a look! If you do not find one specifically for multiplication, you may be able to adapt one I added for the other propagators.

The ideal attitude for writing a CLP(FD) test is to regard propagation as a huge search space, in which we are trying to find an example where it propagates too much. Only exhaustive search can help to find such an example: Random specific tests are extremely unlikely to find such examples, because in the vast majority of cases, the propagator is likely to be correct. The test author should be disappointed, not content, if no such example is found, and try to write an even better test that tests more stringently in order to find such an example in the search space.

kritixilithos:
@triska Perhaps we misunderstand each other, when I said 'Prolog test suite' I intended to mean adding CLP(FD) tests to the rest of SWI-Prolog's tests, addressing Wielemaker's comment. I am finding your multiplication test useful (after adapting it for division) for personal testing of my changes, but if your tests are intended to run forever, I don't understand how these can run with the rest of SWI's tests. Maybe I am misunderstanding something though.

Also, you may be interested to know Code Golf Stack Exchange's Prolog chatroom has been unfreezed just yesterday.

(And note to self, start from Jan 2011 (and go earlier) to see Triska's tests)

JanWielemaker:
Note that I'm just waiting until the two of you think it is done and ready to merge.   Its great to see people diving into this library.   It is very well written and anyone can learn a lot from studying it.

triska:
As far as I am concerned, this is done when you @kritixilithos say it is done: My contribution here is only to mention the test cases in the commit history of `clpfd.pl`, and to encourage an attitude for changes in `clpfd.pl` that I recommend: A change to `clpfd.pl` should only be made when the author runs out of energy in the quest to find mistakes in the change. Until then, I recommend to try more and more ways in which such mistakes can be found.

kritixilithos:
Well, I think both div and the bitshifts should no longer propagate wrongly by not including the correct answers within the new ranges (as my commits previous to Triska introducing me to his tests did), at least as per tests with the variables ranging from -4 to 4. So in the sense of "find an example where it propagates too much", I think this should be ready.

But the propagation of constraints could be made more restrictive, particularly, I left some TODOs in pshift propagation (`pshift` is what I call the propagator for the bitshifts) so its code is more empty than others (but then again, some of the other operations don't propagate domains for certain conditions). However, before I address those TODOs in pshift, I want to work on generalising `domain_shift`, `domain_contract`, `domain_contract_less`, and others to make sign handling easier so that, if it works, many of these propagation TODOs across other propagators also get addressed. But I plan to do these later in another PR (my holiday's ending), along with improving exponentiation's propagation since the bitshifts rely on it, if that's alright with you @triska.

It doesn't seem the docs need editing, nothing there seems to imply bit operations weren't propagated in the first place, so I'll leave it be, unless anyone thinks otherwise.

triska:
Yes, I would like to add this about the docs: It is good to state as few guarantees as possible about any strength of propagation, because application programmers should not come to rely on it: The arising use cases and applications shall dictate the propagation and efficiency trade-offs chosen in the library, not the other way around.

kritixilithos:
Seeing Triska agrees, I'd say this is ready to merge now @JanWielemaker .

> It is very well written and anyone can learn a lot from studying it.

I've only looked at a small portion of CLP(FD) so far, I hope to get to study the rest of it soon.

JanWielemaker:
I'm happy to merge.   Looking through the patches it seems later patches are in part reversing older patches.  I could of course squash the patches, but this may also be a bit too rough.   Would you mind cleaning up the history yourself (i.e.,  using `git rebase -i`)?   I also think the commit tags are not really adequate.   MODIFIED means that the interface or behavior have changed in some incompatible way.   Just extending the domain would be ADDED.  ENHANCED is for stuff that merely improves performance without changing the contract with the user and FIXED fixes real bugs (wrong answers, crashes, etc.).

kritixilithos:
Thanks for the advice of using `git rebase -i`, just performed my first rebase :D (Note to self: after rebasing, I pushed with `git push --force-with-lease`). I reordered the commits grouping them into `div` propagation and bitshift propagation commits, and squashed them accordingly. Hope the commit tags are correct now.

JanWielemaker:
Thanks.  Merged.  Congrats with first rebase and contrib to clpfd!   Faking a clean history is very useful for understanding why the code is as it is today.

PR #796, opened 2021-03-01
REVIEW: Doc reviewed & "setter predicates" can be used to "get"

dtonhofer:
The documentation for library portray_text seems a bit oldish. Reviewed, updated.

Also made the "setter predicates" work as "getter predicates". If the term `pull(X)` is passed as value, `X` is unified with the current setting, Thus we can "get" and "verify" while still working completely distinctly from the case where `X` is nonvar and the value to be "set".

### DIFF
--- a/library/portray_text.pl
+++ b/library/portray_text.pl
@@ -46,24 +46,62 @@
 
 /** <module> Portray text
 
-A Prolog string is a list of character-codes: (small) integers, which
-results in output like this:
+SWI-Prolog has the special string data type. However, in Prolog, text
+may be represented more traditionally as a list of character-codes, i.e.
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Merged after some more extensive edits.   I notably do not like the pull(Old) thing as that makes the signature unclear and doesn't play well with future static analysis, so I added set_portray_text(+Key, -Old, +New).   Also not what I'd choose now, but it has the smallest impact.

Please use `true`, etc rather than the old legacy =true=.   I want to move as much as possible to "modern" Markdown.   Note that `...` implies ... is valid Prolog and use ``...`` for general code style regardless of the syntax.

PR #803, opened 2021-03-21
A fix for `ansi_format` position miscalculations due to prepended escape sequences

phaylon:
The problem this fixes: Positions in user-supplied formats are off by the length of the prepended escape sequences used for coloring the output.

Example:
```
?- format('~t~5|~w', [foo]).
     foo
true.

?- ansi_format([], '~t~5|~w', [foo]).
  foo
true.
```

This changes the order of operations in `ansi_format` for constructing the final output.

* Before: The user-supplied format was wrapped by escape sequences. Then the color code and the original arguments were applied to the new format.
* Now: The original output is constructed into a string. This string is then wrapped by the escape sequences during output.

Note that this is my first time contributing Prolog code, and the first time I'm touching production Prolog. So I wouldn't be surprised if I missed something. For example: I'm not sure if preconstructing the output before wrapping would be considered inefficient. I wouldn't expect styled output to be used with large payloads, but I'd thought I'd mention it.

Another effect this change has is that `ansi_format` is now compatible with `format` with regard to single non-list arguments:

```
?- ansi_format([], '~t~5|~w', foo).
     foo
true.
```

I'd be happy to also supply some tests that ensure that `ansi_format([], Format, Args)` and `format(Format, Args)` behave the same. But since there isn't yet a `src/Tests/library/test_ansi_term.pl` I was wondering if there is a test-file template I could use, specifically with regard to how the copyright notice should look for an outside contributor.

### DIFF
--- a/library/ansi_term.pl
+++ b/library/ansi_term.pl
@@ -137,8 +137,13 @@ ansi_format(Stream, Class, Format, Args) :-
     class_attrs(Class, Attr),
     phrase(sgr_codes_ex(Attr), Codes),
     atomic_list_concat(Codes, ;, Code),
-    format(string(Fmt), '\e[~~wm~w\e[0m', [Format]),
-    format(Stream, Fmt, [Code|Args]),
+    with_output_to(
+        Stream,
+        (   keep_line_pos(current_output, format('\e[~wm', [Code])),
+            format(Format, Args),
+            keep_line_pos(current_output, format('\e[0m'))
+        )
+    ),
     flush_output.
 ansi_format(Stream, _Attr, Format, Args) :-
     format(Stream, Format, Args).
--- a/src/Tests/library/test_ansi_term.pl
+++ b/src/Tests/library/test_ansi_term.pl
@@ -1,0 +1,97 @@
+/*  Part of SWI-Prolog
+
+    Author:        Robert Sedlacek
+    E-mail:        rs@474.at
+    WWW:           http://www.swi-prolog.org
+    Copyright (c)  2021, SWI-Prolog Solutions b.v.
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions
+    are met:
+
+    1. Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in
+       the documentation and/or other materials provided with the
+       distribution.
+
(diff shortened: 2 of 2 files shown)


JanWielemaker:
Thanks.   Yes, this is a problem.   The fix is not this easy though :cry:  The problem you now introduce (well, the current state also suffers somewhat) is that by formatting to a string you ignore the current (line) position of the output stream (that is a new problem) and the output still leads to incorrect line position management as the escape sequences are simply counted as "moves right".   There are several ways out.

  - Add something to format/3 that indicates the low level system that this is an escape sequence and it should save/restore the stream position around it.    So you get e.g. ``~@\e...~#``.  Some care is needed to find a free escape sequence that is not likely to become popular for anything else.   It doesn't need to be user friendly though, so you could use something non-ASCII.
  -  Use three writes, one for the leading escape, one of the real thing and one for the closing escape.  Wrap the escape writing to save/restore the stream positions (can be done in Prolog) and use with_output_to/2 to keep the stream management atomic.
  - Extend the position logic in streams to understand the ANSI sequences and not update the stream position.

They all come with pros and cons.   If you think you can handle any of this, please go ahead.  

> But since there isn't yet a `src/Tests/library/test_ansi_term.pl` I was wondering if there is a test-file template I could use, specifically with regard to how the copyright notice should look for an outside contributor.

Just copy one of the other files, empty it and change the author.   At least, that is what I do ...

phaylon:
@JanWielemaker 

Thanks, for all the hints. Yes, the general stream state is exactly the kind of thing I meant when I said I'm likely missing something, thank you :D

Of the three options, I'd go with the second one. No other parts of the formatting machinery need to know about things outside their domain or watch out for new magic markers.

The stream position handling would simply be `stream_property(Stream, position(Position))` and `set_stream_position(Stream, Position)` from what I can tell, so the second solution should also read really nice. So I'm gonna go ahead and try that one first.

> Just copy one of the other files, empty it and change the author. At least, that is what I do ...

Excellent, thank you!

phaylon:
I just did a force push to my branch with the new implementation. Some comments and questions:

* I found that there already is a `keep_line_pos`, From looking at it I assumed this does what we want it to do and used that.
* Is my double use of `keep_line_pos(current_output, ...)` idiomatic enough or should that be captured?
* I'm not sure how happy I am with the `captured_tty_output´ helper in the tests. I hope it's correct enough for simulating a TTY in the context of a test.
* Any further tests that are worth adding?

JanWielemaker:
Congrats with your first PR!   I only changed the commit message to start with `FIXED:` to make it appear in the changelogs.  I also rebased to patch.  The two makes GitHub think it has not been applied.  It has :smile: , so I close this.

phaylon:
Thank you! :D

PR #817, opened 2021-04-19
Early/better debug output

dmchurch:
Initialize the global debug_level as soon as possible if "-d" is passed
first on the command line, to avoid missing early debug lines. Also,
track use of the DEBUG() macro on a per-thread basis and use Serror
as the default output stream when inside a DEBUG clause, so that
statements that mix e.g. Sdprintf() and write_term() will get printed
to a single output. Finally, prefix debug lines with the debug channel
and the file/line source.

### DIFF
--- a/src/os/pl-stream.c
+++ b/src/os/pl-stream.c
@@ -2113,15 +2113,9 @@ next_chr(const char **s, IOENC enc)
   }
 }
 
-#ifdef O_DEBUG
-#define OUTCHR(s, c)	do { printed++; last_char = (c); \
-			     if ( Sputcode(last_char, (s)) < 0 ) goto error; \
-			   } while(0)
-#else
 #define OUTCHR(s, c)	do { printed++; \
 			     if ( Sputcode((c), (s)) < 0 ) goto error; \
 			   } while(0)
-#endif
 #define valdigit(c)	((c) - '0')
 #define A_LEFT	0			/* left-aligned field */
 #define A_RIGHT 1			/* right-aligned field */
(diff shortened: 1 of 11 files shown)


dmchurch:
I've been working on enabling full-fledged WASM/Emscripten support in swipl, and getting that early debug output has been crucial! I figure it's pretty helpful across the board, as well.

dmchurch:
Adding macro definitions for the non-debug build. Every existing callsite is already guarded by the `O_DEBUG` flag, but no harm in allowing them to be used outside it as well.

src/os/pl-stream.c
```diff
@@ -2101,9 +2113,15 @@ next_chr(const char **s, IOENC enc)
   }
 }
 
+#ifdef O_DEBUG
+#define OUTCHR(s, c)	do { printed++; last_char = (c); \
```
JanWielemaker, line 2117: I think you missed that streams already have an `s->lastc` field.   That should simplify things a little, no?

JanWielemaker:
Thanks.  Looks good.   I've left a comment on the use of `last_char`, proposing to use the stream's `lastc` field.   If that it resolved (either way) I'll merge.

src/os/pl-stream.c
dmchurch: Aha, I absolutely did miss that! Fixing, please stand by...

dmchurch:
Using `lastc` now! Also, fixed a segfault that could happen on bad input.

JanWielemaker:
Will merge soon.  Bit too deep in other code right now.  Thanks!

JanWielemaker:
Thanks!  Merged after rebase and merging the two commits.  This makes GitHub think it is conflicting :cry:

PR #833, opened 2021-05-13
Add max_member/3 and min_member/3 (with an order predicate argument) to library(lists)

dgelessus:
These predicates are compatible with SICStus 4.


JanWielemaker:
Merged after adding the meta_predicate/1 declaration and an example.    I would probably have added it to library(apply), but I guess compatibility with SICStus is a good reason :smile:

dgelessus:
> Merged after adding the meta_predicate/1 declaration

Oops. Thanks for catching that!

PR #837, opened 2021-05-14
Debug wildcard and complement; DEBUG_IF

dmchurch:
The -d command line option now takes an optional asterisk (*) on the
end of a debug flag name to enable all flags that begin with that name.
Also, flag names can be prefixed with a caret (^) to disable that flag;
only particularly useful in combination with the wildcard, to silence
chatty channels. For example,

    swipl -d MSG_SHIFT*,^MSG_SHIFT_FRAME

will enable all MSG_SHIFT channels except for MSG_SHIFT_FRAME.

On the internal side, we now have the DEBUG_IF(cond, channel) macro,
which functions much like DEBUG(channel, statements) except it provides
more control over the conditions, and it avoids statements in a macro.
As a result, the statement

```c
DEBUG(MSG_CHANNEL, Sdprintf(...));
```

is (for a named channel, at least) equivalent to

```c
 int debug_active = DEBUGGING(MSG_CHANNEL);
 DEBUG_IF(debug_active, MSG_CHANNEL)
 { Sdprintf(...); }
```

which itself is roughly equivalent to:

```c
#if O_DEBUG
int debug_active = DEBUGGING(MSG_CHANNEL);
if (debug_active)
{ ENTER_DEBUG(MSG_CHANNEL);
  Sdprintf(...);
  EXIT_DEBUG(MSG_CHANNEL);
}
#endif
```

This commit also includes some improvements to the GC debug messaging.

### DIFF
--- a/src/os/pl-stream.c
+++ b/src/os/pl-stream.c
@@ -110,18 +110,6 @@ locking is required.
 #include SYSLIB_H
 #endif
 
-#ifdef O_DEBUG
-#ifdef HAVE___THREAD
-/* It's not strictly necessary to have this in thread-local storage... */
-__thread
-#elif defined(O_PLMT)
-/* ...and the fail condition for a thread race just isn't that bad */
-volatile
-#endif
-/* Have we just emitted a newline? Should we print file/line on the next Sdprintf? */
-	 static int debug_new_output_line = 1;
-#endif
-
 #define ROUND(p, n) ((((p) + (n) - 1) & ~((n) - 1)))
 #define UNDO_SIZE ROUND(PL_MB_LEN_MAX, sizeof(wchar_t))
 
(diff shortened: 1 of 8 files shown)


JanWielemaker:
Thanks.   Using wildcards for the debug messages has always been on my mind.  It was never urgent enough though.    Unfortunately your patch breaks compilation in normal release mode.

    ../src/pl-inline.h:406:7: warning: unused variable ‘__new_ld_depth’ [-Wunused-variable]

(many times).    Second problem I have is that the messages in `pl-gc.c` are deliberately there even if the system is not compiled for debugging.   There messages are often enough a life saver is something happens that makes knowing about GC and shift actions useful.   I propose to revert `pl-gc.c` changes of this patch.   Also wondering why `internal_debug` is unconditionally part of the release version?

dmchurch:
Oops! Totally right on all counts, not sure how I traced that code wrong! Let me pull some of these changes (and actually test the release build) before I update this.

dmchurch:
There, that ought to do it. Look good to you, @JanWielemaker?

JanWielemaker:
Thanks.   Looks good now.  I've squashed the commits and added the [A-Z]: tag (DEBUG) to make it part of the changelog.   In pl-gc.c I've added one more change to `%zd`.   Reverted the SHIFT `l` to `l?lsize:0`, etc.   The printed value is the request, which is one of 0 (no need to change), 1 (change to next size) or some concrete value (change to next size that provides the request amount of free space).

PR #838, opened 2021-05-18
BUILD: Added self-contained PGO process to CMake.

dmchurch:
This change makes it possible to run a PGO build with a single
invocation of the build process. To do so, use the "PGO" configuration
in CMake rather than "Debug" or "Release", which will:

1) Compile all libswipl objects with profile generation enabled. This
   uses auxiliary target names with "-pgo-instrumented" appended.
2) Link the executable "swipl-pgo-instrumented" in src, using these.
3) Use swipl-pgo-instrumented to create boot.prc (since no other swipl
   exists yet). Clear the PGO-data directory, then run the benchmark.
4) Build the libswipl objects again, using the PGO data generated.
5) Link the final swipl executable from these.

The dependencies are set up such that if any source file changes,
ninja will recompile the instrumented version of the affected file(s)
and then restart the process from step 3, recompiling ALL libswipl
objects using the new PGO data.

Tested with gcc 9/10 and clang 10/11.


JanWielemaker:
This is great.   I tried to achieve this when moving to CMake, but failed.  The pgo-compile.sh script was the work-around.   Works fine on Linux and MacOS.  Cross-compiling for Windows however results in the error below.   Before I dive into this I thought to ask you first as you've probably some experience :smile: 

```
...
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "pgo_data" of type UTILITY
    depends on "bootfile" (strong)
  "bootfile" of type UTILITY
    depends on "swipl" (strong)
  "libswipl" of type SHARED_LIBRARY
    depends on "pgo_data" (strong)
    depends on "swiplobjs" (strong)
  "swiplobjs" of type OBJECT_LIBRARY
    depends on "pgo_data" (strong)
  "swipl" of type EXECUTABLE
    depends on "libswipl" (weak)
    depends on "pgo_data" (strong)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
```
The build is done using the Docker image from the https://github.com/SWI-Prolog/docker-swipl-build-mingw repo.   This runs:

```
    cmake -DCMAKE_BUILD_TYPE=PGO \
          -DSKIP_SSL_TESTS=ON \
          -DJAVA_HOME="$WINEPREFIX/drive_c/Program Files/Java/jdk-13.0.2" \
          -DCMAKE_TOOLCHAIN_FILE=../cmake/cross/linux_win64.cmake \
          -DJAVA_COMPATIBILITY=ON \
          -DJUNIT_JAR=/usr/share/java/junit.jar \
          -G Ninja ..
    ninja
    cpack
```

CMake version is most likely irrelevant: On Ubuntu I use 3.16, on MacOS 3.19 from Macports (both ok) and the Windows cross-build runs Fedora 33 with CMake 3.18.

dmchurch:
Ah! I didn't think about the cross-compiling case. The main fix will be in Ports.cmake, which needs a fix to `PROG_SWIPL_FOR_BOOT` for the cross-compiling emulator mode - getting that to reference "swipl-pgo-instrumented" instead of "swipl" should do the trick. There's probably going to be another issue in `run_pgo_program()`, though, which expects that a single name can refer to both a dependency target and an executable. It's probably enough to add a CROSS_COMPILING_EMULATOR check into run_pgo_program, though I couldn't say for sure without testing.

Does the old PGO script still work in that case? I tried to leave that code path relatively untouched, for people who want to do manual builds or weird edge cases like running PGO on a debug build.

Anyway if you've got a fix in mind for this, go right ahead! I think you can even commit to my `auto_pgo` branch, since you're a swipl maintainer. Or I can just fix it later - I'll need to whenever I get back to the Emscripten build, after all!

JanWielemaker:
> Does the old PGO script still work in that case?

A quick test says no.   That was what I was expecting and I already updated the script to be a dummy script calling the new stuff such that old scripts keep working (reverted that before trying).

> Anyway if you've got a fix in mind for this

I'm afraid not.   I'm sure that if I really want I can figure it out.  I also know this stuff can take some time :cry: Seems you more or less know what to do though.  Can you propose a patch that I can try?

dmchurch:
> Can you propose a patch that I can try?

Here, you can give this one a shot! Anything further will have to wait until I'm actually at my workstation, though.

```patch
diff --git a/cmake/PGO.cmake b/cmake/PGO.cmake
index d479c6150..0026ab293 100644
--- a/cmake/PGO.cmake
+++ b/cmake/PGO.cmake
@@ -139,9 +139,14 @@ function(add_pgo_dependency) # add_pgo_dependency(targets...)
 endfunction()
 
 function(run_pgo_program exec_target) # run_pgo_program(exec_target [args...])
+  if(CMAKE_CROSSCOMPILING AND CMAKE_CROSSCOMPILING_EMULATOR)
+    set(executable ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:${exec_target}${PGO_SUFFIX}>)
+  else()
+    set(executable "${exec_target}${PGO_SUFFIX}")
+  endif()
   add_custom_command(OUTPUT ${PGO_RUN_STAMP} APPEND
         DEPENDS "${exec_target}${PGO_SUFFIX}"
-        COMMAND "${exec_target}${PGO_SUFFIX}" ${ARGN}
+        COMMAND ${executable} ${ARGN}
         VERBATIM)
   # if we rebuild the binary, we have to clear the pgo data
   add_custom_command(TARGET "${exec_target}${PGO_SUFFIX}"
diff --git a/cmake/Ports.cmake b/cmake/Ports.cmake
index b332e558c..6e39fb3c7 100644
--- a/cmake/Ports.cmake
+++ b/cmake/Ports.cmake
@@ -77,6 +77,9 @@ if(CMAKE_CROSSCOMPILING)
        CACHE STRING "SWI-Prolog executable to perform build tasks")
     set(PROG_SWIPL ${CMAKE_CROSSCOMPILING_EMULATOR} ${PROG_SWIPL_TARGET_FILE})
     set(PROG_SWIPL_FOR_BOOT ${PROG_SWIPL})
+    if(CMAKE_BUILD_TYPE STREQUAL "PGO")
+      set(PROG_SWIPL_FOR_BOOT ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:swipl-pgo-instrumented>)
+    endif()
   endif()
 elseif(CMAKE_BUILD_TYPE STREQUAL "PGO")
   set(CMAKE_HOST_EXECUTABLE_SUFFIX ${CMAKE_EXECUTABLE_SUFFIX})
```

JanWielemaker:
Configure now works fine.  Building first claims `src/wine` cannot be found and is a dependency for `boot.prc`.  For now hacked around using `touch src/wine`.    Then we get this though.  Looks like missing quotes?  But why not for Linux?

```
FAILED: src/CMakeFiles/swiplobjs-pgo-instrumented.dir/pl-term.c.obj 
/usr/bin/cmake -P /home/swipl/src/swipl-devel/cmake/PGO.cmake wrap-gcc pgo-instrumented /usr/bin/x86_64-w64-mingw32-gcc  -Isrc -I../src -O2 -D__WINDOWS__ -D_WIN32_WINNT=0x0600 -DWIN64 -Wall -fno-strict-aliasing -fprofile-generate -Wno-maybe-uninitialized -fprofile-dir=/home/swipl/src/swipl-devel/build.win64-pgo/PGO-data -MD -MT src/CMakeFiles/swiplobjs-pgo-instrumented.dir/pl-term.c.obj -MF src/CMakeFiles/swiplobjs-pgo-instrumented.dir/pl-term.c.obj.d -o src/CMakeFiles/swiplobjs-pgo-instrumented.dir/pl-term.c.obj -c ../src/pl-term.c
CMake Error: Parse error in command line argument: -D__WINDOWS__
Should be: VAR:type=value
```

dmchurch:
This might work now! Works for me, at least, in both `CMAKE_BUILD_TYPE=PGO` and `../scripts/pgo-compile.sh` modes.

JanWielemaker:
Thanks.   Merged after some more debugging.   There were a couple of minor problems with your wrap-gcc.   For the Windows build there was an unwanted dependency on boot.prc for `wine` because the executable was set to "wine ..." and the exectable is a dependency.  That is the easy part.   I have a little more doubt about changing your  `set(PROG_WRAPGCC wrap-gcc)` in src/CMakeLists.txt to an absolute path (avoiding the need to have . in PATH, I guess).  First I thought this should not have been used anyway when not cross-compiling, but I think it must and the change is correct.  Please correct when I misunderstood.

Then you added a lot of code to find the need for using wine and which wine to the toolchain files.  In turns out that CMake processes the toolchain file for __every__ `try_compile()`.   This made the build terrible slow.   I see no real reason not to use the CMAKE_CROSSCOMPILING_EMULATOR even if we do not have to.   You may want some of the flexibility back.  It seems you should move any test that costs time outside the toolchain file.

Anyway.  I like it.   It simplifies the build process and allows generating a PGO version using make.  May eventually remove scripts/pgo-compile.sh.  Thanks.

dmchurch:
I'm glad you like it! However, the changes you made have broken the build on my local system. I have a change ready that will keep the improved build speed, but not break it on my side.

The reason you were seeing a dependency on `wine` for boot.prc was _because_ wine was set to a bare name and not an absolute path. That's why I altered the toolchain file to use the full path to wine for `CMAKE_CROSSCOMPILING_EMULATOR` - at least when wine support is even needed. If you'd left the toolchain files alone, you wouldn't have needed to add the DEP_SWIPL_FOR_BOOT variables. Also, the `set(PROG_WRAPGCC wrap-gcc)` without path or extension is intentional: if you use the name of an executable target as a command, CMake replaces it with  the full path to the target file while automatically adding a dependency on its build.

Anyway, I didn't realize CMake was including the toolchain file multiple times. Apparently it swallows output when included for other reasons besides the initial load? That's a good catch. The solution to that, however, is to ensure that the cache has been loaded before running the expensive tests, not to omit the tests entirely. The version that you checked in under my name removes the support I added for people running the build in native mingw or WSL (like I do), who won't have wine installed (and won't want to use it, anyway).

Should I open a new PR to re-add the WSL/mingw support, or just push another commit to this one? Also, what commit guidelines should I follow for direct merging into the main repo? It complicates things for me when my commits aren't merged into the repo history, because then I have to do rebase gymnastics to fix whatever local branches I had that were dependent on that. I'd much rather not have to deal with that—not to mention, I figure I can lessen your workload if my commits are already in the right format for a direct merge.

JanWielemaker:
Ok.  Sorry for messing up.  Seems we need to arrange this differently.  Your initial versions worked on only a few of my platforms and configuration combinations.  After several hours going back and forth between machines and configurations it now works on all (mine).  Any even a little expensive test in the toolchain file seems unacceptable.  It was not just a little slower, but more like 1000 times :cry:   Using `set(PROG_WRAPGCC wrap-gcc)` leads to "command not found" when building the PGO on Linux.   Seems WSL is the exception (guess it has . in PATH, something we really do not want) ...    When cross-compiling from WSL you indeed probably do not want to use wine.

Anyway.   Any new change should start from the current `master`.    I __think__ we still need my changes to `set(PROG_WRAPGCC wrap-gcc)` and the DEP_SWIPL_FOR_BOOT.    WSL might have some use for cross-compiling a native Windows version, although the Docker version might be a more sensible alternative as it is reproducible regardless of the platform.   That means we need your detection in the toolchain files.  They must be cached though and as I understand, caching stuff in the toolchain files is not easy.   Maybe these tests can be outside the toolchain file?

Next is how to deal with this type of complicated PRs.   I could get the PR, create my own patches on top of it and push it to a branch?

dmchurch:
Like I said, I already have a solution ready! It's actually not too hard to have the tests only run at a single inclusion of the toolchain file and ensure they're cached, now that I know it's necessary. I'll finish it up and push it once I'm done with the refactor stuff I'm working on at the moment.

I'm really surprised to hear that it's not working for you, though! Not to mention the huge slowdown you've been reporting, which I still haven't seen; in my pre-push tests (without my upcoming toolchain cache fix) the PGO build was maybe a _tiny_ bit slower but nothing like what you're reporting. My primary compilation target has been and continues to be Linux-native, and It Works Fine On My Machine™—and no, I don't have `.` in my `PATH` (though I almost had a heart attack thinking that somehow it might have snuck in there) 😂 I blame CMake for not adequately insulating against platform variances, I can't imagine what else it'd be.

Anyway, I'm fairly certain that, as the maintainer of the swipl-devel repo, you can push to any branch I've opened a PR from! If that's not the case, let me know and I can just give you write access to my repo as a whole. It _should_ (dangerous word, there) be as simple as adding my repo as a remote and doing a fetch. Also, according to the [GH docs](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally) apparently you can check out the source of _any_ PR with a command line like:

```sh
REMOTE=origin
PR_ID=838
LOCAL_BRANCH=pgo-pull-request
git fetch $REMOTE pull/$PR_ID/head:$LOCAL_BRANCH && git checkout $LOCAL_BRANCH
```
Alternately, you could add to your origin fetch spec:

```sh
git config --add remote.origin.fetch '+refs/pull/*/head:refs/remotes/origin/pr/*'
```

and at that point you could just `git fetch && git checkout pr/838` to get the latest. Wouldn't help with pushing, though—you can't push to the `pull/XYZ/head` ref names.

In any case, though, I've been trying to keep my branch names distinct from the ones in swipl-devel, so if you set up a remote for my fork, it should be as simple as

```sh
# Assuming you've added my repo as a remote called 'dmchurch':
git fetch dmchurch && git checkout auto-pgo
# Make changes, commits, etc, on top of auto-pgo...
git push
```

dmchurch:
> I would be surprised.

Well, it's easy enough to test! I'll open a dummy PR (where I've clicked the "allow edits by maintainers" button) and you can try and push to that branch.

PR #842, opened 2021-05-31
Fix silent failure in CLP(FD) domain calculation for truncated division

dgelessus:
Previously, expressions like this silently failed for no reason:

```prolog
?- A #= B div 3.
false.
```

I'm not a CLP(FD) expert, but this seems to be a simple typo in the code. With that fixed, it works as expected:

```prolog
?- A #= B div 3.
B div 3#=A.

?- A #= B div 3, A #= 5.
A = 5,
B in 15..17.

?- A #= B div 3, B #= 5.
A = 1,
B = 5.
```


JanWielemaker:
Thanks.  Seems to fix f843291352

PR #843, opened 2021-06-01
Shared-inline functions, pl-wam cleanup

dmchurch:
This changes how functions in pl-inline.h are declared and compiled without affecting the inline semantics, and also separates the compilation of pl-wam.c from the compilation of all the other .c files (except for pl-vmi.c).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -73,7 +73,7 @@ set(SRC_CORE pl-atom.c pl-wam.c pl-arith.c pl-bag.c pl-error.c
     pl-copyterm.c pl-debug.c pl-cont.c pl-ressymbol.c pl-dict.c
     pl-trie.c pl-indirect.c pl-tabling.c pl-rsort.c pl-mutex.c
     pl-allocpool.c pl-wrap.c pl-event.c pl-transaction.c
-    pl-undo.c)
+    pl-undo.c pl-inline.c)
 
 set(LIBSWIPL_SRC
     ${SRC_CORE}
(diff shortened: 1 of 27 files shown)


JanWielemaker:
Cleanup is always welcome :smile:    On the benchmarks from `bench` the difference is not reliably measurable.    On the CHAT80 program though I get a good reproducible 4% performance drop.  Both using the PGO compiled version.    To get this program:

```
cd ~/.config/swi-prolog
git clone https://github.com/JanWielemaker/my-prolog-lib lib
swipl -g 'make_library_index(lib)' -t halt
```
After this, I run

```
src/swipl -O
?- time(rtest_chats(100)).
% 7,085,501 inferences, 0.602 CPU in 0.602 seconds (100% CPU, 11772084 Lips)
```
I recall similar changes when merging these files to a single compilation unit.  I assuming this is due to more inlining, specialization and possibly using dedicated calling conventions when all is in one compilation unit.   I wonder what LTO can do to "fix" this.   I was always under the impression its role was to improve locality, merge identical objects and remove unused objects.   Does it do more these days?

Anyway, can you reproduce the above numbers and if so how can we improve on this?

dmchurch:
I'm having a lot of trouble reproducing this, actually! On the dev box I'm unable to get consistent times (presumably, because it's a virtualized host and thus affected by host system load) and on my local workstation I'm only seeing an 0.5% difference in CHAT80 performance between the `pl-inline` branch and its branch point—and that's actually even _less_ of a difference than I get simply by doing a full vanilla PGO build twice and comparing the outputs. What architecture are you seeing the performance difference in?

Oh, and incidentally: LTO does a _lot_ more than that in recent versions. When you pass `-flto` to GCC, it actually only half-compiles each file into an intermediate representation which gets stored in the `.o` file, and it performs the final compilation/optimization pass at link time, once all the function definitions are available.

PR #844, opened 2021-06-02
BUILD: Moved pl-alloc/index/fli out of pl-wam

dmchurch:
**[NOTE: this is an alternative to PR #843. If the performance issues in that one can't be resolved but this one works, this is a perfectly acceptable alternative.]**

This adds USE_*_INLINES as a flag that can be set before including
pl-incl.h, to include some of the function definitions from various
files in a "can-inline" context. It also separates the majority of
pl-alloc.c, pl-index.c, and pl-fli.c out into their own object files,
to speed up the build.

This may result in a slight (1-2% at max, in my testing) performance
loss since the pl-wam compilation doesn't have access to *all* the
function definitions in those three files, but enabling LTO on the
build (which I have not yet, since it currently generates warnings)
avoids the performance drop entirely by allowing the linker itself
to perform any applicable cross-file inlining.


JanWielemaker:
Performance difference is now within noise levels :smile:

PR #849, opened 2021-06-23
Add library(file_systems) to SICStus 4 emulation

dgelessus:
This should be pretty complete. The only missing predicate is `close_all_streams/0` - I have no idea what practical uses that has...

Many file/directory properties aren't implemented, because SWI doesn't have any equivalent APIs. But all of the important ones are supported (size, modified timestamp, access permission).

### DIFF
--- a/library/dialect/sicstus4/file_systems.pl
+++ b/library/dialect/sicstus4/file_systems.pl
@@ -1,0 +1,393 @@
+/*  Part of SWI-Prolog
+
+    WWW:           http://www.swi-prolog.org
+    Copyright (c)  2021, SWI-Prolog Solutions b.v.
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions
+    are met:
+
+    1. Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.    I'm not against adding a mode `search` as alias for `execute` to e.g., absolute_file_name/3 and access_file/2.   Just file a PR :smile:    Implementing `close_all_streams` is not hard, but whether you want it ...  Notably in threaded applications you easily deadlock and even if you didn't, it is not so nice to close someone else's streams ...

dgelessus:
Alright, I've sent a PR for natively supporting `search` access checks: #850

I really wouldn't worry too much about `close_all_streams/0`. The SICStus documentation says

> The time to call this is after an `abort/0`.

Which makes it sound like it's more of a workaround for bad resource management, which should rather be fixed by using `(setup_)call_cleanup` properly...

Our code doesn't need it (thankfully), so I won't worry much about it.

JanWielemaker:
:)  SWI-Prolog used to close all streams on abort.   This behavior was inherited from the systems I knew back then (mostly C-Prolog and Quintus). Single threaded that makes some sense, but multi-threaded this is not a good idea.  There is now a stream property `close_on_abort`.   I doubt anyone uses it ...

PR #850, opened 2021-06-23
Add file access mode `search` as an alias for `execute`

dgelessus:
As suggested in https://github.com/SWI-Prolog/swipl-devel/pull/849#issuecomment-866882326.

This makes `search` a simple alias for `execute`, as SWI's `execute` check on directories already behaves like `search`. This *mostly* matches the SICStus behavior, except in the corner case of applying a check to the wrong file type (i. e. `execute` on a directory or `search` on a file). The exact behavior in that case probably doesn't matter in practice, but just in case I also explained it in the access_file/2 docs.


JanWielemaker:
Thanks.  Merged.

PR #851, opened 2021-06-23
Extend and simplify library(file_systems) based on library(filesex)

dgelessus:
I completely forgot about library(filesex). This allows implementing a few more parts of library(file_systems) (all `delete_directory/2` options, as well as creation/access time properties) and simplifies the code in some places (member of directory predicates).


JanWielemaker:
Thanks.   Merged.

PR #856, opened 2021-07-05
Reimplement SICStus block emulation using attvars

dgelessus:
Updated version of #812 - see my explanation there for the motivation behind this.

Unlike my previous attempt, this implementation is based completely on manual attvars and no longer uses `freeze`. I've also changed the internal attribute structure to be similar to the one used by `when`, so the performance should be comparable to `when`.

I haven't done too much detailed performance testing yet. I attempted some small optimizations related to the attribute representation and avoiding the `unblock/2` predicate, but they didn't give any noticeable performance improvement for our code. I'm still running into other (apparently unrelated) slowdowns under SWI - some of our tests take 170 times as long on SWI compared to SICStus, which makes it difficult to measure if my `block` optimization attempts are actually doing anything.

### DIFF
--- a/library/dialect/sicstus/block.pl
+++ b/library/dialect/sicstus/block.pl
@@ -39,12 +39,10 @@
 
 /** <module> Block: declare suspending predicates
 
-This  module  realises  SICStus  Prolog   =|:-  block  BlockSpec,  ...|=
-declarations  using  a   wrapper   predicate    that   calls   the  real
-implementation through a coroutining primitive   (typically  when/2, but
-freeze/2 for simple cases).
+This module provides SICStus Prolog-compatible
+=|:- block BlockSpec, ...|= declarations for delaying predicate calls if
+certain arguments are unbound.
 
-@tbd	This emulation is barely tested.
 @see	https://sicstus.sics.se/sicstus/docs/3.12.11/html/sicstus/Block-Declarations.html
 */
 
@@ -79,14 +77,13 @@ head(H, Head) :-
 %	more :- block declarations. The   predicate  is suspended untill
 %	all mode patterns that apply to it are satisfied.
 %
-%	The implementation is realised by creating   a wrapper that uses
-%	when/2 to realize suspension of the renamed predicate.
+%	The implementation is realised by creating a wrapper that checks
+%	the block conditions and either calls the original predicate
+%	immediately (if none of the block conditions were true) or uses
+%	attributed variables to delay re-evaluating the block condition
+%	until any of the arguments in question are bound.
 %
 %	@compat SICStus Prolog
-%	@compat If the predicate is blocked on multiple conditions, it
-%		will not unblock before _all_ conditions are satisfied.
-%		SICStus unblocks when one arbitrary condition is
-%		satisfied.
 %	@bug	It is not possible to block on a dynamic predicate
 %		because we cannot wrap assert/1.  Likewise, we cannot
 %		block foreign predicates, although it would be easier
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  Before I apply this, I'd like to point at wrap_predicate/4.  This simplifies the wrapping, makes predicate properties transparent and allows wrapping dynamic predicates, removing one more bug.  You find examples in boot/tabling.pl as well as in trace/1,2.

A 170 times slowdown is pretty ridiculous.   If you can isolate and share the slow code, please do so.  Contact me by email if the code is proprietary but you want to share (some of) it after settling the conditions.

dgelessus:
Ah, nice, I didn't know about wrap_predicate/4 - I just used the manual wrapping code that was already there. Looks like it shouldn't be hard to replace the manual code with wrap_predicate. Just one thing I'm unsure about - according to the docs:

> Registered wrappers **are not part of saved states** (see qsave_program/2) and thus need to be re-registered, for example using initialization/1.

Would that be a problem here? I assume the term expansion on the `:- block` declarations would only run at compile time, so if we use wrap_predicate, we would need to make sure it gets called again at runtime. Would it be enough to generate an extra clause like `:- initialization wrap_predicate(...)`, rather than calling `wrap_predicate` directly from the term expansion?

re. the performance problem: I'll let you know once I have a rough idea where it's coming from. I realize I never actually told you - the project I'm working on making SWI-compatible is the [ProB animator and model checker](https://prob.hhu.de/w/), which is EPL-licensed. A nightly source tarball can be downloaded [here](https://www3.hhu.de/stups/downloads/prob/source/ProB_src.tgz) (the Git repo isn't public). Feel free to take a look if you're curious, but right now it won't run usefully under SWI - it requires a few more SWI patches that I still need to submit. Once I've done that, I'll have another look at the performance problem and can hopefully send you a small test case to reproduce it.

JanWielemaker:
> Would it be enough to generate an extra clause like `:- initialization wrap_predicate(...)`, rather than calling `wrap_predicate` directly from the term expansion?

In think that should do.  

> I realize I never actually told you

Thanks.   I was getting curious :smile:    On all large projects I've been involved in so far SWI-Prolog performed roughly in the same league as SICStus.   On benchmarks the story is different (as you probably know).  I can help :)

dgelessus:
Alright, here's the implementation reworked based on wrap_predicate/4. It seems to be a little bit slower than the previous implementation with manual wrapping, but nothing huge. Our core unit tests took around 330 s with the manual version and around 340 s with wrap_predicate/4. Both numbers are still inflated from the aforementioned performance problem, so the difference might actually be more significant. Fast enough for now though 🙂

dgelessus:
(force-pushed to fix a trailing space that I overlooked)

JanWielemaker:
> Our core unit tests took around 330 s with the manual version and around 340 s with wrap_predicate/4.

This is a little unexpected.  The wrapping is actually doing less work.  As the code is not used often it may of course lead to cache misses.   Had a quick look at the ProB page.  Looks like a great project.   Could the performance issues be related to clp(fd)?   As  most of the SICStus clp(fd) is in C rather than in Prolog I can imagine a rather big difference.   Still, 170 times suggests there is something more fundamentally wrong such as a missed indexing opportunity.   AFAIK SWI-Prolog's indexing has more features, which makes this a little unlikely.

JanWielemaker:
Merged.   If I get it right though, rename_clause/3 is now dead code and the second clause of the term_expansion/2 can also go, no?

dgelessus:
You're right, rename_clause/3 is no longer needed - I forgot to remove that. The second term_expansion/2 clause is still needed though - it calls block_wrapper_clauses/3, which sets up the wrapper and generates a helper predicate used by the wrapper.

By the way, I figured out at least part of the performance problems I was having. I use my IDE (CLion) to set up the CMake build, and for some reason it explicitly sets `CMAKE_BUILD_TYPE` to `Debug` by default, so I was running a completely unoptimized build of SWI. I've now changed the build type to `RelWithDebInfo` - which SWI's CMake build would have used by default, if my IDE hadn't overridden it. That improved the performance by 2-3x, and the one particularly slow test now takes "only" 12 seconds. Still not very fast, but much better than before.

PR #858, opened 2021-07-06
Add partial library(clpfd) to SICStus 4 emulation

dgelessus:
This is obviously nowhere close to a full emulation of the SICStus CLP(FD) API. These are just a few simple predicates/operators that SWI's CLP(FD) already supports under different names.

### DIFF
--- a/library/dialect/sicstus4/clpfd.pl
+++ b/library/dialect/sicstus4/clpfd.pl
@@ -1,0 +1,75 @@
+/*  Part of SWI-Prolog
+
+    WWW:           http://www.swi-prolog.org
+    Copyright (c)  2021, SWI-Prolog Solutions b.v.
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions
+    are met:
+
+    1. Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in
+       the documentation and/or other materials provided with the
+       distribution.
+
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


triska:
This is not a good change:

The reason I am using `(#==>)/2` instead of `(#=>)/2` is that `#=>` can be easily mistyped as `#>=`, and then it means something else entirely. This is a common beginner error.

I therefore decided to support only `#==>` and `#<==` for implication, which is what GNU Prolog uses: GNU Prolog likewise does not support `#=>`, but does provide `#==>`.

The strategic goal of this limited support is to **make SICStus support `#==>`**.

This change undermines this goal, because with this patch, application programmers will no longer feel any need to talk to SICStus about this issue and ask them to provide `#==>` too for GNU Prolog and SWI compatibility. Instead, GNU&nbsp;Prolog, which has done everything right in this case, may now be pressured to cater to the SICStus and SWI behaviour to provide `#=>` because two other systems now support it.

dgelessus:
The rationale for `#==>` instead of `#=>` to avoid confusion with `#>=` makes sense. To clarify, I'm not adding these aliases because I think they're better, but simply because this is the SICStus compatibility library and these are the names that SICStus uses. I'm not planning to add the SICStus spellings to the main library(clpfd), so there should be no risk of confusion for beginners (or most users really), who won't be using the SICStus compatibility library.

I understand the goal of getting SICStus to support the more common spelling of these operators. (By the way, [YAP also uses `#==>` etc.](https://github.com/vscosta/yap/blob/bbb98b24dc36e96db0dea86ea32862bb6131b3d5/packages/clpfd.yap#L33-L35), so you can add that to your list.) But realistically, even if this change wasn't accepted into SWI, I would instead put the same code into the codebase that I'm working on (which is *very* SICStus-specific at the moment, though I'm trying to change that). Even if I could convince the SICStus developers to support `#==>` in the next version, I wouldn't be able to use it until they release SICStus 4.8 in 2023 or so. In the meantime, I think it's more productive to use this solution than to keep writing SICStus-only code until they add support for `#==>`.

triska:
It is heartbreaking to see even such small attempts at improvements undermined with such selfish and shortsighted arguments and changes.

Have you already contacted SICStus about this issue?

Please act for the whole, not only for your own specific use case: The only improvement in this situation is for SICStus to provide `#==>`, every other change makes this improvement less likely.

PR #862, opened 2021-07-19
SICStus-compatible "FD set" API for CLP(FD)

dgelessus:
This is mostly a feature suggestion/proposal. I'm submitting it as a PR so I can attach the WIP implementation that I've already written.

SICStus `library(clpfd)` provides a collection of ["FD set" predicates](https://sicstus.sics.se/sicstus/docs/4.6.0/html/sicstus/FD-Set-Operations.html), which allow inspecting and manipulating the domains of CLP(FD) variables. The purpose of these predicates is similar to the existing `fd_dom/2`, but they are more efficient, because they operate directly on the internal domain representation without translating it back to the user-facing domain syntax. In addition, the FD set API provides basic set operations on domains (member check, add/remove element, union, intersection, etc.), which are inconvenient and inefficient to implement on top of `fd_dom`.

The FD set API does not require any specific domain representation - the domain is treated as an opaque term and may only be manipulated using FD set predicates. As you can see in the PR, I was able to implement most FD set predicates very easily, even though SWI does not use the same domain representation as SICStus.

IMO it would make sense to support the FD set predicates natively and not just as part of the SICStus emulation, because the predicates provide features that are not easily available with SWI's current API, and because the implementation naturally depends on a lot of `library(clpfd)` internals. If this is okay, I can integrate this implementation into the native `library(clpfd)` and finish implementing the remaining FD set predicates.

### DIFF
--- a/library/clp/clpfd.pl
+++ b/library/clp/clpfd.pl
@@ -90,6 +90,7 @@
                   op(700, xfx, #\=),
                   op(700, xfx, in),
                   op(700, xfx, ins),
+                  op(700, xfx, in_set),
                   op(450, xfx, ..), % should bind more tightly than \/
                   (#>)/2,
                   (#<)/2,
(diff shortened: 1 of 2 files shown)


JanWielemaker:
> If this is okay, I can integrate this implementation into the native `library(clpfd)` and finish implementing the remaining FD set predicates.

I think moving this into clp(fd) is a good idea, especially if it is SICStus compatible.  Thanks!

dgelessus:
Alright, I've force-pushed the complete and natively integrated version (with a few bugs less than before). The force-push diff shown by GitHub is useless - I moved all of the FD set code to a different file, and I also did a rebase, so the big LD refactor is part of the diff. The actual commit history is clean though.

This should implement all of the [FD set predicates that SICStus has](https://sicstus.sics.se/sicstus/docs/4.6.0/html/sicstus/FD-Set-Operations.html). There are quite a few predicates, but most have very simple implementations, because SWI library(clpfd) already had internal predicates that do almost the same thing. The only somewhat complex predicate is `fdset_parts/4`, because SWI's representation of split domains is a bit different than the SICStus one.

JanWielemaker:
Thanks.  Merged after rebase.   Added a small comment to the header.

PR #863, opened 2021-07-20
LDFUNC refactor

dmchurch:
This PR applies a refactor across the entire SWI-Prolog codebase, changing the syntax of passing LD from one function to another in internal library code. It should get merged without squashing, so that the separation between automated and manual changes is maintained.


JanWielemaker:
Thanks!   Merged verbatim, but after local rebase github misses that it is merged :cry:

The unexpected result seems to be an almost 10% performance gain!   Is this because having LD in always the same first argument position keeps in in the same register?   Whatever it is, it is welcome :smiley:

dmchurch:
That was one of my reasons for moving it there, yes! It's now (in x86/x64 calling convention, at least) occupying the same register that the `this` pointer would in C++, and I imagine there are quite a few compiler optimizations that can take advantage of.

The other thing that could be helping is that `GET_LD` and (especially) `PRED_LD` functionality is now largely automatic - if the compiler sees a variable called `__PL_ld` or `PL__ctx->engine` in scope, it'll use that as the value for the `LD` macro rather than falling back to the `GLOBAL_LD` definition. (see the notes about `ANY_LD` in src/pl-builtin.h.) So, we may be seeing more usage of passed LD in modules that haven't redefined LD as LOCAL_LD, and/or we may be seeing LD getting passed through to functions that never had a pl-ldpass.h macro defined in the first place? Who knows.

Regardless, hooray! Very exciting results.

PR #874, opened 2021-08-05
ADDED: binformat package, for binary format utils

dmchurch:
Initial checkin of binformat package, with integer-conversion predicates. The repository at https://github.com/dmchurch/packages-binformat can be moved, obviously. 🙂

Todo:
1. Unit tests
2. Float/double primitives
3. Structure/array support (DCG compilation)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -118,3 +118,6 @@
 [submodule "packages/language_server"]
 	path = packages/language_server
 	url = ../packages-language_server.git
+[submodule "packages/binformat"]
+	path = packages/binformat
+	url = ../packages-binformat.git
--- a/cmake/PackageSelection.cmake
+++ b/cmake/PackageSelection.cmake
@@ -39,6 +39,7 @@ endforeach()
 # The pckages below do not depend on external libraries except for
 # the zlib package, but the core system already depends on zlib.
 set(SWIPL_PACKAGE_LIST_BASIC
+    binformat
     chr
     clib
     clpqr


dmchurch:
Closing for now after discussion with @JanWielemaker, this idea needs work before it's ready for integration.

kamahen:
It might be nice to have the protobuf package use these routines ... the current code is at https://github.com/SWI-Prolog/contrib-protobufs/blob/master/protobufs.c and there are a bunch of test cases in test_protobufs.pl and interop/test_interop.pl (I used the Python and C++ versions of protubufs to write/read some of the data, to ensure it was properly compatible).

kamahen:
I would want an error if a value is truncated, not “silently truncate out-of-range integers”. That’s the behavior currently in library(protobufs), in protobufs.c, and it also matches the behavior of the Python protobufs implementation (as opposed to the C++ implementation, which has the usual C++ silent truncation) … in some places, the current protobuf serialization/deserialization does an explicit test; in other places, it gets the test for free from PL_get_uint64_ex() et al.

JanWielemaker:
This is an example why I think a binary format library is not a good idea.  There are simply too many binary formats and too many specific restrictions or extensions in the many binary protocols.  I see protobufs uses _zigzag_ encoding for integers.  I see it uses little endian where most protocols use big endian.   As we get to strings, there are even more choices (length vs end-character, encoding of code points, etc.)

If you want speed, use C.  If you don't care about speed you can eventually most in Prolog or you can use the `ffi` package to get access to bit patterns from e.g., floating point numbers.

kamahen:
I like the binformat idea of using macros. The protobufs.c code is essentially the same for its various combinations (float, int/uint, 32/64 bit) and has a lot of repetition. (But I'm not going to change what seems to be working code.)

BTW, writing the various test cases was a PITA, and I'm not sure how easily that can be parameterized.

dmchurch:
> This is an example why I think a binary format library is not a good idea.  There are simply too many binary formats and too many specific restrictions or extensions in the many binary protocols.

You're not at all wrong. My intent, incidentally, has never been for these types of predicates to be used as-is, but for them to serve as _indicators_. Indicators that can be used on their own, to be certain, but there isn't really a significant performance difference between these two implementations:

```prolog
short_array([]) --> [].
:- if(use_binformat).
short_array([H|T]) --> bin_uint16_be(H), short_array(T).
:- else.
short_array([H|T]) -->
    [Hi, Lo],
    {   var(H)
    ->  H is Hi * 256 + Lo
    ;   divmod(H, 256, Hi, Lo)
    },
    short_array(T).
:- endif.
```

The overhead from calling out to a foreign function is enough that it outweighs any slight gains in calculation speed, and that remains true even if you're reading/writing an array of values and you wrap the `bin_uint16_be` function in a tight C loop.

That's why my plan on this is to be something like Python's `struct` module, only more powerful. Python uses a printf-style format string to describe the structure of a binary blob, but the most Prolog-esque way to represent the same is a DCG. So basically, I'd want to recompile the DCG _itself_ into a binary stream recognizer/generator, so long as it used a limited and predictable subset of Prolog in its specification.

So, the biggest reason for exposing these `bin_*` predicates is that they function as a semantics definition and reference; any DCG that can recognize/generate a given binary stream when called with a standard `phrase(grammar(Value), CodeList)` and is simple enough for the binformat compiler to understand, will recognize/generate the same stream but with near-native speed when called as `bin_phrase(grammar(Value), String)`.

Hope that cleared up some of the intent a bit, and yes, @kamahen definitely feel free to borrow on some of these compiler techniques for the protobufs code! I think a fully-fleshed-out binformat could provide a useful low-level primitive for the protobuf implementation, but as it stands there's no particular reason to pull this in as a dependency or anything.

kamahen:
There's a related discussion for Python here: https://mail.python.org/archives/list/python-ideas@python.org/thread/V2EKXMKSQV25BMRPMDH47IM2OYCLY2TF/#V2EKXMKSQV25BMRPMDH47IM2OYCLY2TF

PR #875, opened 2021-08-09
Avoid rerunning mkvmi repeatedly for no reason

dgelessus:
This is a pretty minor thing, but I noticed that every time I run `cmake --build`, it regenerates vmi-metadata.h for no apparent reason. The CMake scripts say that it depends only on pl-vmi.c and the mkvmi utility itself, neither of which should change on every build. So I debugged this further using `make -d` and found this in the output:

```
       Prerequisite `../../src/pl-vmi.c' is newer than target `src/vmi-metadata.h'.
       Prerequisite `src/mkvmi' is newer than target `src/vmi-metadata.h'.
      Must remake target `src/vmi-metadata.h'.
```

As it turns out, `mkvmi` tries to be smart - if the generated file already exists with identical contents, it doesn't overwrite the file to avoid touching the mtime. Unfortunately this causes problems here - `mkvmi` is being called because the output file is outdated, but then it doesn't actually update the output file's mtime, so on the next build the file is still considered outdated and `mkvmi` is called again, etc.

This PR removes the extra check and simply rewrites the output file every time, which fixes this problem. The downside is that now every time that pl-vmi.c changes, all dependents of vmi-metadata.h (i. e. most of the SWI-Prolog core) are rebuilt, even if the contents of vmi-metadata.h haven't actually changed.

The same problem applies to defatom, but it's less of an issue there, because changing the ATOMS input file will almost always cause a real change to the generated headers.

### DIFF
--- a/src/defatom.c
+++ b/src/defatom.c
@@ -62,45 +62,11 @@ argument that specifies the source directory.
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - */
 
 static int
-cmp_file(const char *from, const char *to)
-{ FILE *f1 = fopen(from, "r");
-  FILE *f2 = fopen(to, "r");
-  int rc = -1;
-
-  if ( f1 && f2 )
-  { char l1[1024];
-    char l2[1024];
-
-    while ( fgets(l1, sizeof(l1), f1) )
-    { if ( fgets(l2, sizeof(l2), f2) )
-      { if ( strcmp(l1, l2) == 0 )
-	  continue;
-      }
-      goto out;
-    }
-    if ( !fgets(l2, sizeof(l2), f2) )
-      rc = 0;
-  }
-
-out:
-  if ( f1 ) fclose(f1);
-  if ( f2 ) fclose(f2);
-  return rc;
-}
-
-
-static int
 update_file(const char *from, const char *to)
-{ if ( cmp_file(from, to) == 0 )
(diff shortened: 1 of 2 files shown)


JanWielemaker:
I'm not sure this is a good or a bad step.   mkvmi is a quick step while an unneeded build of the whole is a lot slower.   I wonder whether there is a more elegant way to solve this?   It should be a known problem ...

dgelessus:
That's true, and I can understand it you want to leave it the way it is right now because it's quicker for active development. mkvmi runs fast enough that it doesn't really matter if it runs more often than needed.

I don't know if there is any way to handle this properly, i. e. mark the output as updated without also causing a rebuild of everything that depends on it. I feel like this is impossible with the simple make-style model where only the mtime matters - either you update the output mtime, which makes all dependents of the output outdated, or you don't update the mtime, which leaves the output as outdated. Maybe there's some advanced CMake feature that can help with this, but if there is, I don't know about it...

JanWielemaker:
There has been lots of discussion on this around ninja.  See e.g. https://github.com/ninja-build/ninja/issues/1459.   I propose to keep things as is for now.   Not that for development of the Prolog libraries there is no need to build the system.   Just use the binary from `src/swipl` and edit the Prolog sources in the build tree.   Only if you change anything to the files in `boot` you need to run `ninja core` to get the quickest rebuild.

PR #879, opened 2021-08-30
Add SWILIBDIR to buildenv.sh

mgondan:
Add linker flag -LDIR pointing to the directory DIR with the shared libraries (e.g., libswipl.so)


mgondan:
not needed, sorry

PR #880, opened 2021-08-30
FIXED: Adding missing newlines in tracer commands

dmchurch:
Listing, in particular, suffers from not having the newline printed after the command name, since the first line ends up being smashed against the right margin.


JanWielemaker:
Thanks!

PR #893, opened 2021-10-10
Move to classic unconfined snap package

LyzardKing:
As discussed this PR adds the unconfined snap package.
The package will be larger since it must include the qt5 libraries needed by the ui.
It should work with most (if not all) external packages, since it has full access to the system (as if it were installed via apt)


JanWielemaker:
We moved to flatpak since.   Closing.

PR #895, opened 2021-10-14
Rbtree robustness

kamahen:
These changes were inspired by a bug I found in my code, that passed an uninstantiated Tree to rb_visit/2. In the process, I added some basic tests for each predicate in the public interface for library(rbtrees).

### DIFF
--- a/library/rbtrees.pl
+++ b/library/rbtrees.pl
@@ -99,8 +99,6 @@ form colour(Left, Key, Value, Right), where _colour_  is one of =red= or
     rb_apply(+,+,2,-),
     rb_fold(3,+,+,-).
 
-:- use_module(library(error), [must_be/2]).
-
 /*
 :- use_module(library(type_check)).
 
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.   I've pushed some further patches to the `rbtree` branch on the main repo.   Should have pushed to this PR, but I has rebased before, so it got a bit of a mess :cry: 

The basis looks good.    I think using det and => makes the code a lot more robust at very low cost in terms of making the code less readable or slowing it down.   I propose to get rid of the msg exceptions.   They fit poorly in the system where all system and library exceptions are error(_,_) exceptions.   One of the commits addresses that.  I think the fold one cannot be marked det as the closure may fail or even be nondet, no?  Most other commits are trivial.   

The one that needs some discussion is the partial map.   If I read the code correctly, success doesn't depend on the closure as that is called in an if-then-else.   That suggests it can be det.   But, the keys need to be sorted without duplicates and all keys need to be in the tree (according to the docs).   I've added a sort/2 to avoid the first two claims.  I very much doubt anyone on purpose wants this to fail if the keys are not sorted.   Now we still have a failure if a key is not in the tree.   I wonder whether there is a sensible use case for that or we better throw and existence_error and mark the predicate as det.

Comments?    If you agree I'll squash my changes into your commit.

kamahen:
I've added a few comments to your commits.

Your commits seem to cherry-pick my changes, so can this PR be deleted?

PR #905, opened 2021-11-14
skip post_install_foreign in pack w/o foreign code

mgondan:
pack_post_install failed when the pack was a normal prolog pack (without foreign code)


JanWielemaker:
Thanks.  Fixed a little differently.

PR #911, opened 2021-11-28
PORT: First attempt to support MacOs M1 with brew deps in /opt/homebrew

jrvosse:
I tried to split the homebrew code in a version for a (Intel) homebrew prefix /usr/local  and a (ARM/M1) prefix /opt/homebrew.

This is ugly but should be ok.

I'm a bit worried about the pathnames including the exact version of the archive, ssl and readline packages:
  set(OPENSSL_ROOT_DIR /opt/homebrew/Cellar/openssl@3/3.0.0_1)
  set(LibArchive_ROOT /opt/homebrew/Cellar/libarchive/3.5.2)
  set(Readline_ROOT /opt/homebrew/Cellar/readline/8.1.1/)

Suggestions welcome.

### DIFF
--- a/cmake/port/Darwin.cmake
+++ b/cmake/port/Darwin.cmake
@@ -1,10 +1,13 @@
 if(APPLE)
 
+set(CMAKE_FIND_APPBUNDLE NEVER) # Make sure cmake does not find the Apple gui version of GIT
 if(NOT MACOSX_DEPENDENCIES_FROM)
   if(EXISTS /opt/local/bin/port)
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.   Added some stuff to allow specifying the deps again as `Homebrew`, causing the system to figure out which install base.   Also added some logic to avoid the exact version.   I wonder though whether we need to look into the "Cellar".   I thought that was for not installed stuff.    Isn't the structure similar than for the old Homebrew?

Don't worry about the conflict.   The patch is merged after edit.

PR #912, opened 2021-12-01
Rbtrees -fix bug in is_rbtree/1, add more tests

kamahen:
Also, rb_map/3 can now backtrack, same as maplist/3 (I assume that this was a design oversight, or an attempt at premature optimization).

### DIFF
--- a/library/rbtrees.pl
+++ b/library/rbtrees.pl
@@ -35,13 +35,13 @@
 :- module(rbtrees,
           [ rb_new/1,                   % -Tree
             rb_empty/1,                 % ?Tree
-            rb_lookup/3,                % +Key, ?Value, +Tree
-            rb_update/4,                % +Tree, +Key,          +NewVal, -NewTree
-            rb_update/5,                % +Tree, +Key, ?OldVal, +NewVal, -NewTree
+            rb_lookup/3,                % +Key, -Value, +Tree
+            rb_update/4,                % +Tree, +Key,          ?NewVal, -NewTree
+            rb_update/5,                % +Tree, +Key, -OldVal, ?NewVal, -NewTree
             rb_apply/4,                 % +Tree, +Key, :G, -NewTree
             rb_insert/4,                % +Tree, +Key, ?Value, -NewTree
             rb_insert_new/4,            % +Tree, +Key, ?Value, -NewTree
-            rb_delete/3,                % +Tree, +Key, -NewTree
+            rb_delete/3,                % +Tree, +Key,       -NewTree
             rb_delete/4,                % +Tree, +Key, -Val, -NewTree
             rb_visit/2,                 % +Tree, -Pairs
             rb_keys/2,                  % +Tree, +Keys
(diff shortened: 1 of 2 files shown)


library/rbtrees.pl
```diff
@@ -35,12 +35,12 @@
 :- module(rbtrees,
           [ rb_new/1,                   % -Tree
             rb_empty/1,                 % ?Tree
-            rb_lookup/3,                % +Key, -Value, +T
-            rb_update/4,                % +Tree, +Key, +NewVal, -NewTree
+            rb_lookup/3,                % +Key, ?Value, +Tree
```
JanWielemaker, line 38: This should be -Value in SWI-Prolog's mode annotation.  Any - argument may be instantiated at call, resulting in the same behavior as `p(X), X = value`.   Things that should strictly be unbound at entry must be marked with `--Arg`

library/rbtrees.pl
```diff
@@ -141,29 +141,31 @@
...
 %   complexity is O(log N) in the number of elements in the tree.
+%   Throws an error if Key is not sufficiently instantiated (see rb_in/3
+%   if you wish to backtrack over all keys).
 
 rb_lookup(Key, Val, t(_,Tree)) =>
     lookup(Key, Val, Tree).
 
-lookup(_, _, black('',_,_,'')) :- !, fail.
-lookup(Key, Val, Tree) :-
+lookup(Key, _Val, black('',_,_,'')), nonvar(Key) => fail.
```
JanWielemaker, line 156: rbtrees (and assoc) do allow for unbound keys.   This breaks code.

library/rbtrees.pl
```diff
@@ -141,29 +141,31 @@
...
+lookup(Key, Val, Tree) =>
```
JanWielemaker, line 157: I have my doubt about => in this case as well as lookup/4.   Given a single clause and fully unbound arguments it merely makes the code slower and less traditional for no reason.  For lookup/4 we know the first argument comes from compare/3 and is thus always >, = or <.    IMO the main purpose of => for libraries like this is to get cheap type tests (for the rbtree) and make sure the code is steadfast.

kamahen:
I've added a few more "=>"s, such as next/8, previous/8, enum/3; I've also added a few more test cases (rb_in/3).

library/rbtrees.pl
kamahen: Done.
I think there were some mistakes in the original code (before I modified it), so please review all the declarations, both in the `module` declaration and the `%!` comments (not just the ones that I've changed).

I'm leaving `+Key` as-is; I'd prefer `++Key` because variables in the keys can cause subtle bugs and I can't think of a use for them. (See my reply to your comment on line 154 for variables in `rb_lookup/3`.)

library/rbtrees.pl
kamahen: Python requires that keys be effectively "const" - that is, they must have a hash function that doesn't change (this is enforced by a convention that mutable types shouldn't have hash functions ... Prolog, of course, is not mutable (except for `setarg/3` et al) but having a variable in the term makes it act like a mutable value when used as a key.

A variable in a key won't behave the way most people expect (even if they read the comment about `(==)/2` being used and not `(=)/2`). I've removed the "nonvar" checks for `rb_lookup/3`, `rb_update/*`, `rb_delete/*` `ord_list_to_rbtree/2`, but I'm not happy with that (I've left them as "%TODO: nonvar(...)"). My view is that variables in the key that participate in a comparison are an error waiting to happen.

I've changed the comment in `rb_lookup/3` and others to strongly advise against variables in keys. I've also marked the tests that I think should throw an error with "TODO"s.

library/rbtrees.pl
kamahen: The "=>" in lookup/3 is to catch uninstantiated leaves in the tree (which are unlikely; the most common situation is already caught in rb_lookup/3. The original code was a cut/fail, so it's probably about the same level of (in)efficiency as with "=>", unless I misunderstand ssu very badly. 
I've also changed rb_in/3 to use "=>" (same reason as for lookup/3).

As for lookup/4 - I've reverted the "=>"s - they're not necessary by code inspection and they're marginally slower. (Also for next/8 and previous/8)

library/rbtrees.pl
JanWielemaker: nonvar/1 is not good enough.   You either need ground/1 or verify that ?=/2 holds between each pair of keys.   Oops, ?=/2 isn't good enough either because a(X,1) and a(Y,2) satisfy ?=/2, but may be reordered using unification.   Similar issues applies to sort/2, @</2 and anything that relies on standard order of terms.   The ordsets library is pretty useful for effectively computing e.g. the set if shared variables between two (large) terms but relies on the same assumptions.

I see little reason to advise against relying on standard order for non-ground terms.   Instead I prefer to point users at the fact that unification may change ordering and therefore one should either understand the consequences or ensure all keys are ground.   Possibly this should be a section in the manual one can refer to.

kamahen: The `nonvar` tests have already been removed, replaced by "TODO"s. I'll add a comment to the code about those TODOs -- as you say, it would be even better to document the subtle problems.

For what it's worth, Python doesn't prevent creating a hash function for a mutable structure and makes no attempt to detect that a key's hash has changed in a Python "dict". This seems to be acceptable behavior in the Python community, so the similar Prolog problem of keys containing variables should also be something that we don't need to check. (Python will throw an error if a (mutable) structure without a hash function is used as a key with a Python "dict" -- in Prolog, `term_hash(foo(X),Hash)` succeeds with `Hash` remaining uninstantiated ... is this intended behavior?)

BTW, I did the original changes from ":-" to "=>" because of a typo in my code (which the singleton variables check didn't catch) that resulted in using an empty tree; I then though it might be useful to at least do a partial check for dangerous keys but doing this properly seems to be too complicated.

kamahen: I've changed the TODOs in commit f52057b.

An earlier commit had already added this warning to rb_lookup, rb_update, rb_delete:

> This predicate may fail or give unexpected results if Key is not sufficiently instantiated - see `rb_in/3` for backtracking  over keys.

JanWielemaker:
Great.   Merged after (1) adding a general warning about non-ground keys to the module overview, (2) removing the instantiation check TODO comments because we will never do these :smile: and (3) re-justifying some of the PlDoc comments.   I've squashed the commits.   The details seem irrelevant.   For now we still have the branch should we need it to fix a problem.

Thanks!

PR #920, opened 2022-01-10
qsave: add missing load_module for shlib

rrooij:
The predicate current_foreign_library is used which is from the module shlib. Unfortunately it is not imported though, which makes it impossible to use qsave without autoloading enabled.

### DIFF
--- a/library/qsave.pl
+++ b/library/qsave.pl
@@ -43,6 +43,7 @@
 :- use_module(library(option)).
 :- use_module(library(error)).
 :- use_module(library(apply)).
+:- use_module(library(shlib)).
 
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.  I messed up a little, trying to edit the commit but only updating the message :cry:  Had to push a subsequent commit to get the changes I wanted: another missing dependency and use autoload/2 rather than use_module/1 so that we still do not load the foreign code when we create a saved state from a pure Prolog program.

Despite the reported conflict, all is pushed.  Closing.

rrooij:
Thanks for looking into it. :smile: :+1:

PR #922, opened 2022-01-14
ADDED: valid_term_position/2 to help debug term_expansion/4

kamahen:
I've also added to some related documentation, partly based on https://swi-prolog.discourse.group/t/term-expansion-2-in-a-module-user-or-system/3055/5


kamahen:
The section I added to module.doc is probably not quite right. I know that module `prolog` is used in some places, but default_module/2 doesn't seem to reference it. Also, default_module/2 should probably be mentioned both in module.doc and in the section on expand_term/2 (and term_expansion/2) and related predicates.

kamahen:
Splitting this PR into two and deleting it.

PR #923, opened 2022-01-15
DOC: some more details about term_expansion

kamahen:
These items came up while I was working on EDCG's term_expansion/4

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -1031,7 +1031,7 @@ If there is no local definition, or the local definition fails to
 translate the term, expand_term/2 will try term_expansion/2 in module
 \module{user}. For compatibility with SICStus and Quintus Prolog, this
 feature should not be used. See also expand_term/2, goal_expansion/2 and
-expand_goal/2.
+expand_goal/2, default_module/2.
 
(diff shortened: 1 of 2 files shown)


JanWielemaker:
This seems a bit ad-hoc to me.   I think we should have a proper description of module inheritance and how it impacts operators, some other syntax flags, visibility of predicates and term/goal expansion.   The predicates are in sec:importmodule, but with little description of the underlying mechanisms and why it is like this.   Given that the various parts affected by module inheritance can just point to this section.   I'll see whether I can write that.

JanWielemaker:
I've pushed 65c3ab7d301ee056c5728b89e1d35b210ffbd6b3 and 761f9625255462f9ab34b7a34722e23d605bf679 to address these documentation issues.   I did not describe the `prolog` module.   I consider that a rather different matter and the `prolog` module is not really special.   We use a similar technique in e.g. the HTTP framework where all hooks that relate to this framework are in the `http` module.

Comments/updates/fixes welcome.

kamahen:
I made a few comments at https://github.com/SWI-Prolog/swipl-devel/commit/65c3ab7d301ee056c5728b89e1d35b210ffbd6b3

PR #924, opened 2022-01-15
ADDED: valid_term_position/2 to help debug term_expansion/4

kamahen:


JanWielemaker:
Reading through this, I'm more and more getting my doubt valid_term_position/2 is something useful.   You point at a lexical ambiguity wrt dict key/value pair ordering.   This isn't the only one though.    With the Prolog flag `double_quotes` set to `codes` or `chars`, the term-type for string_position(From,To) is a list.    As we do not know this flag at the time of reading (and it depends on the module), this gets hard.

A minor issue is that `quasi_quotation_position(From,To, SyntaxFrom,SyntaxTo, ContentFrom,ContentTo)` doesn't do what you think.    These are the positions of `{|Syntax||Content|}`.

So, it is possible to check that a term is a valid term position.   It is also possible that a term and its term position are compatible, meaning they __might__ belong together and it is possible to generate a position term (without positions) that is compatible with a term, but there are infinitely many so and this isn't just theory (parenthesis, strings, dicts, quasi quotations). 

This all makes me wonder about the added value of this predicate.   I can see some point in a check that a term and a term position are compatible.   Notably for testing read_term/2 and any code that plays around with rewriting layout terms.   A semidet predicate is probably not a good choice as for terms you find in practice you'll have a hell of a job to figure out what is wrong.

An alternative could be a subterm_pos(-SubTerm, -SubTermPos, +Term, +TermPos) or maybe  better, a position aware version of mapsubterms/3.   The latter can be a useful rewriting block and as it walks over the terms it may throw an exception of the Term and TermPos are not compatible.    A full check is then established using a failing converter, so the input remains the same as the output and the whole term is processed.    This could go into library(prolog_code) which is a collection of things to examine and manipulate Prolog code.

For now, I merged the comment on unify_clause/5 after edit.   I don't want to talk too much about checking imported dummy predicates.    What to do varies and it is after all not a great solution.    If anything, I'd mention it as a work around that is hopefully temporary.   I also merged the test additions, though without the call to valid_term_position/2.   After we make up our minds we can include that.

kamahen:
My original version used `=>`, which made it easier to find where there was an incompatibility. I later changed to `:-` because I thought it would be useful to be able to generate the term_position information. But -- as you point out -- there's no single term_position for some terms, so it might be best to go back to `=>`. 

(BTW, I ran into a similar problem with protobufs, which has predicates that can both generate and recognize, which I inherited from the original designer ... these turned out to be tricky to extend, a bit less readable (IMHO), and if I had a clean start, I would have two one-way predicates instead of a single "reversible" predicate.)

kamahen:
I've made a ssu-version - validate_term_position/2 here: https://github.com/kamahen/edcg/commit/40894343fe62698a37e5c6b46cb442360de6bf3d

JanWielemaker:
> I've made a ssu-version

I think this is much better as it now serves a clear purpose.    I'm happy to have this in library(prolog_source) (I think).    Note that the string_position layout term may also be associated with a code or char list.   I close this PR.

kamahen:
I've put the ssu version of validate_term_position/2 into prolog_source, and added tests for the various permutations of reading strings in test_read.pl (also added a test for validate_term_position/2 there).

PR #925, opened 2022-01-18
ADDED: prolog_source:validate_term_position/2 to help debug term_expansion/4

kamahen:


JanWielemaker:
Thanks.  Merged after edits.   Roughly:

   - Renamed back to `valid_term_position` instead of `validate_term_position`.     This is a logical proposition rather than an action, so it should not be a verb.   At least, that is what Richard O'Keefe learned me :smile: 
   - Layout and PlDoc fixes.
     - Terms do not need ``=||=``.   Anything functor(....) that is valid Prolog syntax is considered Prolog.  Anything else that is valid Prolog should be in single backquote.  Anything else that should be code style should be in double backquote.  This follows current markdown practices, only distinguishing single and double backquotes.  Also use triple backquote for code blocks rather than `==`.  Both work, but the triple backquote is markdown.
     - `@param` must be `@arg` as once pointed out by the Ciao people when we tried to get agreement.   `@keywords` must be at the end. 
   - We cannot make the string_position related term depend on the Prolog flag.   The reading may have happened in a module with a different setting.   I see little choice but accept all as valid terms.
   - Updated imports (using PceEmacs using ``^C^D``)

The current code checks that nested terms have nested layout.   That is true for read_term/2.   I have my doubt this is ok for the output of term_expansion/4 though.   Expansion may move subterms.

PR #928, opened 2022-01-20
removed a pragma that suppressed a warning

mgondan:
I think the warning has disappeared for recent gcc/glibc


JanWielemaker:
Decided to ignore this patch and instead revert the patch that introduced this #pragma.   That makes it a bit more clear waht happened.

PR #929, opened 2022-01-20
use mallinfo2 instead of mallinfo (deprecated)

mgondan:
/usr/include/mallinfo.h says mallinfo is deprecated and mallinfo2 should be used instead


JanWielemaker:
Same story as the libc package patch.   Thanks for looking into this.  I wasn't aware of mallinfo2().

PR #937, opened 2022-02-15
FIXED: Preproc/build errors

dmchurch:
Adding a couple GCC builtins to the config tests, using them to force
short-circuiting of _LD_WITH_FALLBACK even in low-optimization modes
(e.g. Clang in DEBUG configuration).

Also adding a few missing-but-used defines to config.h.cmake
(HAVE___BUILTIN_EXPECT, HAVE___THREAD, HAVE_VOLATILE, and HAVE_INLINE)
and fixing an unbalanced #ifdef pair that got added in f3b79d6.


JanWielemaker:
Thanks!   A bit related, would it be possible to make LD work in gdb?   As is, macros that use LD, e.g. `(gdb) p gTop` do not work :cry:

dmchurch:
Ooh, good catch! Yeah, that should be possible somehow, I'll look into it 🙂

PR #938, opened 2022-02-16
PORT: Get optional symbols via weak linking

dmchurch:
This unifies the tcmalloc and ptmalloc linking code and also allows
systems without dlsym(), like WASM, to expose the malloc-implementation
predicates, so long as the linker is capable of resolving weak symbols.


JanWielemaker:
Thanks.  Results in this message (and a number more of them):  I guess that with linker support WEAK_IMPORT should really be nothing, (`(void)0`) no?

```
../src/pl-incl.h:2686:39: warning: statement with no effect [-Wunused-value]
 2686 | # define WEAK_IMPORT(Name) (&Name ? 1 : 0)
```

dmchurch:
There we go. The only difference is in `pl-incl.h`, where I added a tiny inline helper to make `&Name ? 1 : 0` look like a function call rather than a dangling expression 🙂

JanWielemaker:
Thanks.   I decided to only support mallinfo if there is no mallinfo2().  Actually, we didn't have to use weak symbols for that as libc always provides them (or not).   We merely need the weak symbols for the tcmalloc library which may not be linked.   Leaving in mallinfo itself leads to deprecated messages and I understand some of the package maintainers really want to avoid these.

Anyway, now compiles on Linux, Windows and MacOS and should allow to go on with WASM :smile:

PR #941, opened 2022-02-19
PORT: adding O_SIGNALS define in config.h

dmchurch:
One thing I've noticed as I work on the wasm port is needing to distinguish between "has the signal() function" (which Emscripten does) and "should perform signal-handling tasks" (which Emscripten should not). This also allows for switching the USE_SIGNALS option in cmake and having it take effect on the next build, rather than having to clear the cache entirely.

This provides a single define to check if Prolog's signal-handling is enabled, rather than checking each of HAVE_SIGNAL etc. This also makes the existence and use of `GD->signals.sig_alert` consistent on whether SIG_ALERT is defined.


JanWielemaker:
This is a good change.  It needs a bit debugging though.  The system fails to compile on MacOS (and probably on any POSIX system).   O_SIGNALS doesn't get set and the fact that it isn't set results in an incoherent usage of signals (it seems).   I probably would have patched it myself with access to a proper dev environment, but I'm behind a small laptop.

Most O_XYZ optiins are handled using `#ifdef`, which seems to make more sense as they are either defined or not in `config.h`

dmchurch:
Hmm, very curious that `O_SIGNALS` isn't getting set, since I'm running my compiles on the dev machine in the first place; I'll have to figure out what's going on with that. Probably a stale build environment or something.

The `O_XYZ` defines are actually some of the safest to use with a bare `#if`, because we know they'll either be defined to `1` or undefined, and the C99 standard specifies that any undefined identifier equates to `0`. You wouldn't want to use `#if` with anything that either (a) isn't a compile-time boolean (SIG_ALERT, for example) or (b) could be defined to an empty string somewhere with a

```c
#define O_XYZ
```

That said, if you'd like me to standardize on `#ifdef FOO`/`#if defined(FOO)` for consistency's sake, I'll certainly switch those up; I'll just continue bemoaning the lack of an `#elifdef` directive 🙂

dmchurch:
Aha, there we go. I forgot that CMake `set(VAR ... PARENT_SCOPE)` will _not_ set that variable in the current effective context, unlike `set(VAR ... CACHE)` which _will_. Rather than adding `O_SIGNALS` to the cache even though the others aren't in there, I've added a bit to src/CMakeLists.txt that looks for any variables named "O_*" defined at directory scope _at the time config.h is written_ and exports them to the top-level where packages can see them. This obviates the need for the `set(O_PLMT 1)` at [packages/cmake/PrologPackage.cmake:21](../blob/4c2748b/packages/cmake/PrologPackage.cmake#L21), though I'm leaving that in just out of caution.

(In case I failed to mention - the reason I ran into the variable-scoping bit in the first place is that I'm trying to get a working signal-free build on Linux, and the libedit package needs to know whether to disable its builtin signal handling. Once this PR is in, I'll submit one on the libedit repo to add in the `O_SIGNALS` tests.)

JanWielemaker:
Thanks.  Works fine now on MacOS, Linux and Windows.  I'm not 100% sure exporting all O_* names makes any sense.  Most are really local to the core.   nyway, there don't seem to be any conflicts so far.    I'm happy with the ``#if O_``.   Maybe we should do that globally, although IU'm a bit afraid that harms backporting fixes too much.   The ideal timing is shortly before a new stable/devel cycle ...

PR #943, opened 2022-02-24
MODIFIED: switching core libswipl code to C11

dmchurch:
This allows for a standards-compatible implementation of the LD macros
that is guaranteed to compile down to a simple reference even at -O0.
I've also added some code that, at least in my testing, makes
`print LD` work again in GDB and any debugger that uses it as a backend.


JanWielemaker:
Thanks!  That looks a lot easier to understand!  Builds and runs fine on MacOS.

PR #953, opened 2022-03-08
DOC: refer to pack `debug_adapter` in the GNU Emacs Interface page

eshelyaron:
Add a reference to [the `debug_adapter` pack](https://github.com/eshelyaron/debug_adapter) for GNU Emacs users.
I tried to keep the language concise in the spirit of the rest of the section, if more information is preferable I'll extend it.

See also https://github.com/SWI-Prolog/plweb-www/pull/47.


JanWielemaker:
Thanks.  Merged after changing \exam to \const.   \exam is meant for (inline) code examples.

PR #969, opened 2022-03-30
REP_NL_POSIX in PL_get_text

mgondan:


JanWielemaker:
Seems this should be unconditionally the case.   All internal strings use POSIX newline encoding.   Only when communicating with the outside we use the OS newline conventions.

mgondan:
Then the change is a single line ??



Von: Jan ***@***.***>
Gesendet: Mittwoch, 30. März 2022 09:06
An: ***@***.***>
Cc: Gondan-Rochon, ***@***.***>; ***@***.***>
Betreff: Re: [SWI-Prolog/swipl-devel] REP_NL_POSIX in PL_get_text (PR #969)


Seems this should be unconditionally the case. All internal strings use POSIX newline encoding. Only when communicating with the outside we use the OS newline conventions.

JanWielemaker:
> Then the change is a single line ??

Might be.    I'm checking where.   I see that this works as expected and wonder why ...   Possibly the flag should be set in Sopenmem() ...

     ?- with_output_to(string(S), write("hello\nworld")).

JanWielemaker:
Ok.   with_output_to/2 sets the POSIX NL encoding explicitly.   All other uses of Sopenmem() I could find should have done so, so I moved this setting to Sopenmem() and removed it from with_output_to/2.   If there is a case where this is wrong we should fix that there.

Lets see what the tests tell us.  They all pass on Linux (as to be expected).

mgondan:
I’ve seen the flag in src/os/pl-file.c, line 1508. And, in fact, if I disable it, I get

$ swipl
with_output_to(string(S), write("hello\nworld")).
S = "hello\r\nworld".




Von: Jan ***@***.***>
Gesendet: Mittwoch, 30. März 2022 09:23
An: ***@***.***>
Cc: Gondan-Rochon, ***@***.***>; ***@***.***>
Betreff: Re: [SWI-Prolog/swipl-devel] REP_NL_POSIX in PL_get_text (PR #969)


Ok. with_output_to/2 sets the POSIX NL encoding explicitly. All other uses of Sopenmem() I could find should have done so, so I moved this setting to Sopenmem() and removed it from with_output_to/2. If there is a case where this is wrong we should fix that there.

Lets see what the tests tell us. They all pass on Linux (as to be expected).

JanWielemaker:
I've pushed b0ca2a1937b5197af99c90c45b847406c0566960.   Could you continue from there and close these PRs if that seems the right direction?

PR #970, opened 2022-03-30
zlib 1.2.12

mgondan:
Due to the bug fixes, any installations of 1.2.11 should be replaced with 1.2.12.


JanWielemaker:
Note that README.mingw is no longer used.   It may still be useful info for your purposes :smile:   The distribution build happens with https://github.com/SWI-Prolog/docker-swipl-build-mingw, which uses the zlib library from the Fedora standard library for MinGW.

PR #971, opened 2022-03-31
Work around for broken modf in MinGW

mgondan:


JanWielemaker:
Thanks!   Merged after squashing and update of the commit message.     FYI, commits with visible impact on the system have a tag `[A-Z]+:` that gets them into the ChangeLog.   In your case the relevant tags are `PORT`, `FIXED` and `TEST`.  `PORT` if it concerns a fix for a specific platform, `FIXED` if it is a general bugfix and `TEST` if a test was updated, added, etc (has no impact on the system itself).

PR #972, opened 2022-03-31
FIX: conditions for test_answer

mgondan:
unsure if it is that simple, but it makes the "answer" tests execute and pass on my linux and skips them on my mingw


JanWielemaker:
Thanks.   Only, there was a closing parenthesis missing :cry:  Should have pulled and tested before merging ...   Anyway, fixing issues by removing code always has my preference :smile:

PR #975, opened 2022-04-01
PORT: Msys2/clang64

mgondan:
On clang64, libgcc_s_seh-1.dll is not found, move on with a message. On the long run, these dlls should be collected not during build but from cpack.


JanWielemaker:
Thanks.  Merged with #974 and pushed.

PR #977, opened 2022-04-03
DOC: Added to "foreign" section

kamahen:
Added regression test for non-deterministic code

This PR is incomplete - I don't know what to put into `src/Tests/foreign2/CMakeLists.txt` - and it might also be in the wrong place.
(I ran the tests in `src/Tests/foreign2` by hand.)

### DIFF
--- a/man/foreign.doc
+++ b/man/foreign.doc
@@ -510,7 +510,9 @@ my_function(term_t a0, term_t a1, control_t handle)
 
   switch( PL_foreign_control(handle) )
   { case PL_FIRST_CALL:
-        ctxt = malloc(sizeof(struct context));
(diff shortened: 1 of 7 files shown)


JanWielemaker:

https://swi-prolog.discourse.group/t/pl-retry-and-non-positive-values/5198/10

kamahen:
As sugested, I'll remove the foreign-function tests and move them to [packages-cpp](https://github.com/SWI-Prolog/packages-cpp)

kamahen:
This PR is messed up because I accidentally included submodules (memo to self: don't use "git commit -a". So, I'm closing it and will create a new PR.

PR #979, opened 2022-04-07
PORT: skip dll in Rtools (only static libs)

mgondan:


JanWielemaker:
Thanks.  Merged after tab/spaces fix

PR #983, opened 2022-04-17
DOC: Added details to "foreign" section for blobs and streams

kamahen:
I used \program{...} in some changes because \ctype{...} seemed wrong and I couldn't figure out how to add another command to pl.sty (e.g., \cfuncname{...}) that also expanded the HTML.

Also, there's a good chance I got some things wrong in what I wrote, even though so of it is copy-and-paste from what Jan had written to me in an email.

### DIFF
--- a/man/foreign.doc
+++ b/man/foreign.doc
@@ -318,6 +318,8 @@ Atoms are the central representation for textual constants in Prolog.
 The transformation of a character string $C$ to an atom implies a
 hash-table lookup.  If the same atom is needed often, it is advised
 to store its reference in a global variable to avoid repeated lookup.
+
+Blobs are also represented as atoms (see \ref{sec:blob}).
     \item[functor_t]
 A functor is the internal representation of a name/arity pair. They are
 used to find the name and arity of a compound term as well as to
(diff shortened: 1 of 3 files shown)


JanWielemaker:
Merged after serious editing.   Some of your changes hinted at old, not properly documented features.   I've tried to clarify most of those and stick with the somewhat more formal documentation style as opposed to some of the snippets you took from Discourse.   I have removed the start of the section on streams.   Yes, it would be great to have a proper documentation on stream handling.  I considered this too immature for the docs as it is though.

I'm aware this is some serious editing.   I hope it moves into the right direction and clarifies some issues.   At some point the entire documentation of the foreign interface needs to be revisited and notably a lot of the atom related functions need to be deprecated.   Many of them still assume the old world where the content of an atom was a simple 0-terminated C string.   Since then we got length+string, UCS atoms (USC-4 or UCS-2 on Windows) and blobs.

kamahen:
Thank-you, @JanWielemaker , for taking the time to fix my my misunderstandings.

JanWielemaker:
> for taking the time to fix my my misunderstandings.

Together we get to something that is correct __and__ understandable :smile:

PR #984, opened 2022-04-20
Fix typos in foreign language API docs

kamahen:
I've also added a few sentences, which need to be verified.


JanWielemaker:
Thanks!

PR #985, opened 2022-04-24
Replace MINGW_ROOT by WIN32

mgondan:
I think -municode needs to be defined on all Windows systems that are currently supported (= several flavors of MinGW). MINGW_ROOT was not defined on my system, so I would like to change this to WIN32 to save a cmake command line switch.


mgondan:
The thing is that in swipl-devel, MINGW_ROOT is used to help the cmake process collect MinGW-specific files during a build. Here, it has a different purpose; here it is used to add a Windows-specific compiler switch. Now that I think about it, it may be considered MinGW-specific, so instead of "if(MINGW_ROOT)" we should have "if(MINGW)" [not if(WIN32)]. Tested on the RTools42.

JanWielemaker:
Merged (squashed).

PR #986, opened 2022-04-27
Wrap elementary functions to correct issues with rounding mode support in various libm's

ridgeworks:
Various platform libm's produce incorrect results when used with IEEE float rounding modes other than 'to_nearest'. This PR wraps these functions with additional code to ensure rounding mode correctness. See comment in 'pl-arith.c' for further details. A few additional test cases are included to detect incorrect rounding, but these are by no means exhaustive. Also see <https://swi-prolog.discourse.group/t/rounding-transcendentals-broken-on-macos-intel/5166> for lengthy discussion preceding this PR.

Note that this fix does not address any precision issues in the various libraries. A proven implementation, such as 'crlibm', is required to address this issue.

### DIFF
--- a/src/Tests/rational/test_ieee754.pl
+++ b/src/Tests/rational/test_ieee754.pl
@@ -538,7 +538,10 @@ test(ieee_rndto) :-
     assertion(test_roundto(-2.0** 1r3)),
     assertion(test_roundto( 2.0** 1r3)),
     assertion(test_roundto(pi)),
-    assertion(test_roundto(e)).
+    assertion(test_roundto(e)),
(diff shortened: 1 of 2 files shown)


JanWielemaker:

https://swi-prolog.discourse.group/t/rounding-transcendentals-broken-on-macos-intel/5166/54

JanWielemaker:
I've merged this after syncing the layout and adding documentation.   It is not ideal, but it seems the  most pragmatic way to resolve this.   Thanks!   If you have further patches, please start a new PR on master.

PR #989, opened 2022-05-10
Fixes for arithmetic function ar_pow() - float**rat case

ridgeworks:
Recent change for correctly rounded elementary function exposed weaknesses in ar_pow() when raising a float to a rational number. This PR largely replaces the previous code in pl_arith.c.

Additional test cases have been added to test the changes and some have been moved from test_rational.pl to test_ieee754.pl  to put all the tests in one place.

Also replaced the pow() calls (non-corrected) in pl_gmp.c (function fpz_fdiv) with calls to ldexp() which seem to be intended for the required functionality. Note that this function is commonly used when promoting rationals to floats, but running ctest does not detect any issues.

I believe these changes are fairly localized and the risk of applying them should be minimal.

### DIFF
--- a/src/Tests/rational/test_ieee754.pl
+++ b/src/Tests/rational/test_ieee754.pl
@@ -528,6 +528,18 @@ test(ieee_rmode) :-
     assertion(test_rounding(pi)),
     assertion(test_rounding(e)).
 
+% rational exponent power tests (3)
+test(float_to_rat) :-
+    assertion(test_value(8.0 is 4.0**3r2)),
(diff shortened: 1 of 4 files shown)


JanWielemaker:
Thanks.   Merged after some layout cleanup, removal of some trivial comments and rewording the commit messages.    Stating that a file is updated in a commit is meaningless ...

If you have the energy, I'd be happy to see ar_pow() split into a number of functions.   As is, it becomes really hard to read.   Note that modern C compilers inline a static function that is  used only in one place anyway, so there is no performance impact in spitting.

PR #991, opened 2022-05-12
Add support for conditional breakpoints

eshelyaron:
This PR implements (and documents) a new API for creating *conditional breakpoints*, which are breakpoints that invoke an arbitrary user-provided condition goal to decide whether turn on trace mode.

### DIFF
--- a/library/prolog_breakpoints.pl
+++ b/library/prolog_breakpoints.pl
@@ -38,7 +38,7 @@
 :- module(prolog_breakpoints,
           [ set_breakpoint/4,           % +File, +Line, +CharPos, -Id
             set_breakpoint/5,           % +Owner, +File, +Line, +CharPos, -Id
-            set_breakpoint/6,           % +Owner, +File, +Line, +CharPos, +Cond, -Id
+            set_breakpoint/6,           % +Owner, +File, +Line, +CharPos, :Cond, -Id
             delete_breakpoint/1,        % +Id
             breakpoint_property/2       % ?Id, ?Property
           ]).
@@ -350,5 +350,5 @@ prolog:break_hook(Clause, PC, Frame, _Choice, _Goal, true, Action) :-
                        ->  Action = trace
                        ;   Action = continue
                        ),
-                       prolog:erase(Ref)).
+                       erase(Ref)).
 
--- a/library/prolog_breakpoints.pl
+++ b/library/prolog_breakpoints.pl
@@ -38,8 +38,7 @@
 :- module(prolog_breakpoints,
           [ set_breakpoint/4,           % +File, +Line, +CharPos, -Id
             set_breakpoint/5,           % +Owner, +File, +Line, +CharPos, -Id
-            set_breakpoint/6,           % +Owner, +File, +Line, +CharPos, :Cond, -Id
-            set_breakpoint_condition/2, % +Id, :Cond
+            set_breakpoint_condition/2, % +Id, +Cond
             delete_breakpoint/1,        % +Id
             breakpoint_property/2       % ?Id, ?Property
           ]).
@@ -63,21 +62,12 @@ Note that the hook must fail  after   creating  its side-effects to give
 other hooks the opportunity to react.
 */
 
-:- meta_predicate set_breakpoint(+, +, +, +, 0, -).
-
 %!  set_breakpoint(+File, +Line, +Char, -Id) is det.
 %!  set_breakpoint(+Owner, +File, +Line, +Char, -Id) is det.
-%!  set_breakpoint(+Owner, +File, +Line, +Char, :Cond, -Id) is det.
 %
 %   Put a breakpoint at the indicated source-location. File is a current
 %   sourcefile (as reported by source_file/1). Line  is the 1-based line
 %   in which Char is. Char is the position of the break.
-%
-%   Cond is a goal that will be invoked whenever the newly set
-%   breakpoint is triggered. If goal fails, the breakpoint is skipped
-%   and execution commences normally.  In the context of Cond,
-%   break_context_frame/1 is made available for obtaining the
-%   execution frame identifier in which the breakpoint was triggered.
 %
 %   First, '$clause_from_source'/4 uses  the   SWI-Prolog  clause-source
 %   information  to  find  the  last    clause   starting  before  Line.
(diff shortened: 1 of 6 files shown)


library/prolog_breakpoints.pl
```diff
@@ -61,18 +61,25 @@
 */
 
 %!  set_breakpoint(+File, +Line, +Char, -Id) is det.
+%!  set_breakpoint(+File, +Line, +Char, :Cond, -Id) is det.
```
kamahen, line 64: Does this need a meta_predicate directive?

library/prolog_breakpoints.pl
```diff
@@ -61,18 +61,25 @@
...
 %!  set_breakpoint(+Owner, +File, +Line, +Char, -Id) is det.
 %
 %   Put a breakpoint at the indicated source-location. File is a current
 %   sourcefile (as reported by source_file/1). Line  is the 1-based line
 %   in which Char is. Char is the position of the break.
 %
+%   Cond is a goal that will be invoked whenever the newly set
+%   breakpoint is triggered. If goal fails, the breakpoint is skipped
+%   and execution commences normaly.  In the context of Cond,
```
kamahen, line 73: normaly => normally

library/prolog_breakpoints.pl
eshelyaron: Added in https://github.com/SWI-Prolog/swipl-devel/pull/991/commits/03a3e5fb542b44d37d583be72c10c1c9736adc90

library/prolog_breakpoints.pl
eshelyaron: Thanks, fixed in https://github.com/SWI-Prolog/swipl-devel/pull/991/commits/03a3e5fb542b44d37d583be72c10c1c9736adc90

library/prolog_breakpoints.pl
```diff
@@ -60,19 +61,28 @@
 other hooks the opportunity to react.
 */
 
+:- meta_predicate set_breakpoint(+, +, +, +, 0, -).
+
 %!  set_breakpoint(+File, +Line, +Char, -Id) is det.
 %!  set_breakpoint(+Owner, +File, +Line, +Char, -Id) is det.
+%!  set_breakpoint(+Owner, +File, +Line, +Char, :Cond, -Id) is det.
```
JanWielemaker, line 68: I had expected an interface set_breakpoint_condition(+Id, :Cond).   That would make the human interface easier: set a breakpoint by in the gui, then select the breakpoint and add the condition.   It would also facilitate changing the condition if you discover that the breakpoint is still called too often.

eshelyaron: I see, that makes sense, I've added `set_breakpoint_condition/2` in https://github.com/SWI-Prolog/swipl-devel/pull/991/commits/8946aa5de2417e20446006f2fd1d3d209d61e17e
(Also rebased to track the latest commits in the master branch, hence the force-push)

JanWielemaker:
Thanks.  That looks good.    Small stuff is that the commit tags are `ADDED` and `MODIFIED` rather than `ADD` and `MODIFY`. 

When you showed this stuff to me, you could reuse variable names in the condition.   I'd say that is almost a must to make this really useful.    I guess this would require the set_breakpoint_condition/2 to get an options argument holding the read_term/3 option variable_names(Bindings).    It also requires some more work in the hook.   As you seem to have all that ...  can we have it too?

And probably you want to evaluate the condition using catch/3.   What would be the best option?  

   - print and trace, so the user can debug the buggy condition?
   - print and continue
   - keep silent and continue (just handle them as failed).

eshelyaron:
> When you showed this stuff to me, you could reuse variable names in the condition. I'd say that is almost a must to make this really useful. I guess this would require the set_breakpoint_condition/2 to get an options argument holding the read_term/3 option variable_names(Bindings). It also requires some more work in the hook. As you seem to have all that ... can we have it too?

Alright, I wasn't sure if that feature was essential, but I'll definitely add it then.
One alternative interface to adding a `variable_names` option that comes to mind, is to provide a convenient `name_variable(+Name:atom, -Value:term)` (or something like that) that will be available inside the condition goal for inspecting the value of variables given their source name.
So setting a condition could look like this:
```
set_breakpoint_condition(1, name_variable('X', foo))
```
Instead of the suggested:
```
set_breakpoint_condition(1, X=foo, [variable_names('A'=X)])
```
This would make the condition goal term a little less "magical" and it doesn't require to use of a new variable in the call site of `set_breakpoint_condition` for every source variable that the user wants to inspect (like `X` is used to inspect the value of `A` above). 
Not sure if it's the more intuitive interface though.
What do you think?

JanWielemaker:
> What do you think?

I think the practical interface will take a condition as a string and a breakpoint id.    In this respect I do not see too much value in the set_breakpoint/6.  I would argue for just set_breakpoint/5 and a set_breakpoint_condition/2 (or /3) as well as disconnecting the breakpoint condition from the Prolog (fact) clause that describes the breakpoint.   Both for IDEs and probably also for the toplevel this is most convenenant:

     ?- set_breakpoint_condition(1, "var(X)").

JanWielemaker:
P.s.  Pushed fixes that makes syncing the deletion of a breakpoint due to reloading a source file work again.   Conflicts in onbreak, should be trivial to fix.

eshelyaron:
@jan thanks for the pointers, I've pushed an update to address your comments:
> I think the practical interface will take a condition as a string and a breakpoint id.

Done, now the examples says:
```
?- set_breakpoint('/...path.../Example.pl', 8, 24, ID).
ID = 1.

?- set_breakpoint_condition(1, "X == rock2").
true.
```
And `X` is unified with the variable that corresponds to X in the source clause as expected:
```
[debug]  ?- test_noun2(X, rock).
X = rock1 ;
X = rock2.

[debug]  ?- test_noun2(rock2, rock).
   Call: (11) noun(rock2, rock) ? creep
   Call: (12) is_a(rock2, rock) ? creep
   Exit: (12) is_a(rock2, rock) ? creep
   Exit: (11) noun(rock2, rock) ? creep
   Exit: (10) test_noun2(rock2, rock) ? creep
true.
```

> disconnecting the breakpoint condition from the Prolog (fact) clause that describes the breakpoint.

Alright, I've added a separate dynamic `breakpoint_condition(Id, CondTerm)` instead.
The `CondTerm` is `Goal-OffsetVarPairs` where `OffsetVarPairs` is a list of `Offset-Var` terms, where `Var` is a variable that is shared with `Goal` which is to be bound with the stack frame value at `Offset`.

>   Pushed fixes... Conflicts in onbreak, should be trivial to fix.

No problem, rebased.

eshelyaron:
These changes were incorporated into the master branch in https://github.com/SWI-Prolog/swipl-devel/commit/5d335612c24a9ca7c54566f97a62cbd613297db3 and https://github.com/SWI-Prolog/swipl-devel/commit/af63d7b648ee527a50894fc57560171f6e2496f8, closing.

PR #994, opened 2022-06-01
FIXED: write_term/2 with numbervars and variable_names (#866)

### LINKED ISSUE #866: write_term/2 with numbervars and variable_names instantiates its argument
Not totally sure if this is a bug or works-as-intended, but if the latter then the documentation should probably be make the pitfall clearer.

If the `numbervars(true)` and `variable_names(...)` options are both given to write_term, the bindings made by `variables_names` aren't undone on completion of the call. This doesn't happen when `numbervars` is _not_ specified. So:

```prolog
?- Term = foo(X,Y,X,Z),
   Opts = [variable_names(['XName'=X]), nl],
   CheckX = ((var(X) -> write_canonical(var(X)) ; write_canonical(nonvar(X))), nl),
   nl,
   CheckX,
   write_term(Term, Opts),
   CheckX,
   write_term(Term, [numbervars|Opts]),
   CheckX,
   nl.

var(_)
foo(XName,_122326,XName,_122330)
var(_)
foo(XName,_122326,XName,_122330)
nonvar('$VAR'('XName'))

Term = foo(XName, Y, XName, Z),
X = XName,
Opts = [variable_names(['XName'=XName]), nl],
Check =  ((var(XName)->write_canonical(var(XName));write_canonical(nonvar(XName))), nl).
```

eshelyaron:
This PR fixes #866.
Note: I tried changing `writeNumberVar` as little as possible seeing as `write_canonical` relies on it under the hood, but I suspect these areas could be simplified a bit otherwise.

### DIFF
--- a/src/Tests/core/test_write.pl
+++ b/src/Tests/core/test_write.pl
@@ -181,6 +181,22 @@ test(variable_names, X = 'a(A,B)') :-
 	    atom(X),
 	    write_term(a('$VAR'(0), A),
 		       [variable_names(['B'=A]), numbervars(true)])).
+test(variable_names, X = 'a($VAR(0),B)') :-
+	with_output_to(
+	    atom(X),
+	    write_term(a('$VAR'(0), A),
+		       [variable_names(['B'=A])])),
+        A = 2.
+test(variable_names, X = 'a(A,B)') :-
+	with_output_to(
+	    atom(X),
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks.  Merged after avoiding an uninitialized variable warning (false alarm)

PR #996, opened 2022-06-18
PORT: win_add_dll_directory info on relative path

mgondan:


JanWielemaker:
Why is this needed?    I'm also not too happy with normal predicates printing messages.   For most stuff except for some things you only use during debugging, I think silent is better.

mgondan:
I explained it a bit in my comment in the forum. According to the docs, the argument to win_add_dll_directory/1 should be an absolute path (AbsDir), but it silently accepts a relative path, as well. I think it is fine to issue a message in such a case, whether it is informational or silent is probably a matter of taste, so I don't mind if you change it to silent (but see the use case below). Given that it's technically wrong to feed win_add_dll_directory/1 with a relative path, I am slightly in favor of informational.

The use case: I used win_add_dll_directory('/mingw64/bin') on MSYS2, which in MSYS2 bash, is correctly expanded to c:/msys64/mingw64/bin because c:/msy64 is mounted on /. From within swipl, /mingw64/bin is expanded to c:/mingw64/bin, which is technically correct (but useless), and it would probably neither possible nor desirable to make swipl aware of mount points etc. Yesterday, it would have saved me a lot of time though yesterday if swipl had informed me about the result of the path expansion, therefore my PR.

PR #997, opened 2022-06-19
ENHANCE: win_add_dll_directory/1 warn if Dir does not exist

mgondan:
Superseeds the other PR. 
* silent message if win_add_dll_directory(AbsDir) is invoked with relative Dir
* error if AbsDir does not exist

### DIFF
--- a/library/shlib.pl
+++ b/library/shlib.pl
@@ -552,17 +552,35 @@ unload_foreign(File) :-
 %
 %   @error domain_error(operating_system, windows) if the current OS
 %   is not Windows.
+%   @error existence_error(directory, AbsDir) if AbsDir does not
+%   exist.
 
 win_add_dll_directory(Dir) :-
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Considering this seems to be about reporting, I've pushed some updates that should make this far more consistent with normal practices.   win_add_dll_directory/2 now throws normal Prolog exceptions instead of some window system call exception.  win_add_dll_directory/1 no longer does the catch/3 stuff, so it propagates the normal exception.   This predicate is now also only defined on Windows systems, like all the other `win_*` predicates.    This required some more work on the autoload index generation to deal with conditional exports (was on the wish list much longer).

Hope this solves your problems.    Not sure where we should add the MSYS2 DLL directory.    I guess either generally in the initialization or if it only affects loading plugins, in `shlib`.

PR #998, opened 2022-06-27
PORT: MSYS2, add $MINGW_PREFIX/bin to dll search path

mgondan:
If swipl is invoked from the bash under MSYS2, /mingw64/bin is not on the dll search path, so that the dlls such as zlib1.dll are not found. Here I check if $MINGW_PREFIX is defined, and if it is $MINGW_PREFIX/bin is added to the dll search path.

### DIFF
--- a/boot/build_home.pl
+++ b/boot/build_home.pl
@@ -208,6 +208,15 @@ add_package_path(PkgBinDir) :-
           add_package(Pkg, PkgBinDir)).
 :- endif.
 
+% Add %MINGW_PREFIX%/bin to dll search path under MSYS2
+:- if(current_prolog_flag(windows, true)).
+:- (   getenv('MINGW_PREFIX', Prefix)
+   ->  format(atom(Bin), '~w/bin', [Prefix]),
+       win_add_dll_directory(Bin, _)
+   ;   true
+   ).
+:- endif.
+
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Looks ok.   Just, this file is only use during build and when running src/swipl.exe from the build tree.   Is there no problem with the installed system?   If so, the fix should be elsewhere.    If it only affects loading dlls dynamically, library(shlib) is probably the most sensible place.

mgondan:
Thank you. Then it's in the wrong place. I actually wanted to ask you first, but I have run out of "responses" on discourse. One is not allowed to add responses to a thread more than three times in a row :-) I'll close this PR and submit a new one instead.

mgondan:
Of course. It wasn’t really a problem.



Von: Jan ***@***.***>
Gesendet: Dienstag, 28. Juni 2022 09:47
An: ***@***.***>
Cc: Gondan-Rochon, ***@***.***>; State ***@***.***>
Betreff: Re: [SWI-Prolog/swipl-devel] PORT: MSYS2, add $MINGW_PREFIX/bin to dll search path (PR #998)


but I have run out of "responses" on discourse.

Hmm. Didn't know there was a hard limit. You can always use a PM or send normal mail.

PR #1002, opened 2022-06-29
PORT: Windows, win_add_dll_directory/2 on startup

mgondan:


JanWielemaker:
Thanks.   Please do not add to PRs after they have been merged.    Examined the situation by hand and decided for a little cleanup.  See 7564d2b610cb9f442756e273a0ed93561c41d772.  Two blocks with the same condition is unnecessary and after that change we can just as well keep using win_add_dll_directory/1.   Also turned it into an initialization/1 call as this is probably needed to make this work in a saved state.

mgondan:
Thank you. Will improve my git skills...

PR #1005, opened 2022-07-01
FIXED: pack_install/2 upgrade version check

kamahen:


JanWielemaker:

https://swi-prolog.discourse.group/t/pack-rocksdb-build-fails-on-ubuntu-22-04-lts-on-wsl-on-windows-reproducible/5559/21

JanWielemaker:
Thanks.   Fixed with a338fa38263b623863c6545753097e6343bc00f7.   Your fix is incorrect because we cannot compare versions using lexical ordering, e.g., 10.0 is lexically before 9.0.

PR #1008, opened 2022-07-14
FIXED: user_or_explicit autoload mode not working

DRKV333:
Currently setting `autoload` to `user_or_explicit` is actually equivalent to `false`:
```
1 ?- set_prolog_flag(autoload, user_or_explicit).
% Disabled autoloading (loaded 17 files)
true.

2 ?- member(X,[1,2,3]).
Correct to: "lists:member(X,[1,2,3])"? no
ERROR: Unknown procedure: member/2
```
This is because autoload is looking for `explicit_or_user`, instead of `user_or_explicit`.


JanWielemaker:
Thanks!

PR #1010, opened 2022-07-16
PORT: Suppress uninit warning in msys2

mgondan:
I think rc can safely be initialized to FALSE


JanWielemaker:
Thanks.  I'm not a big fan of unnecessary initializations as they hide that some control paths indeed do not set the variable 😢

mgondan:
I know, and that was indeed a borderline case. There'll be another PR in a few minutes, hopefully less controversial.

PR #1016, opened 2022-07-27
Use PL_BLOB_WCHAR instead of test for ==&ucs_atom

kamahen:
These changes allow user-defined atom-like blobs. I mainly did this work for some tests (which haven't been finished); if you don't like the changes, feel free to drop this PR.
There are a few places where I wasn't sure if I could make a change safely, to I left a `TODO`


JanWielemaker:

https://swi-prolog.discourse.group/t/changes-to-swi-cpp-h/5601/16

JanWielemaker:
Merged the documentation changes after some additional editing, notably to make it more clear that users should not use the blob API to deal with normal atoms.    Handling text atoms is hard-coded at several places, notably compareAtoms() and callees.

PR #1018, opened 2022-07-28
Performance enhancements to library(simplex) and library(lists)

ridgeworks:
Performance only - should not result in any regressions.

### DIFF
--- a/library/clp/simplex.pl
+++ b/library/clp/simplex.pl
@@ -919,7 +919,7 @@ linsum_negate(Coeff0*Var, Coeff*Var) :- Coeff is -Coeff0.
 
 linsum_row([], _, []).
 linsum_row([V|Vs], Ls, [C|Cs]) :-
-        (   member(Coeff*V, Ls) -> C is rationalize(Coeff)
+        (   memberchk(Coeff*V, Ls) -> C is rationalize(Coeff)
         ;   C = 0
         ),
         linsum_row(Vs, Ls, Cs).
@@ -1044,15 +1044,15 @@ gauss_elimination([Row0|Rows0], PivotRow, Col, [Row|Rows]) :-
 
 row_eliminate(row(Var, Ls0, R0), row(_, PLs, PR), Col, row(Var, Ls, R)) :-
         nth0(Col, Ls0, Coeff),
-        (   Coeff =:= 0 -> Ls = Ls0, R = R0
-        ;   MCoeff is -Coeff,
-            all_times_plus([PR|PLs], MCoeff, [R0|Ls0], [R|Ls])
+        (   Coeff = 0 -> Ls = Ls0, R = R0  % assumes rational Coeff
+        ;   all_times_minus([PR|PLs], Coeff, [R0|Ls0], [R|Ls])
         ).
 
-all_times_plus([], _, _, []).
-all_times_plus([A|As], T, [B|Bs], [AT|ATs]) :-
-        AT is A * T + B,
-        all_times_plus(As, T, Bs, ATs).
+all_times_minus([], _, _, []).
+all_times_minus([A|As], T, [B|Bs], [AT|ATs]) :-
+        % optimization for sparse matrix assuming A is rational
+        (A = 0 -> AT = B ; AT is B - A * T),
+        all_times_minus(As, T, Bs, ATs).
 
 all_times([], _, []).
 all_times([A|As], T, [AT|ATs]) :-
--- a/library/lists.pl
+++ b/library/lists.pl
@@ -292,7 +292,8 @@ nth_gen([H|Tail], Elem, _, N, Base) :-
 nth1(Index, List, Elem) :-
     (   integer(Index)
     ->  Index0 is Index - 1,
-        nth0_det(Index0, List, Elem)        % take nth deterministically
+        '$seek_list'(Index0, List, RestIndex, RestList),
+        nth0_det(RestIndex, RestList, Elem) % take nth det
     ;   var(Index)
     ->  List = [H|T],
         nth_gen(T, Elem, H, 1, Index)       % match


library/clp/simplex.pl
```diff
@@ -1044,15 +1044,15 @@
 
 row_eliminate(row(Var, Ls0, R0), row(_, PLs, PR), Col, row(Var, Ls, R)) :-
         nth0(Col, Ls0, Coeff),
-        (   Coeff =:= 0 -> Ls = Ls0, R = R0
-        ;   MCoeff is -Coeff,
-            all_times_plus([PR|PLs], MCoeff, [R0|Ls0], [R|Ls])
+        (   Coeff = 0 -> Ls = Ls0, R = R0  % assumes rational Coeff
```
JanWielemaker, line 1047: Do you have any proof this helps?  I tried the code below using `-O` (or setting the Prolog flag `optimise` to `true` before compiling).   The timing for 10 000 000 iterations are 0.192, 0.201 and 0.153.  That is more or less in line with what I expected.   =/2 is faster than =:=/2, but as a condition inside if-then-else Prolog can turn =:=/2 into a fast switch because no variables are instantiated.  Using ==/2 is best as it both uses the simplified branch and the test is simpler.

```
l0(N) :-
    (   N =:= 0
    ->  true
    ;   N2 is N - 1,
        l0(N2)
    ).

l1(N) :-
    (   N = 0
    ->  true
    ;   N2 is N - 1,
        l1(N2)
    ).

l2(N) :-
    (   N == 0
    ->  true
    ;   N2 is N - 1,
        l2(N2)
    ).
```

library/lists.pl
```diff
@@ -292,7 +292,8 @@
 nth1(Index, List, Elem) :-
     (   integer(Index)
     ->  Index0 is Index - 1,
-        nth0_det(Index0, List, Elem)        % take nth deterministically
```
JanWielemaker, line 295: I tried to make a benchmark to demonstrate the gain, but it falls into the noise (< 1%).   Given that it makes the code more ugly I think this is not a good idea.

JanWielemaker (commented):
Only the memberchk/2 is probably worthwhile.   If your findings are otherwise, please share the benchmark.

library/lists.pl
ridgeworks: My defining benchmark is the Stigler diet problem from the Google OR tools site previously mentioned on the `nth0/3` optimization thread. It defines 9 constraints in 77 variables so a matrix of 9 rows and 77 columns. It is a sparse matrix - any row will typically have only a few non-zero coefficients but I think sparse matrices will be common in problems of this size. Using `time` and `profile`

SWIP 8.4.15 with current simplex:        ~1.4 secs. with 89% of the time (~1.2 secs.) spent in `row_eliminate`
SWIP 8.4.15 with simplex as modified: ~0.39 secs. with 63% of the time (~0.25 secs.) spent in `row_eliminate`

So on this particular benchmark, the improvement in `row_eliminate` is almost a factor of 5. The `memberchk` mod has a small impact in comparison.

Note that simplex uses rational arithmetic so I'm not sure tests using integers is a good comparison; you probably have a better feel for that than I do. 


RegRegarding `nth1` , currently (8.5.15):
```
?- numlist(1,50,L), time((between(1,1000000,_),nth0(49,L,X),fail)).
% 4,000,001 inferences, 0.457 CPU in 0.457 seconds (100% CPU, 8761729 Lips)
false.

?- numlist(1,50,L), time((between(1,1000000,_),nth1(49,L,X),fail)).
% 99,000,001 inferences, 6.355 CPU in 6.356 seconds (100% CPU, 15578354 Lips)
false.
```
With modified version of `nth1`:
```
?- numlist(1,50,L), time((between(1,1000000,_),nth1(49,L,X),fail)).
% 4,000,004 inferences, 0.471 CPU in 0.471 seconds (100% CPU, 8493822 Lips)
false.
```
I would say it's a lot more than 1% for an index of 49.

JanWielemaker: Thanks.   I was misreading and thought the nt1 patch only avoided an indirection through nth0.   Confirm and merged.    Also merged the simplex.pl test, but using =:= in row_eliminate (faster and consistent) and == in all_times_minus (faster and if the test fails on something that is true under =:=, the result is still the same.

Could you share the program?   If I see this correctly, there is only reference to the google site, but no Prolog code I assume?    Would make a good case in the test and/or benchmark set.

JanWielemaker:
Merged.  See review comments.

library/lists.pl
ridgeworks: Thanks. Can you expand a bit on the differences between `=`, `==`, and `=:=` for this use case, i.e, comparing a rational to 0. As currently implemented:
```
row_eliminate(row(Var, Ls0, R0), row(_, PLs, PR), Col, row(Var, Ls, R)) :-
        nth0(Col, Ls0, Coeff),
        (   Coeff =:= 0 -> Ls = Ls0, R = R0
        ;   all_times_minus([PR|PLs], Coeff, [R0|Ls0], [R|Ls])
        ).

all_times_minus([], _, _, []).
all_times_minus([A|As], T, [B|Bs], [AT|ATs]) :-
        (A == 0
          -> AT = B 
          ;  AT is B - A * T
        ),
        all_times_minus(As, T, Bs, ATs).
```
As measured, `==` seems to be faster than `=:=` in the second instance (as expected), so why isn't it preferable in the first? Is `==` generally faster than `=` for ground comparisons?

My current Stigler program uses a clpBNR wrapper around simplex, but it shouldn't be too hard to extract a version than uses simplex directly. I'll let post it here when I have something useful.

JanWielemaker: =/2 is unification.   This has, between two atomic values, about the same time characteristics as ==/2.   However, the system must be prepared for =/2 to have partial side effects, may cause stack shifts/garbage collection, etc. (part of the unification succeeds, later it fails or the constraint evaluation fails, etc.)   As a result, when used as a condition in -> we must create a choice point and save enough context to recover and go into the _else_ branch.   ==/2 doesn't have all this.   In addition, as we know the target is a simple constant (atom or small integer), == is handled by a dedicated VM instruction that needs to compare two simple handles to figure out it is true or false.   =:=/2 is arithmetic, so it needs to evaluate both sides, promote them to the same types if the types differ and then finally perform the comparison.    Arithmetic doesn't touch the stacks though, so no shift/GC, constraints, etc. and thus -> doesn't have to create a choice point.

As for which to use here, it is a bit unclear.   I thought the first might affect the result, but it does not.   =:=/2 succeeds in more cases (also for 0.0 =:= 0).   As not taking the fast route when this is possible notably affects the first case (I'm not even sure it helps in the second, did you check (with -O)?) I think this is the best choice.

ridgeworks: Good to know about the implementation issues around '=', so '==' definitely favoured in testing if a rational equals 0. But it is somewhat specific to this use case. Note:
```
?- X is 1/42, X == 1/42.
false.
```
I assume this is because there are two instances of the rational `1/42` so they're not "identical".  However extended integers (bigger than `max_tagged_integer`) and strings don't seem to have the same issue:
```
?- X = 99972057594037927935, X==99972057594037927935.
X = 99972057594037927935.

?- X="abc", X=="abc".
X = "abc".
```
Bug or feature?


Also by my tests `==` is always better than `=:=` when testing if a rational is 0, particularly when the test fails:
```
?- X is 0/42, time((between(1,1000000,_),X == 0,fail)).
% 2,000,002 inferences, 0.097 CPU in 0.097 seconds (100% CPU, 20562613 Lips)
false.

?- X is 0/42, time((between(1,1000000,_),X =:= 0,fail)).
% 2,000,001 inferences, 0.135 CPU in 0.135 seconds (100% CPU, 14858407 Lips)
false.

?- X is 1/42, time((between(1,1000000,_),X == 0,fail)).
% 2,000,001 inferences, 0.110 CPU in 0.111 seconds (99% CPU, 18186622 Lips)
false.

?- X is 1/42, time((between(1,1000000,_),X =:= 0,fail)).
% 2,000,001 inferences, 0.409 CPU in 0.410 seconds (100% CPU, 4885833 Lips)
false.
```
In the Stigler benchmark overall, I can see little measurable difference between `==` and `=:=` for the first usage. (Definitely makes a difference for the second which is executed much more frequently for larger number of columns.) However, I don't see why you wouldn't use `==` for the first as well; there are no floats involved so there is only one representation of zero involved. But the current committed version is fine and the discussion has been useful.

JanWielemaker: > I assume this is because there are two instances of the rational 1/42 so they're not "identical".

No.  Prolog == compares values, not instances.    The problem is that both evaluation and reading of 1/42 depends on Prolog flags.    `prefer_rationals` controls evaluation.   `rational_syntax` controls whether `1/42` is read as `/(1,42)` or `1r42`.  If you use both flags, the toplevel goals succeeds.

P.s. for timing these things you should write predicates and use `-O`.   Some of these constructs are optimized when compiled, but not when doing meta-calling because the cost of optimization easily exceeds the cost of one time usage.

ridgeworks:
My defining benchmark is the Stigler diet problem from the Google OR tools site previously mentioned on the 'nth0/3' optimization thread. It defines 9 constraints in 77 variables so a matrix of 9 rows and 77 columns. It is a sparse matrix - any row will typically have only a few non-zero coefficients but I think sparse matrices will be common in problems of this size. Using 'time' and 'profile'

SWIP 8.4.15 with current simplex:        ~1.4 secs. with 89% of the time (~1.2 secs.) spent in 'row_eliminate'
SWIP 8.4.15 with simplex as modified: ~0.39 secs. with 63% of the time (~0.25 secs.) spent in 'row_eliminate'

So on this particular benchmark, the improvement in 'row_eliminate' is almost a factor of 5. The 'memberchk' mod has a small impact in comparison.

Note that simplex uses rational arithmetic so I'm not sure tests using integers is a good comparison; you probably have a better feel for that than I do. 


RegRegarding 'nth1ding'nth1' , currently (8.5.15):

?- numlist(1,50,L), time((between(1,1000000,_),nth0(49,L,X),fail)).
% 4,000,001 inferences, 0.457 CPU in 0.457 seconds (100% CPU, 8761729 Lips)
false.

?- numlist(1,50,L), time((between(1,1000000,_),nth1(49,L,X),fail)).
% 99,000,001 inferences, 6.355 CPU in 6.356 seconds (100% CPU, 15578354 Lips)
false.

With modified version of 'nth1':

?- numlist(1,50,L), time((between(1,1000000,_),nth1(49,L,X),fail)).
% 4,000,004 inferences, 0.471 CPU in 0.471 seconds (100% CPU, 8493822 Lips)
false.

I would say it's a lot more than 1% for an index of 49.

PR #1029, opened 2022-08-31
ENHANCED: fallback to the default allocation functions in the GMP hooks

eshelyaron:
This PR makes the GMP memory hooks detect that they are invoked out of the context of Prolog and fall back to the default allocation functions.

### DIFF
--- a/src/pl-gmp.c
+++ b/src/pl-gmp.c
@@ -92,7 +92,7 @@ mp_alloc(size_t bytes)
 { GET_LD
   mp_mem_header *mem;
 
-  if ( LD->gmp.persistent )
+  if ( LD == NULL || LD->gmp.context == NULL || LD->gmp.persistent )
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.   Reorganized a little to (1) put the condition in a macro, so we can name it and share it between the three use cases and (2) call the saved malloc/realloc/free routines, so it should also work in environments that also redefine these hooks.   Hope it still works with Emacs.

eshelyaron:
> Hope it still works with Emacs.

Works great, thanks.
I've [updated `sweep`](https://git.sr.ht/~eshel/sweep/commit/66ac977debda8c10fa008c19d7ae584b5978c49a) accordingly not to rely on `PL_action(PL_GMP_SET_ALLOC_FUNCTIONS, FALSE)`.

PR #1032, opened 2022-09-07
Fix typos, clarify some return codes

kamahen:


JanWielemaker:
I have done quite a bit of editing to this patch.    Some of the functions terminate the process on failure (e.g., PL_new_atom()).   All of them (that do not abort) leave an exception already, so just returning FALSE is enough.   Some of them have no error condition.   For example PL_functor_name().   They return an existing atom, so nothing can go wrong.   If the argument is not a valid functor they most likely simply crash (C _undefined behavior :smile: )

Clarifying this is certainly useful :+1:

kamahen:
It's nice to see that the fastest way to get an answer is to post something that's incorrect. ;)

I think I found all the functions that return some kind of id (atom_t, term_t, etc.) and that could return 0 (and a few that - as you pointed out can't). I'll make another pass (in another PR) over the updated documentation, looking for functions that return an id and never return 0. I'll then look at SWI-cpp2.h and adjust it accordingly.

PR #1034, opened 2022-09-08
Uninstaller.exe in Windows 10 is not deleting this registry key

xpxaxsxi:
When removing app in Windows 10 Add/Remove App window, swipl is left hanging in the list of apps. When this registry key is deleted with Registry Editor swipl is removed from list of apps


JanWielemaker:
Thanks!   Merged after rebase and edit of commit message.

PR #1035, opened 2022-09-08
Add missing failure return info

kamahen:


JanWielemaker:
Merged after updates for PL_pred().    That story is a bit more complicated.

PR #1036, opened 2022-09-10
Small fixes around prolog_colour and operators

eshelyaron:


JanWielemaker:
Thanks.  Squashed the two commits for qualify_op/3 and got it a bit closer to the original, avoiding the need for another intermediate predicate.   Note that these predicates should not fail on invalid input as that may happen while editing.

PR #1037, opened 2022-09-10
Registry path now has the missing WOW6432Node

xpxaxsxi:
After testing with daily build (swipl-w64-2022-09-10.exe) Swi-prolog is still in the Add/Remove App list after clicking Remove, so this addition to registry path could  work. I am not testing this commit locally (it is not even possible I think??), I am using the daily builds for testing.


xpxaxsxi:
Forget this  ... this is in a wrong 32-bit section of a Nullsoft installer script, I'll make a new pull request

PR #1039, opened 2022-09-12
Moved uninstall line from  32-bit section to 64-bit

xpxaxsxi:
because it was in wrong section and testing with daily build did not work

### DIFF
--- a/cmake/Pack.cmake
+++ b/cmake/Pack.cmake
@@ -101,6 +101,7 @@ if(WIN32)
     set(CPACK_NSIS_EXTRA_UNINSTALL_COMMANDS "
     SetRegView 64
     DeleteRegKey HKLM 'Software\\\\SWI\\\\Prolog'
+    DeleteRegKey HKLM 'Software\\\\Microsoft\\\\Windows\\\\CurrentVersion\\\\Uninstall\\\\Swipl'    
     ")
   else()
     set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS "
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Shouldn't this be in both?

xpxaxsxi:
I would like to see the 64-bit version work in a daily build first, but probaply the line should be in both

JanWielemaker:
> I would like to see the 64-bit version work in a daily build first, but probaply the line should be in both

In that case you can test with the 32-bit version, no?   Removing it and re-adding it doesn't make the history easier to explain :cry:

xpxaxsxi:
But the 32-bit section is only used in a Windows 10 that is running in an 32-bit hardware, I think there is none such systems

JanWielemaker:
> But the 32-bit section is only used in a Windows 10 that is running in an 32-bit hardware, I think there is none such systems

Is that true?   Didn't know that.   Surely the _then_ branch will be taken for the 32-bit version, regardless of the OS version.  I'll apply this (after finishing my current stuff), but will leave the one in the 32-bit section.   It can't do any harm.

xpxaxsxi:
Maybe there is Windows 10 in a 32-bit hardware, but I can't test it. The same uninstall line in both 32 and 64 sections is fine to me

JanWielemaker:
Done.   Reports a conflict, but that is due to my edits.   Hope it builds tonight.

xpxaxsxi:
Sorry about my beginner errors with  Github and git, but I today found the bug that is causing the Uninstall-problems: path to Uninstall.exe is double-quoted in registry after installation and removing double-quotes fixes problems. https://github.com/SWI-Prolog/swipl-devel/commit/4f3a54859f6bb95baf0f0d85f5883f232fc95631

PR #1040, opened 2022-09-13
PL_unify_bool() accepts same atom values as PL_unify_bool_ex() and PL_get_bool()

kamahen:


JanWielemaker:
Makes sense.   Changed the implementation a bit to avoid the double PL_unify_atom() 
calls.  Besides being unnecessarily slow it can also lead to raising exceptions twice.   I think that is harmless, but not nice.

PR #1042, opened 2022-09-16
ADD: License to CBG Chords example

PaulBrownMagic:
I've added in a CC license because this feels a bit like a publication. It's permissive minus commercial (so no-one starts selling it in an app store or something): https://creativecommons.org/licenses/by-nc/4.0/
If that's a problem for SWI, let me know and I'll allow the commercial.

I put the license in the footer, which also contains links so I moved the attribution and the links from the "About" section there too. 

Now it looks like this:
 
image

### DIFF
--- a/src/wasm/cbg.html
+++ b/src/wasm/cbg.html
@@ -6,6 +6,32 @@
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootswatch/4.1.3/sketchy/bootstrap.min.css"/>
 
     <title>CBG Chords</title>
+    <!--
+    BSD 2-Clause License
+
+    Copyright (c) 2022, Paul Brown
+
+    Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


JanWielemaker:
> If that's a problem for SWI, let me know and I'll allow the commercial.

Thanks.   Unfortunately that will not work.   This is -in part- software and Linux distros require compatibility with the GPL and the GPL explicitly allows for commercial use.   I'm not a real expert, but cc and software seems a complicated combination from a little Googling.  I think a software license is more applicable as the core is software.   GPL is enough to stop the app store as you cannot buy the app and modify the code as demanded by the GPL.

PaulBrownMagic:
Ok, I've swapped it to BSD 2-Clause, keeping a mention in the footer and putting the full text in a HTML comment.

JanWielemaker:
Thanks!   Squashed the two commits, which causes github to claim a conflict.   It is processed.

PR #1044, opened 2022-09-19
C/C++ flags replicated to packages

kamahen:


kamahen:
This assumes that a package is positioned "below" the swipl-devel directory, which I think is a safe assumption.

JanWielemaker:
This should not be needed.   Settings are inherited over included subdirectories by default in cmake.  As you observed already, it are the C++ files that do not inherit the ASAN flags.   I've pushed a fix for that yesterday as I was handling a report on supporting ASAN using Clang that affected the same BuildType.cmake.   Just checked `ffi4pl.so` and that now indeed links against the asan instrumentation.

I'll close this for now.    Please reopen if I missed something anyway.

kamahen:
I see that you've added code to src/pl-main.pl for detecting address-sanitizer with Clang.

Would it make sense to either add this to SWI-Prolog.h (or some other header file), or make it available via a config.h?
Some of my tests need to know if address-sanitizer is on because there appears to be a bug in the way that address-sanitizer handles failure from C++ "new".

JanWielemaker:
> Would it make sense to either add this to SWI-Prolog.h

I don't think we want this ugly stuff in the public header.   What could be an option is to define ``__SANITIZE_ADDRESS__`` in cmake, so you only have to add a line 

     #cmakedefine __SANITIZE_ADDRESS__ @__SANITIZE_ADDRESS__@

to the config files where you need it.   Possibly we should rename the condition to avoid warnings from compilers that set this.    For now it might be easier to just copy the stuff from `pl-incl.h`.   That is the price we sometimes must pay for decoupling the packages from the core.

kamahen:
I'm confused.
Both `pl-main.c` and `pl-init.h` have identical code for handling address-sanitizer for both gcc/g++ and clang - the `pl-init.h` code defines `__SANITIZE_ADDRESS__` if the compiler is clang, so code that uses `pl-init.h` can assume the GNU conventions.

Is there a reason for `pl-main.c` not having `#include "pl-init.h"`?

(The only use of `pl-init.h` that I could find is in `pl-incl.h`, which seems to be used almost everywhere, except in `pl-main.c`)

Anyway, for now, I'll do as you suggest, and copy the code from `pl-incl.h`, with a `TODO`.

PR #1045, opened 2022-09-21
^ and ** arithmetic operator optimization for small integer use cases

ridgeworks:


src/pl-arith.c
```diff
@@ -2345,6 +2345,24 @@ ar_pow(Number n1, Number n2, Number r)
 	    r_bits/8 < (int64_t)globalStackLimit()
 	  ) )
       return int_too_big();
+
+    /* optimization when n1 is a V_INTEGER and the result fits in a V_INTEGER */
+    if (n1->type == V_INTEGER && r_bits < sizeof(long)*8-1)
```
JanWielemaker, line 2350: As I feared, this is a little too simple: `?- A is 3037000500^2` --> A = -9223372036709301616 :cry:    For the stack overflow check this is sort of reasonable as filling the whole stack with a single integer is not so useful anyway.

We need something safe.  I guess one bit more margin makes it safe.   Can we prove that?  Alternatively we should use mul64().  This is really complicated, but I just discovered that GCC and Clang have builtin instructions that use the hardware and should thus be not slower than normal multiplication.  See https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html.   I'll have a look at that.

ridgeworks: Something strange is going on - on my config:
```
?- A is 3037000500^2, B is A/3037000500.
A = 9223372037000250000,
B = 3037000500.
```

JanWielemaker: Almost must be some mixup on your side.   r_bits is `msb(3037000500)*2`, which is 62, using the fast-path implementation.   Also confirmed r_bits is indeed this value.   3037000500 is `ceil(sqrt(1<<63))`, i.e. its square just doesn't fit 64 bits.

JanWielemaker:
Edited the patch quite a bit.   Using GCC/Clang builtin support for multiplication with overflow check this is about as fast as normal multiplication, so I kept the conservative size estimate.    Moved the computation to a function for better readability and now use most of the special case code also for the case there is no GMP support.

Jan Burse's exact case now runs in 7.8 sec (vs 24 before).    Thanks.

PR #1047, opened 2022-09-25
PORT: Msys2, fix LibDir for swipl in build env

mgondan:


library/build/tools.pl
```diff
@@ -707,17 +712,21 @@
...
-    expand_file_name(Pattern, MsysDirs),
-    last(MsysDirs, MSysBinDir),
-    prolog_to_os_filename(MinGWBinDir, WinDirMinGW),
-    prolog_to_os_filename(MSysBinDir, WinDirMSYS),
-    getenv('PATH', Path0),
-    atomic_list_concat([WinDirMSYS, WinDirMinGW, Path0], ';', Path),
-    setenv('PATH', Path).
+    !.
+
+%mingw_extend_path :-
```
JanWielemaker, line 719: I think you are assuming people run Prolog from an MSYS2 environment.   They may only have it installed.    This surely was needed for some reasonable setup.   Wouldn't it be better to first check that gcc.exe is on PATH?   That is fairly simple as the Prolog search path `path` maps to PATH, so `absolute_file_name(path('gcc.exe'), _, [access(exist), file_errors(fail)])` should do.

mgondan: Jan: I think you are assuming people run Prolog from an MSYS2 environment.

Response: Technically, no. I assume that if people want to compile packs, they have a gcc somewhere in their path. Of course, this is not too different from your conjecture.

mingw_extend_path created trouble on my computer because I actually have a very old mingw installation in c:\msys\1.0 which is not on the path. I was a bit surprised to see that the path to the old gcc was silently prepended. 

Given that mingw_extend_path autodetects a gcc installation, I suggest to show a little message instead (?). Something like "gcc detected in c:/msys/1.0/bin. Please add gcc.exe and make.exe to your PATH."

JanWielemaker:
Ok.  Merged.  Changed a little to actually do the PATH update and warn.   Together with the existence check this should be enough and doesn't bring regression for setups that relied on extending the PATH.

PR #1049, opened 2022-10-03
Set @rpath correctly for modules when installing to prefix

jamesnvc:
Otherwise on macOS, if libswipl.dylib is in a non-standard location, shared modules (e.g. sweep-module) will be unable to find it

As discussed [on the forum](https://swi-prolog.discourse.group/t/sweep-swi-prolog-embedded-in-emacs/5713/46).


JanWielemaker:
Thanks!

PR #1050, opened 2022-10-03
Improvements to integer to float conversion

ridgeworks:
Faster version of function to convert MPZ to double.
Minor fixes to `mpz_fdiv` for some rounding modes.
Includes additional test cases for rounding various integer sub-types.

### DIFF
--- a/src/Tests/rational/test_ieee754.pl
+++ b/src/Tests/rational/test_ieee754.pl
@@ -573,7 +573,27 @@ test(ieee_rndto) :-
     assertion(test_roundto(e)),
     assertion(test_roundto(sin(pi/2))),
     assertion(test_roundto(atan2(1,-1))),
-    assertion(test_roundto(cosh(-0.5))).
+    assertion(test_roundto(cosh(-0.5))),    
+      current_prolog_flag(max_tagged_integer,MTI),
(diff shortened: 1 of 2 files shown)


JanWielemaker:
Thanks for sorting this out!   Merged after some white space and layout fixes as well as turning mpz_float into mpz_to_double.

kamahen:
It appears that this (or a related change) has caused a small leak. This is what I get if I simply start and stop swipl. (I haven't run ASAN for a while, so don't know when this leak was introduced)

Direct leak of 40 byte(s) in 5 object(s) allocated from:                                                         
    #0 0x7f45559bbc47 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7f4554b2359d in __gmp_default_allocate (/lib/x86_64-linux-gnu/libgmp.so.10+0xa59d)
    #2 0x7f4554b35e13 in __gmpz_init_set_ui (/lib/x86_64-linux-gnu/libgmp.so.10+0x1ce13)

JanWielemaker:
Thanks.  Fixed with 293e90dea07c50f2bf794947da1baa062fe0f1df.   Introduced MPZ_ONE was not reclaimed, but removed again as it is also no longer used.

PR #1053, opened 2022-10-08
Update GNU Emacs support

eshelyaron:
This PR does the following:
1. Add a new Prolog flag `libswipl_dir` which holds the absolute path of the directory in which `libswipl` will be installed according to the effective `cmake` flags.
2. Update the manual section about GNU Emacs interfaces.
3. Make it possible to load `libswipl` itself as a dynamic Emacs module that does nothing except for making its other symbols visible to other dynamic Emacs modules.

Note: I accidentally pushed these changes to the `master` branch directly (my apologies), so I reverted them to allow for a proper review. This PR simply reverts that revert.


eshelyaron:
Superseded by #1055

PR #1055, opened 2022-10-11
Support loading libswipl as a GNU Emacs dynamic module

eshelyaron:
This is another attempt following #1053


JanWielemaker:

https://swi-prolog.discourse.group/t/sweep-module-so-undefined-symbol-pl-version-info/5842/13

PR #1059, opened 2022-10-20
Add missing int types to PL_cvt_i_*()

kamahen:
and some wording changes to foreign.doc

The missing int types fill the list of ints given at https://en.cppreference.com/w/cpp/types/numeric_limits and should handle things such as size_t etc. They are mainly for simplifying some C++ code that uses overloading.

### DIFF
--- a/man/foreign.doc
+++ b/man/foreign.doc
@@ -776,7 +776,7 @@ a single thread is used to manage the various states. The lack of good
 support for destructive state updates in Prolog makes it attractive to
 use threads for dealing with multiple inputs. The fact that Prolog
 discourages using shared global data such as dynamic predicates
-typically makes multi-threaded code easy to manage.
+typically makes multithreaded code easy to manage.
 
 It is not clear how much scalability we gain using Prolog engines
 instead of Prolog threads. The only difference between the two is the
@@ -850,7 +850,7 @@ by the atom or blob regardless of its type. New code that uses blobs
 should use the blob functions such as PL_blob_data() to get a pointer to
 the content, the size of the content, and the type of the content. Most
 applications that need to get text from a \ctype{term_t} handle should
-use PL_get_nchars() or PL_get_wchars(). If it is \emph{known} that
+use PL_atom_nchars(), PL_atom_wchars(), or PL_atom_mbchars(). If it is \emph{known} that
 \arg{atom} is a classical Prolog text atom, one can use PL_atom_nchars()
 to obtain the C string and its length (for ISO-Latin-1 atoms) or
 PL_atom_wchars() to obtain a C wide string (\ctype{wchar_t}).
@@ -1287,6 +1287,8 @@ strlen().  See PL_new_atom() for error handling.
 
     \cfunction{const char *}{PL_atom_nchars}{atom_t a, size_t *len}
 Extract the text and length of an atom.
+If you do not need the length, pass NULL as the value of \exam{len}.
+If this is called for a blob, NULL is returned.
 \end{description}
 
 
@@ -1428,11 +1430,11 @@ Process an \jargon{option list} as we find with, e.g., write_term/2 and
 many other builtin predicates. This function takes an option list (or
 dict) and in the variadic argument list pointers that receive
 the option values.  PL_scan_options() takes care of validating the list,
-ensure the list is not cyclic, validate the option type and store the
+ensuring the list is not cyclic, validating the option type and storing the
 converted values using the supplied pointers.
 
-Below is an example. While \ctype{PL_option_t} is a struct its members
-are inidialised using the PL_OPTION() macro. The data structure is not
+Below is an example. While \ctype{PL_option_t} is a struct, its members
+are initialised using the PL_OPTION() macro. The data structure is not
 constant because PL_scan_options() adds the option names as
 \jargon{atoms} to speed up option processing.  The macro PL_OPTIONS_END
 terminates the option list.
@@ -1461,10 +1463,11 @@ mypred(term_t a1, term_t options)
 
 The only defined value for \arg{flags} is currently \const{OPT_ALL}, which causes
 this function to raise a domain error if an option is passed that is not
-in \arg{specs}. Default in SWI-Prolog is to silently ignore such
-options. The \arg{opttype} argument defines the type (group) of the
+in \arg{specs}. Default in SWI-Prolog is to silently ignore unknown
+options, unless the Prolog flag \prologflag{iso} is \const{true}.
+The \arg{opttype} argument defines the type (group) of the
 options, e.g., \exam{"write_option"}. Option \jargon{types} are defined
-by the ISO standard. SWI-Prolog only uses this if the \const{OPT_ALL} to
+by the ISO standard. SWI-Prolog only uses this if \const{OPT_ALL} is specified, to
 raise a \const{domain_error} of the indicated type if some option is
 unused. The type name is normally the name of the predicate followed by
 \const{_option} or the name of a representative of a group of predicates
@@ -1499,7 +1502,7 @@ Accepts	an atom.  Note that if the C function that implements
 the predicate wishes to keep hold of the atom after it returns
 it must use PL_register_atom().
     \definition{\const{OPT_TERM}   \ctype{term_t}}
-Accepts an arbitrary Prolog term.   The ternm handle is scoped by
+Accepts an arbitrary Prolog term.   The term handle is scoped by
 the foreign predicate invocation.  Terms can be preserved using
 PL_record().
 \end{description}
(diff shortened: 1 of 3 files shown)


kamahen:
I intend to add tests for this in packages/cpp and also test it on win32 and win64.
The changes to packages/cpp aren't finished; they require these changes to SWI-Prolog.h (or equivalent code in SWI-cpp2.h) but the changes are pure additions, so shouldn't affect anything except packages/cpp.

JanWielemaker:
I'm happy with the functions.   I do not understand why PL_cvt_i_bool() is getting integers.   Booleans in Prolog are (canonically) `true` and `false`, why not simply call PL_get_bool_ex() from PL_cvt_i_bool()?

As for the long long functions, I think we can simply call PL_cvt_i_int64() and PL_cvt_i_uint64().   I doubt there is any existing >= 32 bits system where long long is less than 64 bits (and SWI-Prolog can't run on 16 bit systems).  If at some point we'll see systems using 128 bits long long we'll need some reconsideration anyway as non-GMP integers are now pretty much hard wired at int64_t.    Just calling the 64 bit versions will raise a compiler error it the assumption is not true.

As these are trivial changes, I do not need a PR.   Just wanted to check whether you agree.

kamahen:
These additional PL_cvt_i_*() functions are to simplify SWI-cpp2.h.

It seems that the C++ language distinguishes integer types based on their names ([unsigned] {char,short,int,long,long long]), not on the underlying representation. Even if "long" and "long long" are both 64 bits, they are considered different types because their "names" are different. And, if int64_t is defined as "long" and size_t is defined as "long long", then int64_t and size_t would be considered different even though both are 64-bit. (Note: I haven't read the C++ standard - nor do I intend to - but this seems correct from reading other things, particularly https://en.cppreference.com/w/cpp/types/numeric_limits).

So, I want a complete set of cvt_i_*() functions to cover all the possible [unsigned] {char,short,int,long,long long]  possibilities.

In addition, there's cvt_i_bool(). "bool" isn't in C but is in C++, and is sort-of an integer. I had implemented cvt_i_bool() as a range-checked [0,1] integer, similar to how 32-bit is done; but I now think that it's better to do it as a test for non-zero  (I interpret "cvt_i_*" as "convert integer" - that's why cvt_i_bool() doesn't handle atoms "true" and "false)

This will require a bit of rewriting; I'll update this PR "soon".

JanWielemaker:
> So, I want a complete set of cvt_i_*() functions to cover all the possible [unsigned] {char,short,int,long,long long] possibilities.

I understand that.   Just, in the C world we do not have that problem.  As is, `long long` is synonym for `int64_t` and if this is not the case we need something else than what you wrote anyway (we must interface between 128 bit and GMP).    Simply calling the 64 bit versions creates a simple alias in one line of code and I'd suspect modern compilers to merely claim a symbol alias, worst case a jump.   So, my proposal is simply this.   If `sizeof(long long) != sizeof(int64_t)` the compiler will complain and well need to do something depending on GMP or many decide on `int128_t` as non-gmp integer type.

```
int
PL_cvt_i_llong(term_t t, long long *pv) 
{ return PL_cvt_int64(t, pv);
}
```

> In addition, there's cvt_i_bool(). "bool" isn't in C but is in C++, and is sort-of an integer. I had implemented cvt_i_bool() as a range-checked [0,1] integer, similar to how 32-bit is done; but I now think that it's better to do it as a test for non-zero (I interpret "cvt_i_*" as "convert integer" - that's why cvt_i_bool() doesn't handle atoms "true" and "false)

I still do not get this.   In C++ a bool is 0 or 1.  In Prolog it is `false` or `true`.    PL_cvt_i_* converts from Prolog to C, so (IMO) it should translate `true` to 1 and `false` to 0 (as I guess the C++ constants are defined as well.   PL_get_bool_ex() accepts `true`, `false`, an integer and `yes` or `no` and maybe `on` or `off`. Why only accept integers?

kamahen:
I couldn't find the documentation for cvt_i_*() functions at
https://sicstus.sics.se/sicstus/docs/latest4/html/sicstus.html
https://quintus.sics.se/isl/quintus/html/quintus/

I assumed that `PL_cvt_i_int(term_t p, int *c)` means that `p` must succeed with integer/1; similarly, `PL_cvt_i_bool(term_t p, bool *c)` would require integer/1 and not match atoms (except "bool" isn't a type in C, so we'd use "int"). This is to complete the list of all "integer" types (which include bool, char, int, long, unsigned, etc.).

`PL_cvt_llong()` needs to also work on win32 (I think that `size_t` is defined as `unsigned long long`, for example), so it seemed easiest to write it the way I did. But I can look again and see if I can simplify it to either a #define or 1-line function call (which, as you say, modern compilers possibly optimize away; and if they don't, the prefetch queue on the CPU takes care of).

JanWielemaker:
> I couldn't find the documentation for cvt_i_*() functions at

These functions got there to make the automatic wrapper generation using Quintus/SICStus style foreign function declarations easier.  In that sense, supporting long long is a sensible addition.   I don't know how Quintus/SICStus do this.

I do not see bool as an integer type.

Also on win32, long long is 64 bits.   AFAIK, on Win32 we have `int`, `long`, `size_t` as 32 bits and `long long` as 64 bits.  Win32 is pretty much like 32 bit Unix except that `wchar_t` is 16 bits.   Win64 has the same `wchar_t` and the oddity that `long` is the same as `int` (32 bits).   It is the only platform where `sizeof(long) < sizeof(void*)` that I know about.

kamahen:
This is the code I have now in SWI-cpp2.h for class `PlTerm`):
```cpp
  void integer(char               *v) const { PlCheck(PL_cvt_i_char(     C_, v)); }
  void integer(signed char        *v) const { PlCheck(PL_cvt_i_schar(    C_, v)); }
  void integer(unsigned char      *v) const { PlCheck(PL_cvt_i_uchar(    C_, v)); }
  void integer(short              *v) const { PlCheck(PL_cvt_i_short(    C_, v)); }
  void integer(unsigned short     *v) const { PlCheck(PL_cvt_i_ushort(   C_, v)); }
  void integer(int                *v) const { PlCheck(PL_cvt_i_int(      C_, v)); }
  void integer(unsigned int       *v) const { PlCheck(PL_cvt_i_uint(     C_, v)); }
  void integer(long               *v) const { PlCheck(PL_cvt_i_long(     C_, v)); }
  void integer(unsigned long      *v) const { PlCheck(PL_cvt_i_ulong(    C_, v)); }
  void integer(long long          *v) const { PlCheck(PL_cvt_i_llong(    C_, v)); }
  void integer(unsigned long long *v) const { PlCheck(PL_cvt_i_ullong(   C_, v)); }
```
This covers all the "integer" types from the C "limits" header, except `bool`.

So, I defined it for `bool` like this (it's different because C doesn't have `bool`, and `pl-incl.h` defines `bool` as `int`):
```cpp
  void integer(bool               *v) const { int v0; PlCheck(PL_cvt_i_bool(C_, &v0)); *v = v0; }
```
If you think that `PL_cvt_i_bool()` should also support `true`, `false, `on`, `off` in addition to "not zero", I can add that.

This is kind of a corner case because `bool` is a kind of integer in C++ but boolean is treated somewhat different in Prolog by convention. For other kinds of integers, this isn't a problem, so `PL_cvt_i_int()` etc probably shouldn't also support `true`, `false`, etc.

kamahen:
Maybe the best thing (for now, anyway), is to leave out `PlTerm::integer(bool *v)`. If someone needs it, then we can figure out what to do. Having said that, even if we don't support `true` and `false`, we can't define it by
```cpp
PlTerm::integer(bool *v) { int64_t v0; integer(&v0); v = v0; }
```
because that won't work for an integer that's more than 64 bits, and the "non zero" test ought to work for any integer (a "can't represent" error should mean `true`).

JanWielemaker:
Thanks.   I agree with your PlTerm.    Merged the branch after changing the implementation for these three functions as I proposed.   I that is definitely how PL_cvt_i_bool() is supposed to work from the design of the PL_cvt_i_() group.   If you don't like that, role your own as inline function in SWI-cpp2.h.

Also thanks for the doc fixes!

kamahen:
What do you think about PL_get_bool() also allowing 0 and non-0 (integer)?
I could add this to SWI-cpp2.h, of course; but I don't like the idea of diverging semantics between the C and C++ APIs. (And it's non-trivial for the "average" programmer to cover the corner cases if they decide to "roll their own").

JanWielemaker:
> What do you think about PL_get_bool() also allowing 0 and non-0 (integer)?

I'm ok with that.   And yes, if we can avoid it, let us avoid different parts of the system doing different things.   Do you make a PR?

kamahen:
No PR yet - I wanted your approval first. I'll see what I can do today.

PR #1060, opened 2022-10-22
PL_get_bool() accepts integer (zero/non-zero)

kamahen:
+ documentation for PL_cvt_i_*()

This change was discussed in https://github.com/SWI-Prolog/swipl-devel/pull/1059

### DIFF
--- a/man/foreign.doc
+++ b/man/foreign.doc
@@ -1051,6 +1051,15 @@ Returns non-zero if \arg{term} is acyclic (i.e.\ a finite tree).
 The functions {\tt PL_get_*()} read information from a Prolog term. Most
 of them take two arguments.  The first is the input term and the second
 is a pointer to the output value or a term reference.
+The return value is \const{TRUE} or \const{FALSE}, indicating the success
+of the "get" operation.
+Most functions have a related "_ex" function that raises a type or
+instantiation error if the argument is of the wrong type (e.g.,
+PL_get_atom() and PL_get_atom_ex().
+
+For integers an alternative interface exists, which helps deal with the
+various integer types in C and C++. They are convenient for use with
+\exam{_Generic} selection or C++ overloading.
 
 \begin{description}
     \cfunction{int}{PL_get_atom}{term_t +t, atom_t *a}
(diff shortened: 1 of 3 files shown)


kamahen:
Test cases are being added to packages/cpp (PR not ready yet: https://github.com/kamahen/packages-cpp/tree/jw4 )

src/pl-fli.c
```diff
@@ -1489,7 +1489,18 @@ PL_get_bool(term_t t, int *b)
     { *b = FALSE;
       succeed;
     }
+    fail;
   }
+  if ( isInteger(w) )
```
JanWielemaker, line 1494: This is problematic.   First of all this should be in PL_get_bool() as PL_get_bool_ex() is merely an exception-raising version of PL_get_bool().  We do not want other differences.   

More dubious however is whether we should allow anything other than 0 or 1?   If not, we are entering the field of what I think is called _generalized Booleans_ (not sure about the term) which for many types defines some member of the type to be `false` and thus the others to be `true`.    So, we get floats (0.0), strings (""), atoms (''), and what more?    I don't think that is the direction we should go.

I don't really see the need for anything but `true` and `false`.  `on and `off` came there because Prolog flags tended to use `on` and `off` in many systems (still do, I think).   That makes `0` and `1` acceptable.

If we we go that way we must also deal with bigints (that are always `true` because integers always use the smallest representation that can represent them when they live on the stacks).

man/foreign.doc
```diff
@@ -2187,6 +2211,43 @@ As PL_get_list(), but raises a type or instantiation error if
 As PL_get_nil(), but raises a type or instantiation error if
 \arg{t} is not the empty list.
 
+    \cfunction{int}{PL_cvt_i_bool}{term_t p, int *c}
```
JanWielemaker, line 2214: Good to add these.    I think they should go to a new section, probably together with the _o_ versions and some explanation about the rationale behind these.   I can take care of that.

src/pl-fli.c
kamahen: This *is* in PL_get_bool(). 

I was of two minds about 0/not-0 vs 0/1 for boolean. I'll change it to 0/1. That will require a small change to PL_get_bool_ex(), to determine whether the cause of failure is due to not being one of the allowed atoms (true,false,on,off) or being a representation error (not atom / {0,1}).

man/foreign.doc
kamahen: OK, I'll leave this as-is, on the assumption that you can easily cherry-pick the PR. You might want to review the comment at line 2128. I've also added a comment about "generic" here:
https://github.com/SWI-Prolog/packages-cpp/pull/24/files
SWI-cpp2.h line 341 (the definition of PlTerm::integer(...))
The test cases are in https://github.com/SWI-Prolog/packages-cpp/pull/24 and I'll modify them as well.

JanWielemaker:
Thanks.  Merged after some moving and rewrites to the docs.   The implementation of PL_get_bool() would not work for big integers (valInt() returns garbage for those).    Removed the domain error for integers.   Not sure about that, but if we go for it we should also change atoms other than the designated atoms as domain errors.    This is one of the areas where domain and type get problematic, i.e., is _bool_ a type (and thus a subtype of integer and atom) or is it a range of these types?

kamahen:
"is _bool_ a type (and thus a subtype of integer and atom) or is it a range of these types?"
I was merely taking the approach that an _int_ had been passed to a function (predicate) and it was out of range, hence a domain error.

kamahen:
It looks like you removed the "see also PL_cvt_i_int()" for PL_get_integer() documentation (and similar for PL_get_long(), etc.). Was that intentional?

JanWielemaker:
Yes.  There is a general remark about the existence of the PL_cvt_i_*() family.   I'm not yet sure how they relate to the PL_get_*_ex() family and whether we should accept both or deprecate one of the families.   There is a large overlap :cry:

kamahen:
Many of the PL_cvt_i_() functions are implemented as wrappers around PL_get_*_ex() functions, which is why I put the "see also" in with the PL_get_*() documentation. The PL_cvt_i_*() functions seem to match more cleanly with C/C++'s model of integer types, but that model is rather a mess.
As you've said, it's probably best to stick to int64_t and uint64_t; but there are often other libraries that use things like "int" (and which often omit the "unsigned" for flags, which tends to result in compiler warnings, as I've found with the rocksdb library).

JanWielemaker:
For output (to Prolog), the `int64_t` functions are fine.  For input (from Prolog) we need to ensure the C type can represent the Prolog integer.   It depends a bit on the scenarios though.   If you write your own code you typically use some wide type, but if you interface to an existing library that uses `int` you want to have an error if the Prolog integer cannot be represented as a C integer.  That also explains why we have the two families.   The `PL_get_*_ex()` were intended for hand written extensions and the `PL_cvt_i_*()` evolved from automatic wrapping.

JanWielemaker:
Bit too deep into other stuff and I never studied _Generic.  Looks interesting though.  Possibly we can simplify the user experience a lot.

kamahen:
If you don't mind, I'd like to restore the references from PL_get_integer() etc. to PL_cvt_i_int() etc.
If I had known about them earlier, that would have saved me some work.

JanWielemaker:
I'm more in favor of keeping the two families separated and describe their origin and intend.   Possibly we should deprecate the PL_get_*ex() family for integers.   Too early to tell as far as I'm concerned.

kamahen:
I have no problem with keeping the two families separate (and even eventually deprecating some of the duplicates). But people don't read (and, especially, don't re-read) documentation from beginning to end -- they tend to look for a part that does what they want and skip the rest. So, the more "see also"s, the better IMHO.

kamahen:
Preserving this discussion at https://github.com/SWI-Prolog/swipl-devel/issues/1062

JanWielemaker:
> "is _bool_ a type (and thus a subtype of integer and atom) or is it a range of these types?"
> I was merely taking the approach that an _int_ had been passed to a function (predicate) and it was out of range, hence a domain error.

These errors are a difficult topic.   Some people like to devote their life to them :smile:   I do not mind too much.  I do like to keep things consistent and simple though and using a domain error for this would also imply a domain error on atoms not being one of the 4 recognized ones.    

The type error is also consistent with this, where we also have the choice to consider a character code a type or a subdomain on integers.

```
42 ?- char_code(C, 35634843).
ERROR: Type error: `character_code' expected, found `35634843' (an integer)
```

PR #1061, opened 2022-10-23
PORT: suppress uninit warnings in Rtools42/gcc10.3

mgondan:
C:/Users/matth/swipl-devel/src/pl-arith.c: In function 'ar_pow':
C:/Users/matth/swipl-devel/src/pl-arith.c:2388:10: warning: 'exp' may be used uninitialized in this function [-Wmaybe-uninitialized]
 2388 |     if ( mul64(op1_bits, exp, &r_bits) )
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/src/pl-arith.c
+++ b/src/pl-arith.c
@@ -2234,7 +2234,9 @@ get_int_exponent(Number n, unsigned long *expp)
   exp = (long)i;
 #if SIZEOF_LONG < 8
   if ( (int64_t)exp != i )
+  { *expp = (unsigned long)0 ; // suppress an uninitialized warning in ar_pow
     return int_too_big();
+  }
 #endif
 
   if ( exp >= 0 )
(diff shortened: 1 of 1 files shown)


mgondan:
Alternatively,

static int
int_too_big(void)
{ GET_LD
  // return (int)outOfStack((Stack)&LD->stacks.global, STACK_OVERFLOW_RAISE);
  outOfStack((Stack)&LD->stacks.global, STACK_OVERFLOW_RAISE);
  return FALSE; // or fail
}

JanWielemaker:
> Alternatively

That is exactly what I was about to propose :smile:   While checking the code around it I have some bad vibes though, so I want to sort that out first.

JanWielemaker:
Pushed 0c0a31e468373ec84a16d08aaa709a37fa3037bc too address this.  Thanks.

PR #1065, opened 2022-11-04
mpz_rootrem - fix for roots of negative values

ridgeworks:


JanWielemaker:
Was the close just a "pressed wrong button"?  Should I merge?   Note that things do not need to be well tested for now.   If it improves something it is good enough.

ridgeworks:
Yes, just an operator error. Merge should be good, although I'm still wrestling with mpz_tdiv_qr and shared input/outputs (Q and D ?). 

I also note there is an improved version of mpz_pow_ui  so I'll start using that once the dust settles.

And I'm looking at a small improvement gcd that uses the 64 bit int version once the a and b values are small enough.

JanWielemaker:
As for shared input/output, it seems some bf_* functions can handle it and others not.   If you use `cmake -DCMAKE_BUILD_TYPE=Debug ...` you should have a debug build and get an assertion error.

PR #1067, opened 2022-11-06
Performance improvement to mpz_gcd (~15%)

ridgeworks:


JanWielemaker:
Thanks.   Merged, but after local rebase it is signaled as cannot merge.   Pushed quite a few details.   Only a few tests are failing now.

PR #1068, opened 2022-11-08
More tuning of mpz_gcd

ridgeworks:
Rework iterative loop to cut number of iterations in half by combing two cases. Results in further improvements of ~ 15% but still 15-20 times slower than GMP. Only significant operation left in loop is subtract. GMP appears to be ~twice as fast as libBF on sub, so doesn't explain difference. (Haven't figured out how to run valgrind yet, so not clear how GMP manages to do so well.)

### DIFF
--- a/src/libbf/bf_gmp.c
+++ b/src/libbf/bf_gmp.c
@@ -108,48 +108,32 @@ mpz_gcd(mpz_t r, const mpz_t n1, const mpz_t n2)
   // gcd is always positive
   mpz_abs(a, n1);
   mpz_abs(b, n2);
-  int d = 0;
 
-  while (mpz_cmp(a, b) != 0)
-  { als1 = mpz_scan1(a, 0);
-    bls1 = mpz_scan1(b, 0);
-
-    switch ((als1 == 0)*2 +(bls1 == 0))
-    { case 0: // both even
-	k = -bf_min(als1,bls1);
-	mul_2exp(a, k);
-	mul_2exp(b, k);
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.    I'm also puzzled by the observation that GMP does this a __lot__ faster while the basic operations are not that much faster.

kamahen:
I wonder if the size of cache on your machine makes a difference. I noticed that some things I ran on dev.swi-prolog.org were about 3x faster than on my machine, and the most significant difference was cache size (16384 KB vs 512 KB, according to /proc/cpuinfo; but bogomips were almost the same).
This kind of thing might not show with valgrind (I don't know how valgrind is implemented).

JanWielemaker:
> This kind of thing might not show with valgrind

There is `--tool=cachegrind`.   Valgrind simply implements a CPU in software.   By default it mimics the host CPU, but you can configure the CPU cache and a lot more.

JanWielemaker:
running bench(gcd, 1000, 200, 200), 100,000  gcd's on random 1,000 pairs of 200 bit integers we see it calls bf_sub() almost 10 million times (~100 times per gcd), which uses about 90% of the time.   Doing the same on the GMP version shows this picture
gmp-gcd

ridgeworks:
Thanks for the profiling data. It suggests that there's not much more can be done with a pure Stein implementation (90% in bf_sub).  Looks like GMP has some custom built gcd code.

I'll keep digging.

PR #1071, opened 2022-11-13
mpz_gcd minor safety(?) issue & code cleanup

ridgeworks:
1. Use `mpz_mul_2exp` rather than internal version without rounding to round any final result. Internal version now only used for right shifts.

2. Code cleanup to remove any direct calls to libBF.

Should have no measurable impact on performance.


JanWielemaker:
Thanks.   Not sure I like the bf_is_zero() to mpz_sgn() == 0.   Doesn't seem to do any measurable harm though.     Eventually I consider to define a more general binding, abstracting over gmp/libbf, so we can exploit the better embedding API of LibBF.

PR #1072, opened 2022-11-14
Further optimizations for mpz_gcd (libbf.c and bf_gmp.c)

ridgeworks:
Infinite precision fastpath in __bf_round generally useful for any arithmetic op.


JanWielemaker:
Thanks.  Merged (but github doesn't see the rebase done locally).    Roughly 50% is now in add_internal() and 20% in normalize and rounding.

PR #1075, opened 2022-11-25
find windows dlls from packages

mgondan:


boot/build_home.pl
```diff
@@ -197,6 +197,9 @@
 %   probably cheaper to add it anyway.
 
 add_package_path(PkgBinDir) :-
+    (   current_prolog_flag(windows, true)
```
JanWielemaker, line 200: Why is this needed?   This file is there for running Prolog in the build directory, which is needed for running the tests and Prolog steps of the build.    On Windows, all packages are designed to drop the final `.dll` in the `src` directory where `swipl.exe` is installed as well.

mgondan:
I forgot the use case. There was indeed a problem when loading foreign packs from the build directory under the MSYS2 system, but I cannot reproduce it anymore.

PR #1081, opened 2022-12-19
PORT: prolog flag for msys2

mgondan:


library/shlib.pl
```diff
@@ -573,15 +573,16 @@
...
 
 add_mingw_dll_directory :-
-    (   getenv('MINGW_PREFIX', Prefix)
-    ->  format(atom(Bin), '~w/bin', [Prefix]),
-        win_add_dll_directory(Bin)
-    ;   true
-    ).
+    current_prolog_flag(msys2, true),
+    !,
+    getenv('MINGW_PREFIX', Prefix),
```
JanWielemaker, line 581: What if `MINGW_PREFIX` is not defined?

src/CMakeLists.txt
```diff
@@ -398,6 +398,9 @@ target_c_stack(swipl${PGO_SUFFIX} 4000000)
 if(MINGW)
   target_link_options(swipl${PGO_SUFFIX} PRIVATE -municode)
 endif()
+if(MSYS2)
+  add_compile_options(-DMSYS2)
```
JanWielemaker, line 402: Probably more elegant to pass this through config.h.cmake.

src/pl-nt.c
```diff
@@ -332,6 +332,9 @@ CpuCount(void)
 void
 setOSPrologFlags(void)
 { PL_set_prolog_flag("cpu_count", PL_INTEGER, (intptr_t)CpuCount());
+#ifdef MSYS2
+  PL_set_prolog_flag("msys2", PL_BOOL, TRUE);
```
JanWielemaker, line 336: I think this should be a read-only flag.   Also requires documentation (man/overview.doc).

JanWielemaker (commented):
Can fix most of this for you.    What do do with the possible failure to add the dlls is probably something to consider.

mgondan:
> What do do with the possible failure to add the dlls is probably something to consider.

This will raise an error message, that an initialization directive has failed. But that is ok, the msys2 version only works in an msys2 environment, so if the directive fails, there is a real problem.

library/shlib.pl
mgondan: This will raise an error message, that an initialization directive has failed. But that is ok, the msys2 version only works in an msys2 environment, so if the directive fails, there is a real problem.

JanWielemaker: Is this actually necessary?   If dlls like libgmp, gcc runtime, etc are not in the search path, the system will not start at all, so we never get here?    When would we get here?

mgondan: Actually, it does start normally. The problem appears if I use_module(library(archive)), complaining about a missing module, that turns out to be zlib1.dll

src/CMakeLists.txt
mgondan: I see. Two more commits below.

src/pl-nt.c
mgondan: Done, thanks!

mgondan:
> Can fix most of this for you.

I tried it myself :-)

JanWielemaker:
Thanks.   Did some rewrite of the history and squashed some of the commits.   History is good, but enough is enough :smile:

PR #1094, opened 2023-01-11
Minor documentation edits and clarifications for exceptions in foreign predicates

kamahen:


man/foreign.doc
```diff
@@ -3238,10 +3241,11 @@ be used after a call to PL_next_solution() that returns \const{FALSE}
 when the query was created using the \const{PL_Q_PASS_EXCEPTION} flag
 (see PL_open_query()).
 
-Note that an API may only raise an exception when it fails. The
+Note that an API may only raise an exception when it fails; if the API
+call succeeds, the result of \exam{PL_exception(0)} is undefined. The
```
JanWielemaker, line 3245: I did another edit to this.    "Undefined" typically means it is illegal to call PL_exception(0) in that case.   That is not the case.  It returns 0 (no exception) unless there was already a pending exception before calling the API.   This however should not happen in a well behaved client.     This leads to a bit complicated explanation and is why I simply ignored this case earlier, assuming nobody would do this anyway.     In principle, user code can call `if ( PL_exception(0) ) return FALSE` at any time within the definition of a foreign predicate.

JanWielemaker:
Thanks.    I think it is now crystal clear :smile:   Please check the fixes to the remark above.    I'm a little unhappy with that text, but I do not see something that is simpler (and correct).

kamahen:
I rebuilt with the latest commit - 0a145e44bd89e5743e45b0e8828f5d1b98c29620 - and got the same result (except for the version # of course).

kamahen:
I have observed results from PL_exception(0) that caused exceptions when attempting to print. Possibly these are for terms that are now incorrect because of popping the stack. I will investigate further and file bug reports or documentation updates if I find anything.

JanWielemaker:
If a term doesn't print one can be fairly sure it is corrupt :smile:

PR #1100, opened 2023-01-17
Fix some print formats

kamahen:
I'm not sure of the change to pl-trace.c
Some of the changes might not be portable to other platforms - e.g., can they be automatically checked on Windows or on a 32-bit system?

### DIFF
--- a/src/os/pl-cstack.c
+++ b/src/os/pl-cstack.c
@@ -437,17 +437,17 @@ print_trace(btrace *bt, int me)
 #endif
 	       ) &&
 	       addr2line(info.dli_fname, offset, buf, sizeof(buf)) )
-	    Sdprintf("  [%d] %s [%p]\n", i, buf, addr);
+	    Sdprintf("  [%zd] %s [%p]\n", i, buf, addr);
 	  else if ( info.dli_sname )
-	    Sdprintf("  [%d] %s(%s+%p) [%p]\n",
+	    Sdprintf("  [%zd] %s(%s+%lu) [%p]\n",
 		     i, info.dli_fname, info.dli_sname,
 		     (uintptr_t)addr-(uintptr_t)info.dli_saddr,
 		     addr);
 	  else
-	    Sdprintf("  [%d] %s(+%p) [%p]\n",
+	    Sdprintf("  [%zd] %s(+%p) [%p]\n",
 		     i, info.dli_fname, (void*)offset, addr);
 	} else
-	{ Sdprintf("  [%d] ??? [%p]\n", i, addr);
+	{ Sdprintf("  [%zd] ??? [%p]\n", i, addr);
 	}
       }
     }
--- a/src/os/pl-cstack.c
+++ b/src/os/pl-cstack.c
@@ -439,7 +439,7 @@ print_trace(btrace *bt, int me)
 	       addr2line(info.dli_fname, offset, buf, sizeof(buf)) )
 	    Sdprintf("  [%zd] %s [%p]\n", i, buf, addr);
 	  else if ( info.dli_sname )
-	    Sdprintf("  [%zd] %s(%s+%lu) [%p]\n",
+	    Sdprintf("  [%zd] %s(%s+0x%tx) [%p]\n",
 		     i, info.dli_fname, info.dli_sname,
 		     (uintptr_t)addr-(uintptr_t)info.dli_saddr,
 		     addr);
@@ -940,9 +940,11 @@ sigCrashHandler(int sig)
   if ( PL_get_thread_alias(tid, &alias) )
     name = PL_atom_wchars(alias, NULL);
 
+  WPRINT_PUSH
   Sdprintf("\nSWI-Prolog [thread %d (%Ws) at %s]: "
 	   "received fatal signal %d (%s)\n",
 	   PL_thread_self(), name, tbuf, sig, signal_name(sig));
+  WPRINT_POP
   print_c_backtrace("crash");
   Sdprintf("Prolog stack:\n");
   PL_backtrace(25, PL_BT_SAFE);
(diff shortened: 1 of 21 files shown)


kamahen:
This seems to have taken care of all the printf format warnings for the main source ... there are still quite a few warnings from the "packages" directory, which I'll take care of separately.

src/os/pl-cstack.c
```diff
@@ -437,17 +437,17 @@ print_trace(btrace *bt, int me)
 #endif
 	       ) &&
 	       addr2line(info.dli_fname, offset, buf, sizeof(buf)) )
-	    Sdprintf("  [%d] %s [%p]\n", i, buf, addr);
+	    Sdprintf("  [%zd] %s [%p]\n", i, buf, addr);
 	  else if ( info.dli_sname )
-	    Sdprintf("  [%d] %s(%s+%p) [%p]\n",
+	    Sdprintf("  [%zd] %s(%s+%lu) [%p]\n",
```
JanWielemaker, line 442: This probably does not work.   I think addr2line expects hex notation.   So, it is probably better to cast the uintptr_t back to `void*` and use `%p`.    More in general, _long_ is almost always wrong unless the variable is declared `long`.  This rarely happens in SWI-Prolog because long in Win64 is 32 bit.   On this platform `uintptr_t` is the same as `long long`, so it should be `llu` there.   

Typically, we should be using `%zd` or `%p` for pointer-sized objects.   In the purist world this is still incorrect because `size_t` can be smaller than `uintptr_t` on systems with segmented memory.    As is, SWI-Prolog won't run on such systems anyway and I don't know about any actively supported hardware where this happens.

src/os/pl-cstack.c
```diff
@@ -940,9 +940,13 @@ sigCrashHandler(int sig)
   if ( PL_get_thread_alias(tid, &alias) )
     name = PL_atom_wchars(alias, NULL);
 
+  #pragma GCC diagnostic push
```
JanWielemaker, line 943: I think this should be a macro defined in SWI-Prolog.h.    That at least would allow simple switch to other compilers.   In addition, some projects to not allow for `#pragma` to suppress warnings (e.g., the MSYS project)

src/pl-comp.c
```diff
@@ -4786,7 +4786,7 @@ argKey(Code PC, int skip, word *key)
 	goto again;
 #endif
       default:
-	Sdprintf("Unexpected VM code %d at %p\n", c, PC);
+	Sdprintf("Unexpected VM code %ld at %p\n", c, PC);
```
JanWielemaker, line 4789: codes (c) are uintptr_t.   As noted, `l` doesn't work.   As `code` in decoded form is also small, best use %d and cast to `int`.   Should there be cases in encoded form, then cast to `size_t` and use `%zd`.

src/pl-comp.c
```diff
@@ -7081,7 +7081,7 @@ vm_list(Code start, Code end)
   { code op = fetchop(PC);
     const code_info *ci = &codeTable[op];
 
-    Sdprintf("%-3d %s\n", PC-start, ci->name);
+    Sdprintf("%-3ld %s\n", PC-start, ci->name);
```
JanWielemaker, line 7084: Cast to size_t and use %zd

JanWielemaker (changes requested):
All in all it does point at quite a few issues.  Luckily most in debugging code.    To get rid of the `#pragma` we may also consider defining two versions of these functions if `CHECK_FORMAT` is active, e.g. `SdprintfX`, where the `X` denotes it is _eXtended_.   Without `CHECK_FORMAT` we can have a simple `#define Sdprintfx Sdprintf` so there are no costs involved,

src/os/pl-cstack.c
JanWielemaker: Well, the value is a pointer difference, which we compute by casting both to `uintptr_t`.    Possibly would have been better to cast both to `char*`?    The problem is that aliases like `uintptr_t` are `long` on some platforms, but something else on others.     As it was, it was more portable as `uintptr_t` has the size of a pointer and `%p` expects a pointer.    The sizes match.   In theory there could be ABIs that pass pointers differently from integers though ...

In general it seems one need to be careful where the type comes from and what it could be on different platforms.   Then select the format specifier and make sure the passed argument is of matching type, possibly by casting it.    If you make sure the type used for the cast is wide enough, this should always work.

kamahen: The `t` specifier is for `ptrdiff_t` - casting to `char*` seems best as it gives the same result on all platforms.

JanWielemaker:
Merged.   Thanks.   Removed the PUSH/POP macros as we do not use them.    Thanks for adding the functions.   I fear that had to happen sooner or later anyway.    Changed a %ld with a thread identifier to %zd and cast to `size_t` and actually added support to `%tu`.  As we implemented Sdprintf() from the ground we must keep it in sync :cry:    I understand current glibc allows extending printf() and it also allows extending `FILE` AFAIK, but we cannot use these as we must support other C runtime libraries.

kamahen:
Good catch that the `%tu` needed to be supported. There might be a few more in the list: https://en.wikipedia.org/wiki/Printf_format_string#Length_field 

I would have kept the PUSH/POP macro because (a) it took me more than 5 minutes of Googling and testing and (b) not all the *printf functions have a corresponding *printfX function (I can add them if you wish).

JanWielemaker:
On 1/20/23 15:00, Peter Ludemann wrote:
> Good catch that the |%tu| needed to be supported. There might be a few 
> more in the list: 
> https://en.wikipedia.org/wiki/Printf_format_string#Length_field 
> <https://en.wikipedia.org/wiki/Printf_format_string#Length_field>

If you want to add them, be my guest :)

> I would have kept the PUSH/POP macro because (a) it took me more than 5 
> minutes of Googling and testing and (b) not all the *printf functions 
> have a corresponding *printfX function (I can add them if you wish).

I'd say on need-to-have basis.   We'll remember where to look should
we need the PUSH/POP again.  I'm glad they are gone as I'm pretty afraid
they would have let to questions from package managers.  Eventually I
hope GCC or Clang gets its act together and provide a way to add your
extensions to the format template.  There are surely more people who
want that.

	Cheers --- Jan

PR #1101, opened 2023-01-19
ENHANCED: library(prolog_colour): handle arithmetic functions

eshelyaron:
This PR teaches `library(prolog_colour)` to recognize arithmetic expressions when they appear as an argument of the general purpose arithmetic predicates from https://www.swi-prolog.org/pldoc/man?section=arithpreds.
It checks the used arithmetic functions with `current_arithmetic_function/1` and classifies them accordingly.

Here's a screenshot from Sweep leveraging this addition:
Screenshot 2023-01-19 at 20 26 43

### DIFF
--- a/library/prolog_colour.pl
+++ b/library/prolog_colour.pl
@@ -1605,6 +1605,69 @@ colourise_list_args([], none, _, _, _) :- !.
 colourise_list_args([], TP, T, TB, How) :-
     specified_item(How, T, TB, TP).
 
+
+%!  colourise_expression(+Term, +TB, +Pos)
+%
+%   colourise arithmetic expressions.
+
+colourise_expression(_, _, Pos) :-
+    var(Pos),
+    !.
+colourise_expression(Arg, TB, parentheses_term_position(PO,PC,Pos)) :-
+    !,
+    colour_item(parentheses, TB, PO-PC),
+    colourise_expression(Arg, TB, Pos).
+colourise_expression(Var, TB, Pos) :-                     % variable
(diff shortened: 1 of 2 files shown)


library/prolog_colour.pl
```diff
@@ -1605,6 +1605,70 @@
...
+
+colourise_expression(_, _, Pos) :-
+    var(Pos),
+    !.
+colourise_expression(Arg, TB, parentheses_term_position(PO,PC,Pos)) :-
+    !,
+    colour_item(parentheses, TB, PO-PC),
+    colourise_expression(Arg, TB, Pos).
+colourise_expression(Var, TB, Pos) :-                     % variable
+    var(Var), Pos = _-_,
```
JanWielemaker, line 1621: You can have rules for all the stuff you want to take care of and call colourise_term_arg/3 to do the general coloring.    That avoids some duplication.   That applies to variables and numeric types, I guess.   Note that you can have `Dict.key`, so you should also propagate ./2 terms.

eshelyaron: Thanks, done in [724822e](https://github.com/SWI-Prolog/swipl-devel/pull/1101/commits/724822ee2757354a926d45c8ab0c99fb1feb5165).  I've added a simple test case that covers also `Dict.key` in expressions in another commit.

JanWielemaker:
Thanks.   Squashed the commit.    Also fixed the dict key predicate.   Good to add it as a predicate.   The bug was mine: `functor(f(), '.', 2)` raises an exception and we must avoid these.

PR #1088, opened 2023-01-08
feat: create a bundled build of wasm

jeswr:
Follows up on the discussion https://github.com/rla/npm-swipl-wasm/pull/8 to produce a `swipl-bundle.js` build which inlines the `.wasm` and `.data` files.

PR #1103, opened 2023-01-23
handle `preRun` as a function rather than an array

jeswr:
Resolves the issue observed in https://github.com/SWI-Prolog/swipl-devel/pull/1088#issuecomment-1399651681


JanWielemaker:
Thanks!   That fixes the issue running shell.html and the other demos using the bundle.

PR #1104, opened 2023-01-25
Enhancement: add new exact numeric comparison functions

ridgeworks:
New arithemtic functions: cmpr, maxr, and minr.

New function cmpReals added to pl-gmp.

New unit test file test_real.pl in Tests/rational.

Draft documentation for new functions added to manual.

### DIFF
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -8496,6 +8496,15 @@ $\arg{Expr} = 0$.  If \arg{Expr} evaluates to a float, the return value
 is a float (e.g., -1.0, 0.0 or 1.0). In particular, note that sign(-0.0)
 evaluates to 0.0.  See also \funcref{copysign}{2}.
 
+    \function{cmpr}{2}{+Expr1, +Expr2}
+Exactly compares the values \arg{Expr1} and \arg{Expr2} and returns -1 if 
+\arg{Expr1} < \arg{Expr2}, 0 if they are equal, and 1 if \arg{Expr1} > \arg{Expr2}.
+
+ISO standard numeric comparison is inexact because it requires the conversion of any
+non-floats to floats before comparing, potentially incurring a rounding error. \funcref{cmpr}{2}
+implements an exact comparison enabling, for example, the creation of "correct" sort of a
+list of numbers of any type using predsort/3.
+
     \function[ISO]{copysign}{2}{+Expr1, +Expr2}
 Evaluate to \arg{X}, where the absolute value of \arg{X} equals the
 absolute value of \arg{Expr1} and the sign of \arg{X} matches the sign
@@ -8556,6 +8565,18 @@ float -0.0 is smaller than 0.0 as well as the integer 0.
     \function[ISO]{min}{2}{+Expr1, +Expr2}
 Evaluate to the smaller of \arg{Expr1} and \arg{Expr2}. See
 \funcref{max}{2} for a description of type handling.
+
+    \function[ISO]{maxr}{2}{+Expr1, +Expr2}
+Evaluate to the larger of \arg{Expr1} and \arg{Expr2} using exact comparison
+(see \funcref{cmpr}{2}). If the two values are exactly equal, and one of the values
+is rational, the result will be that value; the objective being to avoid "pollution" of
+any precise calculation with a potentially imprecise float. So \exam{max(1,1.0)} 
+evaluates to 1.0 while \exam{maxr(1,1.0)} evaluates to 1. This also means that 0 is preferred over 0.0 or -0.0;
+-0.0 is still considered smaller than 0.0.
+
+    \function[ISO]{minr}{2}{+Expr1, +Expr2}
+Evaluate to the smaller of \arg{Expr1} and \arg{Expr2} using exact comparison 
+(see \funcref{cmpr}{2}). See \funcref{maxr}{2} for a description of type handling.
 
     \function[deprecated]{.}{2}{+Char, []}
 A list of one element evaluates to the character code of this
--- a/man/builtin.doc
+++ b/man/builtin.doc
@@ -8574,6 +8574,10 @@ any precise calculation with a potentially imprecise float. So \exam{max(1,1.0)}
 evaluates to 1.0 while \exam{maxr(1,1.0)} evaluates to 1. This also means that 0 is preferred over 0.0 or -0.0;
 -0.0 is still considered smaller than 0.0.
 
+\funcref{maxr}{2} also treats NaN's as missing values so \exam{maxr(1,nan)} evaluates to 1.
+(This is consistent with the current standards, which also define a propagate version, but 
+may not be implemented in all programming languages.)  
+
     \function[ISO]{minr}{2}{+Expr1, +Expr2}
 Evaluate to the smaller of \arg{Expr1} and \arg{Expr2} using exact comparison 
 (see \funcref{cmpr}{2}). See \funcref{maxr}{2} for a description of type handling.
--- a/src/ATOMS
+++ b/src/ATOMS
@@ -145,6 +145,7 @@ A close_on_exec		"close_on_exec"
 A close_option		"close_option"
 A close_parent		"close_parent"
 A cm			"cm"
+A cmpr			"cmpr"
 A cntrl			"cntrl"
 A co			"co"
 A codes			"codes"
@@ -513,6 +514,7 @@ A max_table_answer_size_action "max_table_answer_size_action"
 A max_table_subgoal_size "max_table_subgoal_size"
 A max_table_subgoal_size_action "max_table_subgoal_size_action"
 A max_variable_length	"max_variable_length"
+A maxr			"maxr"
 A memory		"memory"
 A merged		"merged"
 A message		"message"
@@ -525,6 +527,7 @@ A meta_predicate	"meta_predicate"
 A method		"method"
 A min			"min"
 A min_free		"min_free"
+A minr			"minr"
 A minus			"-"
 A mismatched_char	"mismatched_char"
 A mod			"mod"
@@ -1025,6 +1028,7 @@ F codes			1
 F codes			2
 F colon			2
 F comma			2
+F cmpr			2
 F compound		1
 F context		2
 F copysign		2
@@ -1153,9 +1157,11 @@ F lsb			1
 F lshift		2
 F dict_position		5
 F max			2
+F maxr			2
 F max_size		1
 F message_lines		1
 F min			2
+F minr			2
 F minus			1
 F minus			2
 F mod			2
--- a/src/Tests/rational/test_real.pl
+++ b/src/Tests/rational/test_real.pl
@@ -1,0 +1,178 @@
+/*  Part of SWI-Prolog
+
+    Author:        Rick Workman
+    WWW:           http://www.swi-prolog.org
+    Copyright (c)  2020, University of Amsterdam
+                         VU University Amsterdam
+		         CWI, Amsterdam
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions
+    are met:
+
+    1. Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in
+       the documentation and/or other materials provided with the
+       distribution.
+
+    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+    "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+    LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+    FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+    COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+    INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+    BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+    LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+    CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+    LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+    ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+*/
+
+
+
+:- module(test_real, [test_real/0]).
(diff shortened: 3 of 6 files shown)


man/builtin.doc
```diff
@@ -8557,6 +8566,18 @@ float -0.0 is smaller than 0.0 as well as the integer 0.
 Evaluate to the smaller of \arg{Expr1} and \arg{Expr2}. See
 \funcref{max}{2} for a description of type handling.
 
+    \function[ISO]{maxr}{2}{+Expr1, +Expr2}
+Evaluate to the larger of \arg{Expr1} and \arg{Expr2} using exact comparison
+(see \funcref{cmpr}{2}). If the two values are exactly equal, and one of the values
+is rational, the result will be that value; the objective being to avoid "pollution" of
```
JanWielemaker, line 8572: Is this difference needed?   I consider it rather strange.    We have a precise number and an imprecise number that simply happens to be the same.   To me, this implies the min/max is also imprecise.     Read (im)precise as "polluted" or "tainted" as you wish.   To me, it is all the same.   Floats are mathematically rationals but they typically have a value that differs from what they were intended to represent.

ridgeworks: >We have a precise number and an imprecise number that simply happens to be the same. To me, this implies the min/max is also imprecise.

I disagree. Floating point values are not "imprecise"; floating point arithmetic is. It may be a small distinction, but it's an important one. Further, mixed mode arithmetic currently favours using imprecise floating point arithmetic rather than precise rational arithmetic.  

From that perspective, `minr/maxr` are precise. So I think it's preferable when given the choice between equal values of different types, to choose the rational value which prolongs the use of future "precise" arithmetic operations. As you say, wrap it in `float/1` if some other result, e.g., always a float, is required.

JanWielemaker: I guess we'll never agree.   I'd rather avoid this difference to min/max though.  Do we have prior art in this case?    Notably from dynamically typed languages that tried to get it right like CommonLisp.    According to http://www.lispworks.com/documentation/HyperSpec/Body/f_max_m.htm, either is allowed.  JavaScript refuses Math.min for big integers (and normal integers are floats ... we can barely take this language serious when it comes to numbers).   Python seems to take the first ...

ridgeworks: > I guess we'll never agree.

I don't think I can ever agree to anything that claims a floating point value is anything other than a specific point on the real number line. By my interpretation, that's how IEEE, and every computer language that supports floats language that I can think of, defines it.

As for prior art, I doubt you'll find much agreement if mixed modes are even allowed. I find the Python approach quite odd in that `max(N1,N2)` can return a different value than `max(N2,N1)` when the arguments are mixed mode.

I'd also like to make the point that if one really wanted to implement in Prolog the semantics I suggest for `maxr/minr` it's a bit convoluted and inefficient. I think you would have to do a `cmpr` and, if equal, select one or the other using a filter, e.g., for maximum:
```
C is cmpr(N1,N2), 
(C =0 -> (rational(N1) -> R = N1 ; R = N2)
 ; C = -1 -> R = N2
 ; R = N1
)
```
Now for similar reasons, we have other operations, e.g., '`/`' and '`**`', whose return value is subject to the setting of the `prefer_rationals` flag. I guess I could buy into using that flag to determine which value got returned (float or rational) when a mixed mode `maxr`/`minr` was evaluated. The general argument regarding retaining precision seems to apply in all these cases.

kamahen: @ridgeworks 's  words have made me change my view from "floating point numbers represent a range of values" (which is kinda sorta what I got from my Numerical Analysis courses) to "floating point numbers are precise but are the result of imprecise computations". (For one thing, figuring out the range of values made up more of my Numerical Analysis courses than I care to think about, and also my Physics and Chemistry courses ... and floating point numbers don't capture the range information.) And that point of view seems to simplify things as well.

As for Python, I think `max(x,y)` is defined as `x if x>=y else y`, so it somewhat makes sense that `max(1.0,1)==1.0` but `max(1,1.0)==1`. But of the many flaws in Python, this is one of the more trivial. (Would you prefer `x if x > y else y`?)

JanWielemaker: The `prefer_rationals` flag has a quite different meaning.   Surely I typically want a rational from `1/3`, but a float from `max(1,1.0)`.   I don't see why you care considering your view on floats.  For your purposes both should be equivalent.

I do agree floats are exact values and it is the computation (or just reading as for e.g. `0.1`) that is imprecise.   Still, that makes a float in general _tainted_ and I like to preserve the fact the computation is tainted.

Which is also why I would like to see max(NaN,1.0) to be NaN.   There also seem to be different NaNs.  `sqrt(-2.0)` is NaN as there is clearly no number that satisfies this (if we ignore complex numbers).   `0.0/0.0` is less clear.   If we have float underflow defined to valuate to 0.0 it may well be that the mathematical result of the computation is well defined.   In this case our `max(NaN, 1.0)` would best evaluate to a constraint `X >= 1.0`.    If we consider the NaN from `sqrt(-2.0)` as logical failure we would never have reached the `max` evaluation and thus the result should thus also be `NaN`.

ridgeworks: > The prefer_rationals flag has a quite different meaning.

How so? For both '`/`' and `maxr/minr` don't they both say given the choice between and rational and a float, use the flag to choose? At least that's one valid interpretation IMO. 

> Surely I typically want a rational from 1/3, but a float from max(1,1.0). I don't see why you care considering your view on floats. For your purposes both should be equivalent.

Let me assure you they are not equivalent to me. I rather dislike arguing my points on specifics of my application, but here goes:

I'm trying to build a constraint system over the domain of reals (and I do mean reals) using interval arithmetic. In this domain a real is either a precise number or a closed interval represented in Prolog by a tuple `(L,H)` where `L` and `H` can be any Prolog numbers other than NaN, and `L < H`. This tuple is contained in the attribute of a Prolog attributed variable. If an interval ever narrows (and intervals can only get narrower) such that  `L` is exactly equivalent to `H` then the attributed variable is unified with that value. 

Because rational arithmetic is precise and floating point arithmetic is not, there is a strong preference for rational bounds to stay rational ( i.e., *untainted*). Floating point bounds get rounded outwards as they're calculated to ensure the real value stays contained in the interval, so the precision (width of the interval) is degraded as new floating point bounds are calculated. In fact, it is quite rare that any interval with floating point bounds can ever narrow to a point value given outward rounding.

(At some point however rationals can get converted to floats via the tripwire, and that's OK because the correctness properties are maintained.)

Constraint propagation works by calculating a new set of bounds for a real, and then intersecting that set with the existing set to define  the new set. Intersection is basically a pair of min/max operations:
```
intersect((Lold,Hold),(L,H),(Lnew,Hnew)):-
    Lnew is maxr(Lold,L), Hnew is minr(Hold,H),
    1 =\= cmpr(Lnew,Hnew).  % Lnew =< Hnew 
```
Note that `minr/maxr`, as I've defined them, calculates the new bounds while, at the same time, eliminating any NaNs resulting from the bounds calculations and achieving the objective of keeping rational bounds rational.

This is a huge performance win for me because there are two or three interval intersections performed for each relational interval arithmeitic primitive (`add`, `mul`, `le`, `powr`, `abs`, `sqrt`, etc.). And each step of a fixed point iteration (which propagates constraints) typically involves the execution of ten's of RIA primitive ops. So we're talking about thousands these operations in even a relatively simple problem, e.g., find the roots of a Horner form 4th degree polynomial:
```
?- clpStatistics,X::real,{3+X*(-4+X*(4+X*(-4+X)))==0},solve(X),writeln(X),fail;clpStatistics(SS).
_96{real(0.9999999967903018,1.000000009629096)}
_96{real(2.999999999999999,3.000000000000001)}
SS = [userTime(0.04706700000002684), gcTime(0.002), globalStack(1768/524256), trailStack(48/264168), localStack(2008/118648), inferences(187531), narrowingOps(4875), narrowingFails(31), node_count(0), max_iterations(720/3000)].
```
so almost 5000 primitive ops and probably close to twice that many interval intersections. After correctness, performance is a primary concern and that's why I care.

ridgeworks: > As for Python, I think max(x,y) is defined as x if x>=y else y, so it somewhat makes sense that max(1.0,1)==1.0 but max(1,1.0)==1. But of the many flaws in Python, this is one of the more trivial. (Would you prefer x if x > y else y?)

I looked at the Python function again and it's really a superset of what we're talking about. It's variadic and supports a list/iterable argument, as well as non-numeric values. In that context, when two values are equivalent, making the choice based on argument position perhaps isn't so odd - pick the first max/min value in the sequence.

It still strikes me a bit odd for the case under discussion, i.e., comparing two numeric values. The specific (corner) case of interest is when the two values are numerically equivalent but have additional different properties (type = representation, arithmetic semantics) on which to base a decision. To me it makes more sense to use this as a criteria rather than argument position.

ridgeworks: > Which is also why I would like to see max(NaN,1.0) to be NaN.

I avoided getting into this issue in my earlier post because I thought there might already be too much detail there. In this case there seems to no consensus in the programming language community,  but I believe the IEEE FP standard has been pretty clear, although they've allowed both in the 2019 version (by different names). Kahan (considered to be the father of the standard) states in his [1996 notes](http://www.cs.berkeley.edu/~wkahan/ieee754status/ieee754.ps) : 

> Some familiar functions have yet to be defined for NaN . For instance max{x, y} should deliver the same result as max{y, x} but almost no implementations do that when x is NaN . There are good reasons to define max{NaN, 5} := max{5, NaN} := 5 though many would disagree.

Unfortunately, he fails to elaborate on the "good reasons".

The Eclipse proposal, as well as C (libm), Rust, Python (as long as NaN isn't the first argument in a sequence), SWI-Prolog (currently), all follow the IEEE lead. And I honestly don't see the problem with min/max choosing a number vs. not-a-number. As I suggested in the earlier post, it's a practical way of culling an unwanted result in the middle of expression evaluation.

And there are other arithmetic functions that take NaN as an argument and don't produce an NaN, e.g., `1.0 =:= nan**0.0`, an example taken from straight Kahan's notes. In the same vein, but less justifiable perhaps: `1.0 =:= 1**nan` (as defined by the `pow` function in libm).

Another reference point which appears to recognize the lack of consensus: [ISO 10967 Information technology — Language independent arithmetic](https://www.open-std.org/JTC1/SC22/WG11/docs/iso10967-2_2001_e.pdf) 

> 5.2.2 Floating point maximum and minimum
The appropriate return value of the maximum and minimum operations given a quiet NaN (qNaN) as one of the input values depends on the circumstances for each point of use. Sometimes qNaN is the appropriate result, sometimes the non-NaN argument is the appropriate result. Therefore, two variants each of the floating point maximum and minimum operations are specified here, and the programmer can decide which one is appropriate to use at each particular place of usage, assuming both variants are included in the binding.

IMO that leaves the door open to the interpretation you're suggesting but you can only do that by adding another set of min/max functions; changing the existing max/min would be a NUC change. At the very least, that's a discussion that should be moved back to Discourse.

ridgeworks: > Which is also why I would like to see max(NaN,1.0) to be NaN.

And as I just discovered, this is hardly a new discussion. If you have the patience (not a realistic expectation on my part) see:
https://discuss.python.org/t/nan-breaks-min-max-and-sorting-functions-a-solution/2868

ridgeworks: > I do understand the efficiency claim. To me it smells as wrong though. You have e.g., max(2,2.0). If this was just 2.0, you'd round outward, so we get (1.9,2.1) (well, boundaries are just an ulp, but for the argument this writes easier). How can you now claim max(2,2.0) to be 2? It should be (2,2.1), no?

No. Outward rounding is part of the calculation, so it's done prior to intersection. The max operation is done with a value that's already rounded outward. 

> To me, that is also the reason why we can't do rational comparison. The outward rounded float may contain the rational.

Bounds are precise numbers, rational or float. It's the interval that captures the uncertainty. And there aren't any "outwardly rounded floats" (only outwardly rounded calculations as I keep trying to say).
```
?- L is roundtoward(1.1,to_negative), H is roundtoward(1.1,to_positive).
L = H, H = 1.1.
```
And we can do rational comparison because it has nothing to do with rounding, unless you insist that anything non-float must be converted to a float before comparison (which was a mistake, according to O'Keefe).

> A similar argument holds for NaN: If we assume there are two NaNs, being either "completely unknown" or "can't exists" When completely unknown, max(NaN, 1) should be (1,Inf). When "can't exists" it, as you claim further down, depends on the context. Blindly continuing with the 1 will give you an answer that is possibly wrong. Continuing with a NaN gives no answer.

There do seem to be two *interpretations* of NaN - 'missing value' and 'propagate' and the computer science community seem to say it depends on context (i.e.., the application - see ISO LIA standard). The 2019 IEEE standard seems to have aligned with this view whereas earlier versions only defined the "missing value" semantics.

As to the value of the "missing value" option let me defer to Kahan from his 1996 notes. The first point he makes:
> Were there no way to get rid of NaNs, they would be as useless as Indefinites on CRAYs; as soon as one were encountered, computation would be best stopped rather than continued for an indefinite time to an Indefinite conclusion. That is why some operations upon NaNs must deliver non-NaN results.  Which operations?

The discussion then makes a case for comparisons, some cases of **, as well as min/max (the latter still under discussion in 1996).  He then goes on to provide a specific example: root-finding (an instance of a search algorithm):

>In general, a root-finder that does not know the boundary of an equation's domain must be doomed to abort, if it probes a wild guess thrown outside that domain, unless it can respond to NaN by retracting the wild guess back toward a previous guess inside the domain.

I think this is analogous to constraint propagation in clpBNR - a constraint calculation which produces a NaN does not invalidate the "previous guess"; instead the current interval bound is maintained by culling the NaN using max/min with missing value semantics.

But this PR is not about max/min NaN semantics. It's about a preference for rationals over floats in maxr/minr. I think I have a credible use case for preferring rationals over floats, given they are precisely equivalent. So let me flip the script and ask: is there a comparable scenario where a float is preferred over a rational? (You have tried to make the case that it shouldn't matter.)

ridgeworks: > That does make sense. It mostly makes sense in your application though were, because your float is actually an upper/lower boundary, you can consider the float a precise value.

This will probably be an ongoing source of disagreement, but I don't want it to hold up the PR. I think it's a bigger issue than just my application, because floats are what they are. If a programmer chooses to attach additional meaning, like imprecision, that's their right, but that shouldn't be baked into the language semantics. 

And I appreciate there is reasonable grounds for two versions of min/max with respect to the semantics of NaN's; that's been recognized by the standards committees even if it hasn't fully percolated down to the language designers. I have no problem with changing the current min/max to implement propagate semantics as long as minr/maxr implement missing value semantics. This gives the programmer two criteria to base their decision of which version to use - correct comparison and Nan propagation. The only thing they can't do is mix and match, but perhaps that's getting into a such a remote corner case, that an explicit Prolog implementation isn't a big deal. (We're just fretting about efficiencies here.)

> If you agree to all that, I can take care of the PR and make the rest of the changes myself.

Yes, I think we have agreement (at least on the practical side of things). It's been an interesting discussion.

ridgeworks: If we proceed on this basis, I should modify the doc for mar/minr to include NaN behaviour. I'll let you know when this is included in the PR.

ridgeworks: Doc now modified. Feel free to change it if, for example, you want to refer to modified max/min.

JanWielemaker:
Ok.  Squashed most of your commits, some layout and doc changes (minor).   Changed NaN forwarding of min/max and use cmpReals for standard order of terms.    Hopefully everybody can live with the current status.   Thanks for the patience and work.

PR #1105, opened 2023-01-26
Added some details about PL_raise_exception

kamahen:


man/foreign.doc
```diff
@@ -3164,7 +3166,14 @@ print_message/2 through the C interface.
 
 \begin{description}
     \cfunction{int}{PL_raise_exception}{term_t exception}
-Generate an exception (as throw/1) and return \const{FALSE}. This
+Generate an exception (as throw/1) and return \const{FALSE}.
```
JanWielemaker, line 3169: I don't like this level of detail.   Preserving exceptions is a difficult topic that has changed several times over the years.   Besides, this only covers the exception `term_t`.    The underlying term also needs protection.

JanWielemaker:
Merges the other part and updated the claim that if there is already an exception pending PL_raise_exception() does nothing.   Exceptions are ordered in urgency and the most urgent is kept.

PR #1108, opened 2023-01-30
feat: add wasm bundle with no data file

jeswr:
Adds [this suggestion](https://github.com/SWI-Prolog/swipl-devel/pull/1088#discussion_r1064262924) from #1068 .

The bundle without the data file is good for use with pre-generated `pvm` images such as [here](https://github.com/eyereasoner/eye-js/blob/a8c288858de7d8a3bab9b47ec6c63fbacb67fddb/lib/transformers.ts#L14-L21).

The resultant file should be approximtely 3.4MiB


cmake/EmscriptenTargets.cmake
```diff
@@ -69,3 +69,14 @@ target_link_libraries(swipl-bundle libswipl)
...
 	     ${POSTJS} ${PREJS})
+
+set(WASM_NO_DATA_BUNDLE_LINK_FLAGS
+    -s SINGLE_FILE)
+join_list(WASM_NO_DATA_BUNDLE_LINK_FLAGS_STRING " " ${WASM_NO_DATA_BUNDLE_LINK_FLAGS} " " ${WASM_SHARED_LINK_FLAGS})
+add_executable(swipl-bundle-no-data ${SWIPL_SRC})
+set_target_properties(swipl-bundle-no-data PROPERTIES
+		      LINK_FLAGS "${WASM_NO_DATA_BUNDLE_LINK_FLAGS_STRING}")
+target_link_libraries(swipl-bundle-no-data libswipl)
+add_dependencies(swipl-bundle-no-data wasm_preload)
```
JanWielemaker, line 80: If we have no data, why do we need the preload?

jeswr: Indeed it does nothing - have now removed that line.

JanWielemaker:
Thanks.   Merged (squashed) after some layout changes and added comments.    This seems the ideal solution for distributing applications.    I guess there is no easy way to bundle a saved state into it, i.e., use the bundle-no-data
and a saved state to create a stand alone bundle?

Would you mind checking out `man/wasm.doc` (unlike the name suggests, this is LaTeX that is first pre-processed) and add a section on how to use the no-bundle version?    I'm also be happy with a plain text description.

jeswr:
>  I guess there is no easy way to bundle a saved state into it, i.e., use the bundle-no-data and a saved state to create a stand alone bundle?

In eye-js we have [this build step](https://github.com/eyereasoner/eye-js/blob/main/scripts/generate-pvm.ts) which generates the image and then embeds it in a file as part of the library. Which means it gets embedded in any webpack distribution of the library like https://eyereasoner.github.io/eye-js/latest/index.js.

My suggestion would be to create a generic utility for this image generation that is part of https://github.com/rla/npm-swipl-wasm; and another generic utility for starting up `SWIPL` with an image (basically just [this function](https://github.com/eyereasoner/eye-js/blob/daf998e29f798ec94bc6d5aab572f1529c920fdf/lib/transformers.ts#L14-L21) where the `EYE_PVM` image is parameterizable).

Then in the `man/wasm.doc` we can essentially say "use these utilities to generate and load the images".

PR #1109, opened 2023-01-31
DOC: tweak WASM port introductory text

eshelyaron:
This PR is a small suggestion for adjusting the introduction section of the WASM port manual to put less focus on the implementation and more on the potential usage (running SWI-Prolog in the browser).


JanWielemaker:
Better :smile:

PR #1119, opened 2023-02-09
Improve foreign.doc + add printf format checking

kamahen:
2 separate commits.

I'm not 100% sure about the change to foreign.doc that removes the call to PL_copy_term_ref() - but the code seems to work fine.

I'll make a separate PR with changes to test cases in packages/cpp, also a change to package/pcre that fixes its blob save/load functions.

### DIFF
--- a/man/foreign.doc
+++ b/man/foreign.doc
@@ -282,7 +282,8 @@ pl_mypredicate(term_t a0, term_t a1)
 Create a new term reference and make it point initially to the same term
 as \arg{from}. This function is commonly used to copy a predicate
 argument to a term reference that may be written. Raise a resource
-exception and returns \exam{(term_t)0} on failure.
+exception and returns \exam{(term_t)0} on failure. An example of its use
+is given below, in the sample code \exam{pl_write_atoms()}.
 
 \cfunction{void}{PL_reset_term_refs}{term_t after}
 Destroy all term references that have been created after \arg{after},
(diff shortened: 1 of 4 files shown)


JanWielemaker:
Thanks.   Merged.   Please make sure patches do not include submodule updates.   They are a bit hard to get rid of again.

kamahen:
My apologies about the submodule updates (I did "git commit -a" without thinking instead of "git commit <files>". Next time this happens, please reject that PR and I'll make a new PR.

man/foreign.doc
```diff
@@ -1906,33 +1907,40 @@ PL_unify_functor().
...
-(the code based on PL_put_list() is probably slightly faster). If the
+(the code based on PL_cons_list() is probably slightly faster). If the
 argument is bound, the code below may fail before reaching the end of
 the word list, but even if the unification succeeds, this code avoids a
 duplicate (garbage) list and a deep unification.
 
 \begin{code}
 foreign_t
 pl_get_environ(term_t env)
-{ term_t l = PL_copy_term_ref(env);
```
kamahen, line 1922: I'm curious why this (rather old) code has PL_copy_term_ref() ... is it because when you wrote the original code, you were going to use PL_cons_list() on `l` or is there something more subtle, where the PL_copy_term() is actually needed? (Code examples in ffi_test.c -- ffi_get_environ3_() and ffi_write_atoms_(), which correspond to examples  put_list() and pl_write_atoms() in foreign.doc.)

JanWielemaker: Oops.   I should have read more carefully.   The PL_copy_term_ref() is needed as you are not allowed to do anything but unify against the argument term references.   PL_unify_list() unifies, but also overwrites where the term reference points at through the 3rd argument, so this is not allowed.   What actually happens is that you destructively overwrite the predicate argument.

I'd have to think about a use case where this actually breaks things.   That might not exist in this case because the argument reference of the foreign predicate is a reference pointer to the argument used in the parent environment, so after the first cell, the one in the parent environment still points to the whole list and the one in the foreign predicate points at the tail.   It probably only upsets stack inspection in this case.   Just for purity, it shouldn't be used.

Note that the `term_t` you get as argument is a direct reference into the environment of the predicate.   That is unlike `term_t` created using `PL_new_term_ref()`, etc, which are term references that are allocated outside the normal Prolog predicate environments and these may thus safely be overwritten and reused.

I guess you want to update both the code and document this?

kamahen: Yes, I'll update both the code and document this. (Hopefully, later today)

PR #1124, opened 2023-02-13
Fix typos + fix example code

kamahen:
The example code could have left a dangling string buffer. (C++ is so much nicer in this respect, because its destructors can be set to do cleanup). 
packages/cpp/test_ffi.{c,pl} have the modified code and test cases - please take a look (the test code also fixes a potential missing PL_release_stream()).


JanWielemaker:
Thanks.   write atoms now got so ugly that I changed it a bit.   Basically by making explicit that PL_get_nil_ex() propagates the existing exception, so we do not need the complicated stuff at the end.   This used to happen anyway because PL_raise_exception() keeps the existing exception unless the new one is more urgent, which is not the case.

Hope it is right now.    Surely the docs get better every day :smile:

PR #1125, opened 2023-02-17
ENHANCEMENT: profiler - improve port count accuracy

ridgeworks:
Intent is that port counts should reflect events on standard box model ports. This means no internal failures/backtracking, e.g., head unifications, inline "predicates" etc., are visible to the profiler.

External interface/display not affected so existing code is not affected by this change. Review of profiling doc probably required moving forward.


JanWielemaker:
Thanks.  Merged after squash and some minor changes (made profFailToCP static and avoid C recursion).
It is surely a lot better.    More analysis is probably needed to figure out whether all cases are ok.   For profiling purposes this probably doesn't matter too much.    We might consider adding shallow backtracking to the stats (failed head unification).

PR #1126, opened 2023-02-18
fix: make prolog compatible with emsdk 3.1.31

jeswr:


JanWielemaker:
Merged after adjusting commit message.  Thanks.

PR #1128, opened 2023-02-22
Implement OpenBSD alphabeta for base64 encoding

jamesnvc:
Needed for bcrypt implementation in packages/ssl (see https://github.com/SWI-Prolog/packages-ssl/pull/169).


jamesnvc:
It should be good now, those force pushes were just some cosmetic issues I noticed right after creating the PR.

JanWielemaker:
Merged after update of commit message and merging the two new clauses.   This should be faster and create fewer choicepoints.

PR #1129, opened 2023-02-27
Clarify behavior when multiple exceptions occur

kamahen:
The footnote is derived from an example that @JanWielemaker gave in the forum.
Also fixes one unrelated example.


kamahen:
This PR has been replaced

PR #1131, opened 2023-02-27
DOC: add a few background papers

kamahen:
I did a build and didn't get error messages, but don't know if it actually processed this.

Also, I've put 2 URLs for one entry (not 100% sure of the "source of truth"); and there were multiple publication dates for Aït-Kaci's tutorial on WAM.


JanWielemaker:
Merged after fixing some syntax issues.    Note that this has no effect to the ouput as long as there are no `\cite{}` commands using this.

PR #1132, opened 2023-02-27
DOC: Clarify exceptions behavior

kamahen:


JanWielemaker:
Made some changes.  We can't have PL_get_nchars() there twice.   There is also a pending issue about the PL_get_chars() and other 0-terminating functions: I think they should check that the input does not hold 0 bytes, so we effectively truncate.    Also changes the mark/release strings remark a little.    Revered the changes to pl_write_atoms().   Not only was it incorrect (stream was not initialised), but I think it is wrong to complicate what is supposed to be a simple example with the complications of stream handling.

Issue #1120, opened 2023-02-09
Make some functions in pl-wic.c public


kamahen:
Possibly we should make some of the low-level functions of pl-wic.c public (as PL_*). I.e., we do have the stuff for saving atoms and integers there that takes care of all issues related to that, such as encoding, big/little endian, different sizes, etc. Maybe just for atoms and integers, at least initially.

These can be tested with the save/load functions for packages/pcre.

Please assign this issue to me.

(This is in response to https://github.com/SWI-Prolog/packages-pcre/pull/27#issuecomment-1423823764)

kamahen:
@JanWielemaker - what does "wic" mean? Is it an old name for "qlf"? 
Reason for asking: I'll want to put some kind of uniform prefix on the public functions, e.g. `PL_qlf_putString()` (or `PL_wic_putString()`).

JanWielemaker:
Warren Intermediate Code or something like that.  As the VM is not a WAM, it is a bad name.   qlf is at least not as bad (Quick Load File/Format).   I wouldn't mind a better name.    The file was called like this from day 1 of SWI-Prolog (before it actually had a name :smile: )

kamahen:
Also need to update foreign.doc. There is some preliminary stuff in https://github.com/SWI-Prolog/swipl-devel/pull/1122

kamahen:
Leaving this open because not all functions have been exported

JanWielemaker:
You probably notably want to add atoms.   That saves all the encoding issues and atoms are saved only once in .qlf files,
saving time and space.    After that, I think we can close.   If there is a serious demand we can always add more.

kamahen:
`putAtom()` has a parameter `wic_state*` (`getAtom()` just has `IOSTREAM*`). From tracing the code, it seems that the following fields of `wic_state*` are used:
* `wicFd` (`IOSTREAM*`)
* `idMap` (table of mapped identifiers)
* `savedXRTable`

It's not clear to me how these can be disentangled without changing the "save"/"load" API.

kamahen:
It appears that `state->idMap` can be `NULL`. However `saavedXRTable` is more problematic -- it's used (indirectly from `saveXRBlobType()` in `savedXR()`.

kamahen:
[...]

PR #1134, opened 2023-02-28
Make some functions in pl-wic.c public

kamahen:
Bug https://github.com/SWI-Prolog/swipl-devel/issues/1120
A test has been added to test_ffi.c, which will be in a separate PR.
TODO: make a test that verifies the read/write functions work across different architectures (big-endian, little-endian)


JanWielemaker:
Thanks.  Merged as is (but after local rebase).

Issue #1133, opened 2023-02-28
PL_atom_to_encoding() in wrong header file?


kamahen:
`PL_atom_to_encoding()` and `PL_encoding_to_atom()` in `SWI-Stream.h` are inside `#ifdef _SWI_PROLOG_H`, which means that `SWI-Prolog.h` is required to be included before `SWI-Stream.h`.

However, `foreign.doc` says in the documentation for `\cfunction{int}{write}{IOSTREAM *s, atom_t a, int flags}`:
```
The \arg{flags} are a bitwise
\emph{or} of zero or more of the \const{PL_WRT_*} flags defined in
\file{SWI-Prolog.h}. This prototype is available if the
\file{SWI-Stream.h} is included \emph{before}
\file{SWI-Prolog.h}.
```
One of these is wrong.
(There's also one code example in `foreign.doc` that has `SWI-Stream.h` included before `SWI-Prolog.h`.

Ideally, the order of includes shouldn't matter.

kamahen:
Whatever decision is made for this, there are also a few functions in pl-wic.c that will need to be treated similarly for issue https://github.com/SWI-Prolog/swipl-devel/issues/1120, such as
`PL_qlf_putStringW(const pl_wchar_t *s, size_t len, IOSTREAM *fd);`
because of `pl_wchar_t`.

JanWielemaker:
Making the order irrelevant is probably best.     I guess that may involve a helper to get the common definitions in place?

kamahen:
Should I create SWI-Prolog2.h, which is included by both SWI-Prolog.h and SWI-Streams.h and contains (for now) just `atom_t`?

JanWielemaker:
If that is all that is needed, just duplicate the definition and add some macro such that they are not defined twice.   That seems easier to manage than adding a file that must be handled by installers, etc.  Something like can simply appear in both files.

```
#ifndef PL_COMMON_TYPES_DEFINED
#define PL_COMMON_TYPES_DEFINED
typedef ...
...
#endif
```

kamahen:
Please assign this to me and I'll fix it as part of the "Make some functions in pl-wic.c public" work (PR https://github.com/SWI-Prolog/swipl-devel/pull/1134 and follow-ons)

JanWielemaker:
Thanks.

JanWielemaker:
Closing as fixed.

PR #1137, opened 2023-03-03
SWI-Stream.h and SWi-Prolog.h can be in any order (issues 1120, 1133)

kamahen:
Start work on making putAtom() in pl-wic.c public. 
Fix a couple of names that were mistakenly made part PL_qlf_*().

This is a bit of a hack, but a proper fix looks like quite a bit of work for very little benefit.

https://github.com/SWI-Prolog/swipl-devel/issues/1133
https://github.com/SWI-Prolog/swipl-devel/issues/1120


kamahen:
This PR is obsolete (and partly wrong). I will try making another PR for these bugs later.

PR #1140, opened 2023-03-07
Fix some printf formats

kamahen:


src/pl-index.c
```diff
@@ -2046,7 +2046,7 @@ listIndexGenerations(Definition def, gen_t gen)
 	    for(cr=cl->first_clause; cr; cr=cr->next)
 	    { Clause clause = cr->value.clause;
 
-	      Sdprintf("  %p: [%2d] %8u-%10u%s%s\n",
+	      Sdprintf("  %p: [%2d] %8" PRIuMAX "-%10" PRIuMAX "%s%s\n",
```
JanWielemaker, line 2049: Do we have support for PRIuMAX in Sdprintf?    Better yet, use generationName(), which was added when generations became sometimes huge numbers when in a transaction.

JanWielemaker:
All generations should probably be handled using generationName()?

src/pl-index.c
kamahen: Done. Please check that this is the correct way to call generationName() -- it appears to allocate something on the stack and I want to ensure that this is a valid thing to do here. (also at line 2060 and pl-proc.c:4118)

kamahen:
I have used `generationName()` in `pl-index.c` and `pl-proc.c` ... there might be other places that should use it, but - if so - they didn't trigger a format mismatch.
Please verify that these are valid places to use `generationName()`, which contains a call to `buffer_string(..., BUF_STACK)`.

JanWielemaker:
Thanks.    Merged after a few more edits.

PR #1142, opened 2023-03-09
Add options 'ports' and 'sample-rate' to profile/2

ridgeworks:
See commits for details.

Note that changes for Windows haven't been tested (or built). More testing in the wild on any platform would be useful.

I'll work on updating the documentation if this all looks good.


JanWielemaker:
Thanks.  Merged after various edits.    Also updated GUI to show only one column if ports(false) is used.   Text output should also be corrected as this is a little weird with ports(false).   Possibly we should also change counters to be int64_t rather than uint64_t, so we get the negative ones.    int64 doesn't overflow anyway.

PR #1145, opened 2023-03-12
PORT: msys2, fix linker problem with ws2_32

mgondan:


mgondan:
This does not fully solve the problem. Closing it.

PR #1153, opened 2023-03-27
Doc fixes + WUNUSED for some query-related functions

kamahen:


kamahen:
I turned off WUNUSED for `PL_call_predicate()` because it produced warnings that aren't turned off by using `(void)` [there's a long and pedantic discussion in the gcc mailing lists, where the gcc people don't seem to realize that there are uses for gcc outside the Linux kernel).

I've added '(void)` to some calls to `PL_call_predicate()` where the return code isn't checked -- I'm guessing that these should have a check but because I'm not sure, I just marked the calls.

kamahen:
BTW, I haven't tested the cmake changes on clang or MacOS. My changes were nothing more than copying the C flags - and they appear to have worked for C++ debug, at least. However, some things might need to be changed - e.g., `WUNUSED` seems to be ignored by `g++` ... that might not be important because I'm adding `[[nodiscard]]`s to SWi-cpp2.h.

JanWielemaker:
Thanks.   Removed docs from PL_get_term_value().   It is an old interface that should not be used.   It is still in use for xpce AFAIK.

> I've added '(void)`to some calls to`PL_call_predicate()` where the return code isn't checked -- I'm guessing that these should have a check but because I'm not sure, I just marked the calls.

Removed the `(void)`.  It doesn't work for WUNUSED.   The calls are correct.   One has PASS_EXCEPTION, but it is called from a signal handler and the signal wrapper will use PL_exception(0).    The others do not specify exception handling, which means Prolog will take care of that before returning and we do not care whether the call succeeded.

Note that with WUNUSED it is rather hard to ignore the value anyway.    Possibly it is too string for PL_call_predicate() as ignoring is typically a bad idea, but there are exceptions.

Issue #1165, opened 2023-04-14
PL_record_external() doesn't support blobs


kamahen:
Presumably this could be added now, using Sopenmem() and the blob save/load interface.

kamahen:
Added to the documentation with https://github.com/SWI-Prolog/swipl-devel/pull/1170

PR #1170, opened 2023-04-26
DOC: limitation of PL_record_external() with blobs

kamahen:


kamahen:
Related to Issue https://github.com/SWI-Prolog/swipl-devel/issues/1165

PR #1173, opened 2023-05-20
DOC: details about building packages

kamahen:


kamahen:
I noticed that `prolog_pack:pack_install_from_local/4` uses `relative_file_name/3` to generate the link that's used with `pack_install(.)` ... this results in a symlink like
```
/home/peter/.local/share/swi-prolog/pack/rocksdb -> ../../../../src/rocksdb/
```
... wouldn't it be better to use `absolute_file_name/2` for the symlink? That is, the link would become
```
/home/peter/.local/share/swi-prolog/pack/rocksdb -> /home/peter/src/rocksdb/
```

JanWielemaker:
> wouldn't it be better to use `absolute_file_name/2` for the symlink? That is, the link would become

Absolute and relative links are probably as dynamic and static linking :smile: They both have there advantages and disadvantages and some people are rather religious about them.     In this case, I'd say that is is likely that the package itself is checked out somewhere in the homedir of the user and the link comes from the homedir, so a relative link makes (some) sense.

kamahen:
About absolute vs relative links from `pack_install(.)` ...

If the source code is kept in a separate file system (this is how Google does things, for example; and probably they're not unique in this), then the relative links go all the way up to `/`, so for this, an absolute file name would be best.

If the `swipl` executable is in `/usr/bin`, then the relative link also goes all the way up to `/`.

And if the `swipl` executable is in `/home/<user>/.local/bin`, then the relative path will typically be longer than an absolute path.

So, for all three use cases that I can think of, an absolute file name is better. Or - we could compromise and use whichever is shorter, which will usually be the absolute file name. :)

JanWielemaker:
The normal place for packs is `~/.local/share/swi-prolog/pack/<name>`.   The installation location of Prolog is not relevant.

I forgot why I used a relative link.   Guess I had a reason as it is more work.    As long as there is no reason to really prefer absolute links I prefer not to change things ...

PR #1180, opened 2023-06-05
ENHANCED: indicate DCG rules in xref_defined/3

eshelyaron:
Add a new predicate `xref_grammar_rule/2`, allowing consumers of `library(prolog_xref)` to find out if a predicate is a DCG grammar rule.

Determining whether a predicate is a DCG non-terminal is important, for example, when adding a that predicate to the module's export list (if it's a non-terminal we want to add `foo//N-2` rather then `foo/N` to the export list).

### DIFF
--- a/library/prolog_xref.pl
+++ b/library/prolog_xref.pl
@@ -127,7 +127,7 @@
     (mode)/2,                       % Mode, Src
     xoption/2,                      % Src, Option
     xflag/4,                        % Name, Value, Src, Line
-
+    grammar_rule/2,                 % Head, Src
     module_comment/3,               % Src, Title, Comment
     pred_comment/4,                 % Head, Src, Summary, Comment
     pred_comment_link/3,            % Head, Src, HeadTo
@@ -530,6 +530,7 @@ xref_clean(Source) :-
     retractall(uses_file(_, Src, _)),
     retractall(xmodule(_, Src)),
     retractall(xop(Src, _)),
+    retractall(grammar_rule(_, Src)),
     retractall(xoption(Src, _)),
     retractall(xflag(_Name, _Value, Src, Line)),
     retractall(source(Src, _)),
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Distinguishing a normal predicate from a grammar rule is surely useful.    I think it is more in line with the rest of the interface to extend xref_defined/3 rather than adding xref_grammar_rule/2.

kamahen:
`//` is good for documentation, but other than that is there any difference between `foo//2` and `foo/4`?

(I'm not asking just to be annoying ... there are other transformations than DCGs (e.g., EDCGs), and we don't do anything special for them. Also, in future, we might want to add `==>`, analogous to `-->` but with the SSU semantics, and might want to consider whether we want special handling for that - I'm guessing not.)

eshelyaron:
Your question is far from being annoying, this PR is indeed not a general solution for making `library(prolog_xref)` aware of code transformations, but I think DCGs are common enough to deserve special treatment.

A more general solution might involve a hook that allows packages implementing code transformations to extend `xref` such that transformed terms are annotated with a transformation-specific indication.

JanWielemaker:
> Would we want `xref_defined('/foo/bar.pl', baz(_, _), How)` to be deterministic and succeed only with `How = grammar_rule(Location)`? Or should it be an addition to other (existing) solutions?

A predicate may be defined several ways.   So, adding a `How = dcg` is just fine.   It is unlikely to break anything.    That also covers @kamahen's issue: in the end it is a predicate with a name and arity.    And yes, DCG has a bit of a special status :smile:

eshelyaron:
Alright, I (force) pushed an updated version and changed the PR title accordingly.

JanWielemaker:
Thanks! Merged, including the doc addition.    Changed is/2 to =:=/2

PR #1183, opened 2023-06-12
Added a few details about equality and blobs

kamahen:


kamahen:
This also adds a few citations (using `\cite{...}` but they produce warning messages ... does something else need to be run first, to update the bibliography?
```
No bibliography entry for "Clocksin:83"
No bibliography entry for "AÃ¯t-Kaci:99"
No bibliography entry for "Appleby:86"
```

kamahen:
I think that this PR is obsolete; if it is, I'll delete it.

JanWielemaker:
This is handled (except for the citations as I noticed).    There is somewhere a section with some literature pointers.  Possibly that can be improved.    Low priority as far as I'm concerned.

PR #1189, opened 2023-08-07
Clarify some details of blob and I/O functions

kamahen:


JanWielemaker:
Thanks.   Further fixed and clarified some things :smile:

PR #1194, opened 2023-08-20
BF: add wrapper for free()

mgondan:
avoid non-portable realloc(ptr, 0)


mgondan:
libbf.h, line 218:
s->free_func(s->realloc_opaque, ptr, ***0***);

The wrapper function for free() requires a size argument, I was unsure what to do with it.

JanWielemaker:
Thanks.  I'll have a look next week as I have no quick access to my main testing machines right now.   Looks good.

JanWielemaker:
Thanks.   Applied as is, but after squashing the multiple commits.

PR #1196, opened 2023-09-16
Documentation improvements to blob, IO, and a few predicates

kamahen:
When I did a build, I got these errors:
No bibliography entry for "Clocksin:83"
No bibliography entry for "AÃ¯t-Kaci:99"
No bibliography entry for "Appleby:86"

### DIFF
--- a/man/foreign.doc
+++ b/man/foreign.doc
@@ -353,14 +353,23 @@ are blobs that represent textual content. Textual content is also
 represented by Prolog string (see \secref{string}), which makes the
 general notion of \jargon{string} in Prolog ambiguous. The core idea
 behind blobs/atoms is to represent arbitrary content using a
-\emph{unique} handle, such that comparing the handle is enough to prove
-equivalence of the content; i.e., given two different atom handles we
+\emph{unique} handle, such that comparing the handles is enough to prove
+equivalence of the contents; i.e., given two different atom handles we
 know they represent different texts. This uniqueness feature allows the
-core engine to reason about atoms without considering their content.
+core engine to reason about atom equality and inequality without considering
+their content.  Blobs without the \const{PL_BLOB_UNIQUE} feature are
+also tested for uniqueness without considering their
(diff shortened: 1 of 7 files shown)


man/foreign.doc
```diff
@@ -951,6 +969,42 @@ advised to use the *_chars() or *_nchars() functions whenever
...
+    \const{stdin}, \const{stdout}, \const{stderr}, you can use
+    \const{Sinput}, \const{Soutput}, \const{Serror}.
+\end{itemize}
+
+In general, if a stream is acquired via PL_acquire_stream(), an error
+is raised when PL_release_stream() is called, so in that situation,
+there's no need to check the return codes from the IO functions. Blob
+write callbacks are also called in the context of an acquired stream,
+so there is no need to check the return codes from its IO function
+calls.  However, if you use one of the standard streams such as
```
kamahen, line 997: Is  this correct? I've added checks for return codes in the example code snippets, on the assumption that an IO error also sets qf->exception or similar.

JanWielemaker: It doesn't set qf->exception.   PL_release_stream() should set the default context exception, which means it binds a global internal `term_t` reserved for exceptions.    The qf->exception is to deal with exceptions when using PL_open_query() and friends.

JanWielemaker:
Merged after some minor edits.   Sdprintf() uses stderr.   Removed the references to VM background material.  I don't really see the contribution.   Most of the design of the foreign interface comes from the Quintus/SICStus/Ciao WAM based implementations and they apply just as well for the zip.   Whether or not there are registers is not very relevant.    Thanks!

man/foreign.doc
kamahen: But what about errors happening on `Suser_output` etc.? Should the programmer do `PL_acquire_stream(Suser_output)` in the foreign function? That might be good practice anyway, to ensure the stream is locked so that multithreading doesn't interleave outputs.

(Which reminds me: I'll add an RAII version of PL_acquire_stream())

PR #1199, opened 2023-10-01
DOC: clarify some details of stream error handling

kamahen:


JanWielemaker:
Thanks.   Merged.   Modified the use of PL_acquire_stream() following by PL_release_stream() for checking errors to use Sferror() to check for errors and then use this pair to raise a Prolog exception.

PR #1202, opened 2023-10-22
Fix typos, add an xref

kamahen:


JanWielemaker:
Thanks.   Removed hint at PL_get_compound_name_arity(t,NULL,NULL).   It is not true (it is the same as PL_is_compound(); PL_is_functor() checks that a terms is compound and has a particular functor) and I find it mostly confusing.   PL_get_size_ex() reference to PL_get_intptr() is a bit dubious.   Probably there should be a PL_get_size(), although I don't think there are any sensible cases for using it as you almost always want the type and domain check.    The C standard does not guarantee size_t and uintptr_t to be the same (although they are on almost all platforms).

PR #1212, opened 2023-12-22
Use recommended way for checking big/little-endianness

mgondan:


JanWielemaker:
Seems this needs some more careful checking.   Both setting these macros and using them seems pretty outdated.   The new CMake macro is nice, but only provided since CMake 3.20.   We must remain compatible with much older versions for long term Linux releases.  Currently 3.5 for most of the systems, 3.9 if you also want to build the ODBC interface.

Anyway, this first needs some study.    I'd also like to know what defender is complaining about.   Is that CMake trying to read (search) the .exe file?    In that case, using static libraries is a defensible strategy as I can imagine that searching `.exe` files is considered suspect behavior.  If it is just that this specific executable is a false positive these programs should be reported as false positives with Microsoft.

mgondan:
The idea of the patch is to use the new CMake macro only if it is defined, therefore the if(DEFINED CMAKE_C_BYTE_ORDER). If it isn't, we fall back to the old behavior and use CMake's TEST_BIG_ENDIAN macro. I can also add the SET(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY) around TEST_BIG_ENDIAN, if you wish (I'll update the PR to include both).

The Windows Defender message is raised when the temporary file is copied,
````
-- Check size of unsigned short
CMake Error at D:/rtools43/x86_64-w64-mingw32.static.posix/share/cmake-3.24.3/Modules/CheckTypeSize.cmake:146 (try_compile): Cannot copy output executable 'D:/temp/RtmpYrk62w/R.INSTALL1a848615c393b/rswipl/mingw64/CMakeFiles/CMakeTmp/cmTC_db78b.exe' to destination specified by COPY_FILE: 'D:/temp/RtmpYrk62w/R.INSTALL1a848615c393b/rswipl/mingw64/CMakeFiles/CheckTypeSize/CMAKE_SIZEOF_UNSIGNED_SHORT.bin'
````
You find the full build log here (for a few days): https://win-builder.r-project.org/y6aEdKK3D2rT/00install.out

JanWielemaker:
Ok.   It is as good as it gets :cry: Merged after squashing the two static lib patches and rewording the commit message to deal with the changelog.

Thanks!

PR #1226, opened 2024-01-29
Maintenance: updated vendored minizip code to the version from zlib v1.3.1

jayaddison:
This pull request brings the vendored `minizip` code -- distributed currently as part of [`zlib`](https://github.com/madler/zlib/) -- in SWI-Prolog up-to-date with the recently-released v1.3.1 package.

The goal is to bring that code closer inline with upstream, making it simpler to track and apply future updates and patches.

In detail:

* Determined the [existing `minizip` baseline to be v1.2.11](https://github.com/SWI-Prolog/swipl-devel/blob/7b050ce95fb6d9bb3942403935e2dfa8c8735144/src/minizip/configure.ac#L4).
* Cherry-picked commits from `7d60b86782f96854974fdec6f17cb95cfaf020a6...51b7f2abdade71cd9bb0e7a373ef2610ec6f9daf` -- i.e. next-commit-after v1.2.11 up-to-and-including v1.3.1 -- of `zlib.git`
* Omitted https://github.com/madler/zlib/commit/aa154e3da0d98629bbefe79cdfba8bc7178dc4ad because I wasn't sure how to apply it correctly.
* Reverted c794af0f61da8dd263c986d9f78b1b09f7792403 because the `OF` macro is no-longer-required.
* Backported https://github.com/madler/zlib/commit/bf2578be2d0566dc3ee47655d56b62063fe6bae6 to the `unzClone` function.

I've checked that the resulting code compiles and that `make test` passes.

Two areas where I'm not confident about merge-conflict resolution:

* `local` / `static` / `extern` declarations on some functions (`zipOpen3` for example).
* 1c1299afbddbd8f2b02017285be7fc6ba86973c9 (the largest single commit re-applied, requiring some edits).

### DIFF
--- a/src/minizip/MiniZip64_Changes.txt
+++ b/src/minizip/MiniZip64_Changes.txt
@@ -1,5 +1,5 @@
 
-MiniZip 1.1 was derrived from MiniZip at version 1.01f
+MiniZip 1.1 was derived from MiniZip at version 1.01f
 
 Change in 1.0 (Okt 2009)
  - **TODO - Add history**
--- a/src/minizip/configure.ac
+++ b/src/minizip/configure.ac
@@ -1,7 +1,7 @@
 #                                               -*- Autoconf -*-
 # Process this file with autoconf to produce a configure script.
 
-AC_INIT([minizip], [1.2.11], [bugzilla.redhat.com])
+AC_INIT([minizip], [1.2.11.1], [bugzilla.redhat.com])
 AC_CONFIG_SRCDIR([minizip.c])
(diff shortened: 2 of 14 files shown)


JanWielemaker:
Thanks.   Quickly scanned through it, built and tested on Linux, Windows, MacOS and WASM.   All seems fine!

jayaddison:
Thanks @JanWielemaker - yep, I'm reasonably confident it's correct; in any case, I'll follow the issue tracker for any post-release bugreports.

PR #1229, opened 2024-02-02
ENHANCED: Handle C++ functions in doc files'

kamahen:
\cxxfuncref currently is the same as \cfuncref. It was useful for debugging `doc2tex.pl`; and it's possible that the definition will change in future.

This change depends on https://github.com/SWI-Prolog/packages-ltx2htm/pull/5

### DIFF
--- a/man/doc2tex.pl
+++ b/man/doc2tex.pl
@@ -95,22 +95,30 @@ add_result(One, [One|Tail], Tail).
 		 *            RECOGNISERS	*
 		 *******************************/
 
+tr(\cfuncref(FName, Args)) -->
+    pl_c_func_prefix(Prefix),
+    c_identifier(Name), "(", string_without(")", Chars), ")",
+    !,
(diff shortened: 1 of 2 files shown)


man/doc2tex.pl
```diff
@@ -327,6 +341,7 @@
 	prolog_id_cont(CL),
 	{ atom_codes(Name, [C0|CL]) }.
 
+prolog_id_cont(_) --> "(", !, { fail }.
```
JanWielemaker, line 344: Why is this?   `(` is not a Prolog id continue character, no?

kamahen: This is a left-over from when I was using your original ordering of `tr//2` rules.
I ended up re-ordering them, so that the _identifier(...)_ stuff is done before detecting an atom -- otherwise, things such as `PlTerm::unify_blob()` weren't handled correctly.
It appears to be unnecessary (and a cut-fail isn't good style either).

PR #1232, opened 2024-02-06
Improve the example code for foreign frame

kamahen:


kamahen:
Test cases are in packages/cpp, which will be added to https://github.com/SWI-Prolog/packages-cpp/pull/66

man/foreign.doc
```diff
@@ -1963,20 +1963,36 @@ returned \const{FALSE}.
 
 Undoing (partial) unifications is achieved using
 PL_open_foreign_frame() and PL_rewind_foreign_frame(), as shown in the
-snippet below.
+snippet below:
 
 \begin{code}
-    { fid_t fid = PL_open_foreign_frame();
+/* Same as: can_unify(T1,T2) :- unifiable(T1,T2,_). */
```
JanWielemaker, line 1969: The original example is not ideal, but this illustrates another problem.    This code tries to illustrate what is needed if you try to unify an argument against set of terms.   If the first unification fails, you must check for an exception and otherwise rewind.  Then you can start the new unification attempt.   Please put comments in the text around, not in the code fragment.

JanWielemaker (commented):
Picked the typos and missed rational from the PR and added a new example for the partial unification that hopefully illustrates the original intend.

Closing as completed.

PR #1234, opened 2024-02-08
Added markup processing for C++ and non-PL function references

kamahen:
This PR makes the documentation for the C++ API look nicer.
It has two kludges:
- A check for `{END_OF_FILE}` ... I used when trying to track down a markup bug that caused a crash in `latex2html.pl`.
- Special handling of `{{}}` in an arguments list, which caused `packages/swipy/janus.doc` to not be processed properly. The proper solution for this is to recursively process the args, but that appears to require changing other parts of the processing pipeline.

### DIFF
--- a/man/doc2tex.pl
+++ b/man/doc2tex.pl
@@ -76,6 +76,9 @@ doc2tex(InFile, OutFile) :-
         emit(Result, Out),
         close(Out)).
 
+doc2tex([]) -->
+    "{END_OF_FILE}", !, % Useful for finding markup that caused a crash
+    remainder(_).
 doc2tex(Result) -->
     tr(One),
     !,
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.   Pushed some further commits to deal with nested parenthesis in C(++) function arguments (they must match) and recursively call the translation on the arguments rather than using a hack for a single issue.

Removed the END_OF_FILE.    I think that may lead to getting wrong results silently, no?   Seems more like a temporary hack for debugging.

kamahen:
I think that the clause
```prolog
tr(Object) -->
    unquoted_atom(Name),
    !,
    tr_name(Name, Object).
```
needs to go _after_ the `\cfuncref` clauses ... in an earlier PR (which I cancelled), I had left it in that position but had to add a "not followed by '('" test (which you  didn't like), so it was easier to move the clause.

I tried moving it, and prefer the result: for unknown functions, it bolds them but there's no link, of course.

JanWielemaker:
It also turns Prolog calls into C function references, no?     I wonder whether we want that.     That also makes the cxx_identifier() clause that is not protected by pl_cxx_func_prefix/1 rather dubious.     If we do not know, assuming foo(a) to be Prolog makes more sense to me than assuming it is C.   After all, this is for documenting Prolog :smile:  We could make a true c++ identifier, e.g., if it contains :: starts with ~.

So, my proposal would be

   - unquoted atom "(" args ") to become Prolog links (needs some more work to compute the arity.   We could be a bit more restricting and see whether we can actually parse the term as Prolog, so we also have the arity safely.   Then we need a \predref version with three arguments: name, arity and args.
   - Name "(" args ")" to be a C name if name is a valid C identifier (and not a valid Prolog one).    We could also consider Name() to refer to C.

Does that make sense?

kamahen:
I had forgotten that `doc2ltx.pl` is mainly for documenting Prolog, not C/C++. 🤡

At the risk of more backtracking (which doesn't seem to slow things down much), maybe change
```prolog
tr(Object) -->
    unquoted_atom(Name),
    !,
    tr_name(Name, Object).
```
to
```prolog
tr(Object) -->
    unquoted_atom(Name),
    tr_name(Name, Object),
    !.
```
`tr_name/2` is somewhat misnamed -- it only succeeds for `Module:Name/Arity` or `Name/Arity`.

For distinguishing C/C++ calls from Prolog calls, it's trivial if the name starts with upper-case (which `code_type/2` can do, if we test for not-valid Prolog atom first, and we can also check for `::`, as you say (checking for `.` or `->` probably doesn't make sense, because those would be in expressions, handled by `\exam`). For detecting C references, I think it makes sense to consider `foo()` to be a C call and not a call to Prolog `foo/0`; `foo(bar)` is ambiguous, so should default to Prolog but can be overriden by `\cfuncref{foo}{bar}`.

The other thing that `tr_name/2` does is it adds a cross-reference:
```prolog?- doc2tex:phrase(tr(Object), `foo:bar/2`).
Object = [\index("foo:bar/2"),\qpredref(foo,bar,write(2))].

?- doc2tex:phrase(tr(Object), `bar/2`).
Object = [\index("bar/2"),\predref(bar,write(2))].
```
but the indexing for C/C++ calls is done as part of `cfuncref`:
```prolog
?- doc2tex:phrase(tr(Object), `PL_foo(x)`).
Object = \cfuncref('PL_foo',"x").
```
Right now, we can control the automatic transformation by explicitly using `\cfuncref` but there's no equivalent for Prolog - it requires putting in both `\index` and `predref`) ... perhaps `\predref` could be changed to also do the indexing? (I noticed that for C, if the indexing fails, there's no `href` but there is a font change).
This would allow for an explicit override for both Prolog and C/C++ if the automatic translation is wrong.
We might also want an `\asis` for where we don't want to do anything with a string like `foo()`.

JanWielemaker:
Moving the cut doesn't help as tr_name//2 is det.    Pushed 5b7205ebc385fd1f84cf6bc683cc97daaf8ca941 to address some of this.   I guess moving the index calls is a good idea as well.     You don't need an asis.    If you want foo() as normal string,  write `{foo}()`.     As `{}` just delimits a scope for LaTeX, but doesn't do anything otherwise, it is safe to use and avoids re-interpretation.

PR #1243, opened 2024-02-23
janus check first checks if python is found

xpxaxsxi:
Now the failing janus check --when python is not in path-- fails nicely with check_installation/0


JanWielemaker:
Thanks.    Fixed using 8646e23183c77969bcebd55c37b17f489d8b07e8.    Using `shell/1`  is rather dubious as `python` may be named differently or not be on `PATH` or may exist, but we have been compiled against another version of the embedding library, etc.  

The real issue was that shared object loading errors were supposed to be hidden from the user, but were incorrectly detected after several changes on how we load shared objects to facilitate WASM.

PR #1244, opened 2024-02-24
DOC: Clarify details of PL_get_chars() et al

kamahen:


JanWielemaker:
Thanks!  That is a lot cleaner.   Did a few small edits.

PR #1246, opened 2024-02-25
Make documentation match header file for CVT_NUMBER

kamahen:
The definition of CVT_NUMBER in the documentation didn't match the definition in SWI-Prolog.h

I've also made the descriptions of CVT_ATOMIC and CVT_ALL into the same form. (CVT_ALL isn't actually "all of the above" because it doesn't include CVT_INTEGER)


JanWielemaker:
Merged after some more edits.    Looks crisp and clear now :smile:

PR #1249, opened 2024-03-03
Port: ifdefs for Rtools44 

mgondan:
(from CRAN): The upcoming version of Rtools44 has sys/mman.h, which however doesn't have all the Unix functionality of the same header file (e.g. it misses mmap, sysconf). It is from the mman-win32 project.


mgondan:
I haven't run into problems, so far. So I revert this to a draft.

PR #1258, opened 2024-03-25
Find pthread_setname_np/2 or /3

mgondan:
The code is meant to find pthread_setname_np/2 or /3. Unclear what -Werror does here.

### DIFF
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -303,7 +303,6 @@ include(TestRecursiveMutex)
 
 if(HAVE_PTHREAD_SETNAME_NP)
 function(check_pthread_setname_np)
-  set(CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS} -Werror)
   check_c_source_compiles(
       "#define _GNU_SOURCE
        #include <pthread.h>\nint main()


JanWielemaker:
What is the idea?   The `-Werror` ensures this code won't compile if pthread_setname_np() takes 3 arguments.      More in general, please be specific about what problem you want to fix in the commit.  Getting these tests to work in all platforms is hard and if it fails on some platform you want to understand why the current conditions are there.

mgondan:
Sorry for being so unspecific. My point is that -Werror is not needed for this check because the compiler raises an error if pthread_set_name_np is invoked with the wrong number of arguments. It's not a matter of warnings. 

More importantly, warnings may be raised for completely unrelated reasons.* Then, compilation fails because of -Werror, although the number of arguments was correct. The bottom line is that Werror is actually harmful here.

*my use case was a -Wstrict-prototypes that I inherited from R's CFLAGS. I admit that strict prototypes is a bit stupid warning, but still...

JanWielemaker:
Thanks.   Merged after update of comment.    I verified that todays gcc and clang indeed reject a wrong number of arguments even at the lowest warning level.    Once upon a time that was different ...

mgondan:
Interesting. I wasn't aware of that. On a slightly related note, I noticed that check_c_source_compiles also invokes the linker, which sometimes also causes unrelated errors (e.g., on Windows, a library may be missing etc.)

So, it may make sense to add a -c in some places (here, it didn't seem to be necessary).

PR #1268, opened 2024-04-17
Use LIBSWIPL_DIR in swipl.pc

mgondan:
Let pkg-config rswipl/inst/swipl/share/pkgconfig/swipl.pc --libs return the correct -L if SWIPL_INSTALL_IN_LIB=ON


JanWielemaker:
Thanks.  Merged after squash and clarify commit message.

PR #1270, opened 2024-05-03
FIXED: bf_set_si for -INT_MIN

mgondan:
avoid undefined behavior raised by UBSAN

### DIFF
--- a/src/libbf/libbf.c
+++ b/src/libbf/libbf.c
@@ -262,7 +262,10 @@ int bf_set_si(bf_t *r, int64_t a)
 {
     int ret;
 
-    if (a < 0) {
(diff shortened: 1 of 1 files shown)


mgondan:
First error message in https://www.stats.ox.ac.uk/pub/bdr/memtests/gcc-UBSAN/rswipl/rswipl-Ex.Rout

JanWielemaker:
Thanks.    Seems correct.   Except that UINT64_MIN does not exist.   Must be INT64_MIN.  Fixed.    Pushed an additional patch to remove the instantiation as the initialized variable is not used wit LibBF, only with gmp.

PR #1274, opened 2024-05-03
CLEANUP: workaround for an UBSAN error

mgondan:
UBSAN complained about a left shift of -1


mgondan:
Another rather low-hanging fruit

PR #1275, opened 2024-05-04
CLEANUP: Silence UBSAN message

mgondan:
Silences the following warning:

/.../swipl-devel/src/pl-thread.c:7649:5: runtime error: null pointer passed as argument 1, which is declared to never be null
/.../swipl-devel/src/pl-thread.c:7649:5: runtime error: null pointer passed as argument 2, which is declared to never be null

In the call that raises the error, arg1 and arg2 from a memcpy(arg1, arg2, bytes) are NULL, but bytes is also 0.


mgondan:
I see it has already been implemented. Closing this PR.

PR #1276, opened 2024-05-04
CLEANUP: Avoid UBSAN error

mgondan:
Test for arity > 0


JanWielemaker:
Thanks.   Merged after changing `n` to `size_t` as we are at it.

PR #1279, opened 2024-05-24
test_interrupt: skip test if prolog_alert_signal = 0

mgondan:


JanWielemaker:
Why did you close this?    Note that it is broken as `prolog_alert_signal/2` is not defined on all platforms (notably not on Windows).   So, we should __not__ run the test if this predicate is defined and `prolog_alert_signal(0, 0)` succeeds.    If this predicate is not defined we assume there is some other mechanism to make this work.

mgondan:
Ah, ok. I wasn't aware of this. Is an additional current_predicate needed? I'll reopen it in the afternoon

JanWielemaker:
I'm happy to adjust the PR.    I was more wondering whether the PR had become irrelevant for some reason.   If it is still relevant, I can deal with it.

mgondan:
All right. Thanks. No, it's not irrelevant.

JanWielemaker:
Ok.   Updated to 30fae0a22e871525e1cd8ba2bd640cf77e5994ba

PR #1280, opened 2024-05-25
DOC: fix typo and improve cmake instructions

ssardina:
- Fix small typos and fonts
- Improve documentation of how to meet pre-reqs before building
- Add more explanation on running SWIPL when having multiple installs (distribution, local install, build)

### DIFF
--- a/CMAKE.md
+++ b/CMAKE.md
@@ -24,6 +24,11 @@ from https://cmake.org/download/
 
 ## Native build
 
+The [build SWI-Prolog from source](https://www.swi-prolog.org/build/)
+page has information one how to get and build SWI-Prolog for various
+platforms. Check that page also for the prerequisites depending on the
+platform.
+
 ### Getting the source
 
 The   source   may   be    downloaded    as     a    tar    ball    from
@@ -52,10 +57,7 @@ latest version:
     git pull
     git submodule update --init
     cd build
-	cmake .
-    ninja
-    ctest -j 8
-    ninja install
+	cmake .. ninja ctest -j 8 ninja install
 
 If the build fails, one could try to remove the entire `build` directory
 and re-create it as  above.  Note  that   the  build  process  makes  no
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks.   Merged after adding a test for the conflicting variables to `swipl-activate` and print a warning from there rather than documenting the issue.

CMAKE.md
```diff
@@ -52,10 +57,7 @@ latest version:
     git pull
     git submodule update --init
```
kamahen, line 58: Isn't `git submodule  update --init` rather out of date? Instead: `git pull --recurse=yes`? (The `git clone` has the recursive option)

CMAKE.md
```diff
@@ -52,10 +57,7 @@ latest version:
...
     cd build
-	cmake .
-    ninja
-    ctest -j 8
-    ninja install
+	cmake .. ninja ctest -j 8 ninja install
```
kamahen, line 60: I did:
```
cd ~/src/swipl-devel/build && cmake .. ninja ctest -j 8 ninja install
```
and got:
```
CMake Warning:
  Ignoring extra path from command line:

   "install"


CMake Error: Unknown argument -j
CMake Error: Run 'cmake --help' for all supported options.
```

CMAKE.md
ssardina: Good point, yes the recurse option has been added to most git commands, including `pull`. 

Presumably it is `--recurse-submodules=yes` or simply `--recurse-submodules`, but seems `--recurse`  also works!

CMAKE.md
ssardina: @kamahen , but that does not seem to be the right command. Isn´t that meant to be:

```
cd ~/src/swipl-devel/build && cmake -G ninja .. && ninja && ctest -j 8 && ninja install
```

ssardina:
> Thanks. Merged after adding a test for the conflicting variables to `swipl-activate` and print a warning from there rather than documenting the issue.

Fantastic, much better handling with a better script now. :clap:

CMAKE.md
kamahen: (Yes - I forgot the `cmake -G Ninja ..` because I have the commands in a bash script with `-e`)

JanWielemaker: Apparently @ssardina's edits changes the multi-line command into a single line.  I changed that back in the final edit, so all should be fine with the file as it is now.

ssardina: ouch! sorry about that, that's sloppy. But it is strange, I didn't mean to do that and I dont see see that glitch on my side:

image

need to keep an eye on this, quite confused.

PR #1281, opened 2024-05-26
DOC: use modern update of submodules directly via git pull

ssardina:
- Address comment in previous PR #1280 from @kamahen . Basically update the doc to update submodules via `git pull` directly, as it is more modern

### DIFF
--- a/CMAKE.md
+++ b/CMAKE.md
@@ -54,8 +54,7 @@ as follows:
 In most cases the following should  update   an  installed system to the
 latest version:
 
-    git pull
-    git submodule update --init
+    git pull --recurse-submodule
     cd build
     cmake ..
     ninja
--- a/CMAKE.md
+++ b/CMAKE.md
@@ -54,12 +54,15 @@ as follows:
 In most cases the following should  update   an  installed system to the
 latest version:
 
-    git pull --recurse-submodule
+    git pull
+    git submodule update --init
     cd build
     cmake ..
     ninja
     ctest -j 8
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I tried this, but it is a lot slower as it fetches all the submodules, also if their commit hash has not been changed.    It seems to work better using 

     git pull --recurse-submodules=on-demand

But the command gets so long that I do not see the advantage any more.

kamahen:
`git pull --recurse=on-demand` is a bit shorter than `git pull && git submodule update --init` :)
(the option doesn't need to be fully spelled out but the value does)

And on my system, when things are up-to-date, `git pull --recurse=yes` is 3.6 seconds.

ssardina:
Agree with @JanWielemaker , not worth it, definitively not worth it with `yes`, too slow.

What about the fix I just did: leave the new extended pull with on-demand as a note?

JanWielemaker:
> What about the fix I just did: leave the new extended pull with on-demand as a note?

I don't see much point in that.    Anyone can copy/paste as is and  knowledgeable with git can do it otherwise.   Why make it more complicated than that?    Note that as-is, recursive pull does __not__ deal with new modules, so you still need an extra instruction for that.   It is in the bug list of git, so eventually it will be fixed (I guess).

ssardina:
OK let us then leave it as is and don´t go into more details and subtleties. The init command is the de-facto standard, at least for now.. :-)

PR #1282, opened 2024-05-27
Command-line options for swipl processes in unit tests

mgondan:


JanWielemaker:
Seems innocent enough.    I wonder how this is going to be used though.    I think the purpose was to avoid trying to run the save-program tests when embedded?     In general the distinction between embedded and not is not so easy.   What does it mean?    One embedding is not the other.    Some cause the loss of command line processing, some loss of multi-thread processing, some loss of signal handled, etc. 

I'd be ok if the embedding into Rtools add a flag rtools=true and a hack to disable the save program tests.    If you want to generalize a bit, we could define a flag that indicates that command line arguments are not processed.   There are already flags to detect thread and signal support.

That said, if the above solves your problems and you are happy with it, I'm happy to merge this.

mgondan:
The suggested patch is actually better than just disabling the tests, because it allows me to actually run the tests even if "swipl.exe" is not found on the system. The embedded system, at some point, calls "PL_initialize()", and that's where I can hand over the necessary command line arguments for swipl. So, instead of 

`swipl -f none -t halt -g goal`. 

I can invoke 

`R -e "rswipl:::swipl()" -q --no-echo --args -f none -t halt -g goal`.

So, my list of extra arguments becomes: -e "rswipl:::swipl()" -q --no-echo --args [argv 0 is taken from prolog flag "executable"]

Please note that the patch reduces the number of failed unit tests from 30 to 6, but not to zero. You might leave this PR open until I have fixed the other unit tests, too (it's probably signal handling again).

JanWielemaker:
As this is dedicated to R, I'll wait with merging until it fixes all you had planned to fix.    Just let me know!

Issue #1283, opened 2024-06-01
.gitmodules file contains relative paths and not absolute paths


gl-yziquel:
Hi.

I've been trying to integrate swipl into my build system. Here is the problem I encountered:

I have a main repository, on which I tinker with swipl-devel code. Then, I have a tool that clones that clone into another repository to perform builds. This allows me to decouple where I store and play with the code with builds on multiple SHAs.

Basically, when you "clone the clone",  the remote repository of the clone of the clone is mapped to the clone and not to the github repository. As such, the relative paths for submodules are interpreted as relative to my filesystem and not as relative to the github setup. And it thus fails to clone the submodules.

The way I fix this is via a relatively arcane piece of trickery:

```sh
for submodule in $(grep -E "^\[submodule" .gitmodules | sed 's/^\[submodule "//;s/"\]//'); do git config submodule.$submodule.url "$(git config --file /path/to/my/local/git/clone/of/swipl-devel/.git/config --get submodule.$submodule.url)"; done
```

This is awkward.

A better fix would be to make paths in your repository absolute for submodules in your `.gitmodules`file.

It would be cleaner anyway: as of now, it couples your repository with the implicit data of the hosting setup, which is counter to the original intent of git as a distributed system.

gl-yziquel:
For any potential reader: discussion occurred on the #1284 pull request.

rkaminsk:
We use relative paths to consistently checkout via ssh/https. Afaik this is not possible with absolute paths. I don't want to change this setup.

JanWielemaker:
The relative paths in .gitmodules are intentional so submodule URLs resolve correctly on any fork of the repository (GitHub, GitLab, etc.). For clone-of-a-clone workflows the git config submodule.<name>.url override you already use is the recommended approach. Closing.

PR #1284, opened 2024-06-01
Paths in .gitmodules are now absolute.

gl-yziquel:
https://github.com/SWI-Prolog/swipl-devel/issues/1283

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,123 +1,123 @@
 [submodule "packages/chr"]
 	path = packages/chr
-	url = ../packages-chr.git
+	url = https://github.com/SWI-Prolog/packages-chr.git
 [submodule "packages/jpl"]
 	path = packages/jpl
-	url = ../packages-jpl.git
+	url = https://github.com/SWI-Prolog/packages-jpl.git
 [submodule "packages/clpqr"]
 	path = packages/clpqr
-	url = ../packages-clpqr.git
+	url = https://github.com/SWI-Prolog/packages-clpqr.git
 [submodule "packages/inclpr"]
 	path = packages/inclpr
-	url = ../packages-inclpr.git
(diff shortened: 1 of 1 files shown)


JanWielemaker:
IMO relative URLs are the way to go for a single project split into multiple modules.    Only when using truly external repos as submodules is a good reason to use absolute URLs.    For one thing, it allows cloning using SSH or HTTP for the entire repository.

Unfortunately, GitHub is not very cooperative as it does not allow organizing repositories in directories.   When the project was still self-hosted we used directories and one could nicely clone clones.    I did a quick checks with github and gitlab, but I failed to get definitive answers.    The other hope would be that git cloning gets smarter ...

Alternative to your solution, you can create links from the parent of swipl-devel like

      ln -s swipl-devel/packages/clib packages-clib
      etc.

It is a bit awkward too :cry:

gl-yziquel:
> Unfortunately, GitHub is not very cooperative as it does not allow organizing repositories in directories. When the project was still self-hosted we used directories and one could nicely clone clones. I did a quick checks with github and gitlab, but I failed to get definitive answers. The other hope would be that git cloning gets smarter ...

Well, I'd maybe suggest using the repo or jiri tool to create a manifest. This way, you can pull everything in one go, with relative paths, and have only one place in the manifest to change if one wishes to move from github, gitlab, sourcehut, or self-hosting.

Anyway. It's just a pain. I experience that git clone of clone + submodules with quite a number or repos, as a matter of fact.

JanWielemaker:
It seems this can be solved using gitlab groups and subgroups.    All in all, I'm not convinced that a migration to gitlab is worth the trouble.     I might at a script to the `scripts` directory that can be called to prepare the repo for recursive cloning.     That seems to be as good as it gets with github.

gl-yziquel:
Your call. I take things as they are.

PR #1291, opened 2024-06-24
FIXED: intptr_t

mgondan:
compiles again on windows


mgondan:
I think this has already been fixed

Issue #1292, opened 2024-06-24
Crashes at command line


kamahen:
I've had a few "random" crashes at the interactive prompt, which aren't reproducible. They occur after I've reloaded my code a number of times (using `make/0`). The code is fairly basic Prolog (no foreign code, unless `aggregate/4` has some).

Attached are a couple of tracebacks, in case they point to a fairly recent change. `version 9.3.7-47-g1ff28598f-DIRTY)` commit 1ff28598f7c5da16396bbb7a3293b3ba9fb26878

```
?- taxes_usd(us, T).

ERROR: Received fatal signal 11 (segv)
Time: Sun Jun 23 09:19:52 2024
Inferences: 6256440
Thread: 1 (main)
C-stack trace labeled "crash":
  [0] save_backtrace() at /home/peter/src/swipl-devel/src/os/pl-cstack.c:337 [0x7fd94a917a38]
  [1] sigCrashHandler() at /home/peter/src/swipl-devel/src/os/pl-cstack.c:937 [0x7fd94a8b5b57]
  [2] __restore_rt() at libc_sigaction.c:? [0x7fd94a654050]
  [3] PL_same_term___LD() at /home/peter/src/swipl-devel/src/pl-prims.c:2136 (discriminator 1) [0x7fd94a850806]
  [4] PL_next_solution___LD() at /home/peter/src/swipl-devel/src/pl-vmi.c:5108 [0x7fd94a821861]
  [5] query_loop() at /home/peter/src/swipl-devel/src/pl-pro.c:155 [0x7fd94a8fcff2]
  [6] prologToplevel() at /home/peter/src/swipl-devel/src/pl-pro.c:594 [0x7fd94a8fce8d]
  [7] PL_toplevel() at /home/peter/src/swipl-devel/src/pl-fli.c:4953 [0x7fd94a9106aa]
  [8] swipl(+0x10c0) [0x55b232d550c0]
  [9] __libc_start_call_main() at ./csu/../sysdeps/x86/libc-start.c:74 [0x7fd94a63f24a]
  [10] call_init() at ./csu/../csu/libc-start.c:128 [0x7fd94a63f305]
  [11] swipl(+0x10f1) [0x55b232d550f1]


PROLOG STACK:
     [19] us_tax_usd(3767568.6, _6012)
     [18] item_and_tax_usd(non_retirement, us, _6012)
[...]

PR #1293, opened 2024-07-05
ENHANCED: Added 'c_cxx' flag and env SWIPL_CXX for packs

kamahen:
I did a quick check of `buildenv.sh` when I did `pack_install(.)` for a pack and it appears to be correct -- `SWIPL_CXX='g++'` was added.


cmake/BuildType.cmake
```diff
@@ -49,7 +49,7 @@ if(CMAKE_COMPILER_IS_GNUCC)
     set(GCC_OPTFLAGS -O2)
   endif()
 
-  set(CMAKE_C_FLAGS_DEBUG "-DO_DEBUG -DO_DEBUG_ATOMGC -O0 ${GCC_GFLAGS} $ENV{CFLAGS}"
+  set(CMAKE_C_FLAGS_DEBUG "-DO_DEBUG -DO_DEBUG_ATOMGC -O0 ${GCC_GFLAGS} $ENV{CFLAGS} -DO_CHECK_TERM_REFS -DSECURE_GC -DCHK_SECURE -DO_MAINTENANCE -gdwarf-2 -g3"
```
kamahen, line 52: You might not agree with this change -- but it has the flags that caused https://github.com/SWI-Prolog/swipl-devel/issues/1292 when doing a debug build.

library/build/tools.pl
```diff
@@ -435,12 +442,25 @@
...
+    has_program(path(CXX), _),
+    !.
+
 preferred_c_compiler(CC) :-
     current_prolog_flag(c_cc, CC).
 preferred_c_compiler(gcc).
 preferred_c_compiler(clang).
 preferred_c_compiler(cc).
 
+preferred_cxx_compiler('g++').
```
kamahen, line 459: Add:
```prolog
preferred_c_compiler(CXX) :-
    current_prolog_flag(c_cxx, CXX).
```

library/build/tools.pl
```diff
@@ -435,12 +442,25 @@
...
+    getenv('CXX', CXX),
+    !.
+default_cxx_compiler(CXX) :-
+    preferred_cxx_compiler(CXX),
+    has_program(path(CXX), _),
+    !.
+
 preferred_c_compiler(CC) :-
     current_prolog_flag(c_cc, CC).
 preferred_c_compiler(gcc).
```
kamahen, line 455: See https://github.com/SWI-Prolog/swipl-devel/issues/1294

Changing this to `preferred_c_compiler(path(gcc))` breaks `build_step(configure, ...)`.

kamahen (commented):
This PR needs more work - See https://github.com/SWI-Prolog/swipl-devel/issues/1294

cmake/BuildType.cmake
JanWielemaker: Skipped this :smile:

library/build/tools.pl
```diff
@@ -431,15 +438,30 @@
...
+    preferred_cxx_compiler(CXX0),
+    has_program(CXX0, CXX),
     !.
 
 preferred_c_compiler(CC) :-
     current_prolog_flag(c_cc, CC).
-preferred_c_compiler(gcc).
-preferred_c_compiler(clang).
-preferred_c_compiler(cc).
+preferred_c_compiler(path(gcc)).
```
JanWielemaker, line 455: I could find no good reason for the path() wrapper.   has_program/2 adds this if necessary.   If this fixes some issue I think it is has_program/2 that needs fixing.    So, I skipped this.

JanWielemaker:
Applied after the mentioned changes and some documentation changes.

library/build/tools.pl
kamahen: Line 521 (program_install_prefix/1) has `has_program(path(Name), EXE),` ... is it wrong?

`has_program/2` uses `pack_build_path(...)`, not `path(...)` ... presumably the former is better?

PR #1309, opened 2024-08-27
modify cmake ilink to install symlink in CMAKE_INSTALL_PREFIX

matko:
The `ilink` utility function in the cmake build was installing symlinks under `CMAKE_INSTALL_PREFIX` as it was known at configure time. This PR changes `ilink` to use `CMAKE_INSTALL_PREFIX` at install time, and modifies the calls to `ilink` accordingly.

`ilink` now takes two absolute paths without the `CMAKE_INSTALL_PREFIX`, and prepends `CMAKE_INSTALL_PREFIX` at install time.

This is a 3 part PR, since two submodules (swipl-win and jpl) also have to be changed to this new `ilink` contract. They need to be merged first, and then this PR can be updated to also include a submodule update.


matko:
Thanks! I'll be sure to properly annotate commits from now on.

PR #1319, opened 2024-09-09
TEST: prevent ASAN breaking sanity check

kamahen:


JanWielemaker:
This does not really seem to be the right way to fix as we still get negative stack sizes.  The real seems to be that this way to check for stack overflows does not work with ASAN (though I do not see any errors so far).  

Anyway, the entire function is disabled by 

      #if O_PLMT && !defined(__SANITIZE_ADDRESS__)

Is that not set for some reason for you?    It should be set by GCC AFAIK and it is set in pl-incl.h for clang, which does it differently.

PR #1326, opened 2024-10-17
FIXED: build order for dependencies in packs

mgondan:
I think the order is in the wrong way. Say, pack a depends on pack b, and both have post-installation scripts. Then, pack b's script should be run before pack a's script. This does not seem to be the case here. If I reverse the order, it is running fine.

There's probably a better way to fix this (e.g., in build_order).

### DIFF
--- a/library/prolog_pack.pl
+++ b/library/prolog_pack.pl
@@ -1415,8 +1415,9 @@ build_plan(Plan, Ordered, Options) :-
     (   ToBuild == []
     ->  Ordered = []
     ;   order_builds(ToBuild, Ordered),
-        confirm(build_plan(Ordered), yes, Options),
-        maplist(exec_plan_rebuild_step(Options), Ordered)
+        reverse(Ordered, Reversed),
+        confirm(build_plan(Ordered), yes, Reversed),
+        maplist(exec_plan_rebuild_step(Options), Reversed)
     ).
 
 needs_rebuild_from_info(Options, Info) :-


JanWielemaker:
Thanks. Should be fixed with 1cf4acbe722604857a2627ecc161f80e16f330ed, which changes build_order/2.

mgondan:
I am a bit unsure if I am doing everything correct, but on my system, the build order hasn't changed :-/

mgondan:
pack_install(interval).

It's a bit special since it requires connects to R, but you see the build order before the installation (that currently fails).

````
% Contacting server at https://www.swi-prolog.org/pack/query ... ok
The following packs have post install scripts:
  Build interval in directory /home/gondan/.local/share/swi-prolog/pack/interval
  Build rologpp in directory /home/gondan/.local/share/swi-prolog/pack/rologpp
Run scripts? Y/n?
% Building pack interval in directory /home/gondan/.local/share/swi-prolog/pack/interval <<<<<< rologpp should have come first
-- The C compiler identification is GNU 13.2.0
-- The CXX compiler identification is GNU 13.2.0
````

JanWielemaker:
Ok.   This one now works fine.   The previous patch on this was wrong, but it did reverse the order.   I suspect that the reverse (implemented as a swap of the edge directions) was the incorrect fix, possibly for some other problem.   I'm pretty convinced this part of the logic is now correct (and always was), so if a new problem arises we must carefully analyse it.

PR #1336, opened 2024-12-13
Add SWIPL_SO_VERSIONS CMake option.

rmanhaeve:
Add the SWIPL_SO_VERSIONS Cmake option. By default it is ON, and then the VERSION and SOVERSION properties of the library are set. 
Setting SWIPL_SO_VERSIONS to OFF does not include this information, leading to only one shared library being built. This is useful to reduce bloat when packaging SWI Prolog in a format where symlinks are not supported (e.g. Python Wheels).


JanWielemaker:
Thanks.   Merged after squash and some minor edits.

PR #1348, opened 2025-02-26
Add predefined function del/1 on dict with some minimal tests

esad:
This may need more work, but I just wanted to post this PR for discussion and see if this is something others may find useful. While working with dictionary based data structures, I've come to rely on `put(KeyPath, Value)`, but there is no "function" for deleting keys. This would add another predefined `del` consistent in naming with `del_dict/4` that also accepts a keypath.


JanWielemaker:
Thanks.  Makes sense.   We could consider a del(Key,Value), where Value is unified with the current value.    The code could be simplified using SSU (=>).    And of course, we need documentation ...

PR #1356, opened 2025-03-20
DOC: install with symlinks

kamahen:


JanWielemaker:
Thanks.  I moved and rephrased this as there is already a paragraph on using symlinks in
the "Development" section of this file.

Issue #1370, opened 2025-05-30
Running strip on standalone binary removes save state


xelxebar:
## Overview

Running [GNU binutils][2] `strip` on a `qsave_program/2` standalone ELF binary removes the save state data, resulting in a binary that just drops to a dry swipl REPL, _i.e._

```
$ prog        # Works
$ strip prog  # This should be harmless, right?
$ prog        # Now broken
```

The standalone binary doesn't contain debug symbols in the first place, so arguably the correct action is to just not use `strip` in the first place. However, there is at least [one previous report][1] on the forums, and [in my case][0], `strip` is automatically run by a distribution package builder, causing very confusing breakages.

## Analysis

The issue is that `strip` removes any data in an ELF not referenced by an ELF section, and `qsave_program/2` produces a Zip archive concatenated to an ELF stub, meaning that the trailing Zip data remains anonymous:

```
$ binwalk prog

DECIMAL       HEXADECIMAL     DESCRIPTION
--------------------------------------------------------------------------------
0             0x0             ELF, 64-bit LSB executable, AMD x86-64, version 1 (SYSV)
294522        0x47E7A         End of Zip archive, footer length: 22
```

but

```
$ cp prog prog.stripped
$ strip prog.stripped
$ cmp prog prog.stripped
cmp: EOF on plwm.stripped after byte 14392, in line 2
$ binwalk prog.stripped

DECIMAL       HEXADECIMAL     DESCRIPTION
--------------------------------------------------------------------------------
0             0x0             ELF, 64-bit LSB executable, AMD x86-64, version 1 (SYSV)
```

This show that `strip` was a no-op on the ELF stub but removed everything after byte `14392 = 0x3838`.

## Proof of Concept Fix

The idea for mitigation is simple: just create an ELF section that references the Zip archive data. Below is a manual demonstration of the idea that I have confirmed works:

```
$ unzip a.out; zip -Dr '$prolog' a.zip     # Extract the zip archive
$ strip a.out; mv a.out a.stub             # Extract the ELF stub
$ objcopy --add-section .zipdata=a.zip --set-section-flags .zipdata=readonly,data a.stub plwm
```

Is it reasonable to patch `qsave_program/2`, or wherever relevant, to add such an ELF section? Other thoughts?

[0]:https://github.com/Seeker04/plwm/issues/92#issuecomment-2920620411
[1]:https://swi-prolog.discourse.group/t/swi-prolog-standalone-binary-vs-strip/7527
[...]

PR #1371, opened 2025-05-31
WIP: Protect qsave_program/2 from strip clobbering

xelxebar:
Addresses #1370.

### DIFF
--- a/library/qsave.pl
+++ b/library/qsave.pl
@@ -137,7 +137,8 @@ qsave_program(File) :-
 qsave_program(FileBase, Options0) :-
     meta_options(is_meta, Options0, Options1),
     check_options(Options1),
-    exe_file(FileBase, File, Options1),
+    exe_file(FileBase, ExeFile, Options1),
+    zip_file(FileBase, ZipFile, Options1),
     option(class(SaveClass), Options1, runtime),
     qsave_init_file_option(SaveClass, Options1, Options),
     prepare_entry_points(Options),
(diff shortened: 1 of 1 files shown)


xelxebar:
Correctly setting the `.zipdata` ELF section, but apparently I've broken something:

```
[FATAL ERROR: at Sat May 31 17:21:56 2025
        Resource database "/src/swipl-devel/build/man/pldoc2tex" does not contain "$prolog/state.qlf"]
```

and yet,

```
$ unzip -l /src/swipl-devel/build/man/pldoc2tex
Archive:  man/pldoc2tex
warning [man/pldoc2tex]:  24 extra bytes at beginning or within zipfile
  (attempting to process anyway)
  Length      Date    Time    Name
---------  ---------- -----   ----
      176  05-31-2025 17:21   $prolog/options.txt
  1141536  05-31-2025 17:21   $prolog/state.qlf
---------                     -------
  1141712                     2 files
```

mgondan:
To reproduce the problem:
````
$ cp ../test/save/input/plain.pl .
$ swipl -o plain.exe -g echo_true -c plain.pl --stand-alone
$ ./plain.exe
true.
$ strip plain.exe
$ ./plain.exe
(broken, just invokes swipl)
````
To fix it outside of qsave_program/2:
````
$ swipl -o plain.exe -g echo_true -c plain.pl --stand-alone
% Disabled autoloading (loaded 42 files)
% Disabled autoloading (loaded 8 files)
% Disabled autoloading (loaded 0 files)
$ unzip plain.exe
Archive:  plain.exe
  inflating: $prolog/options.txt
  inflating: $prolog/state.qlf
$ zip -r plain.zip '$prolog'
  adding: $prolog/ (stored 0%)
  adding: $prolog/options.txt (deflated 20%)
  adding: $prolog/state.qlf (deflated 61%)
$ strip plain.exe
$ mv plain.exe plain.stub
$ objcopy --add-section .zipdata=plain.zip --set-section-flags .zipdata=readonly,data plain.stub plain.exe
$ ./plain.exe
true.
$ strip plain.exe
$ ./plain.exe
true.
````
Instead of changing anything in the "prolog"-part of qsave_program.pl, I suggest to append the shell commands to the end of the generation of a stand-alone executable:

````
...(finalize),
    option(stand_alone),
    current_prolog_flag(..., elf),
    absolute_file_name(path(zip), Zip, ...),
    absolute_file_name(path(unzip), Unzip, ...),
    absolute_file_name(path(objcopy), Objcopy, ...),
    shell(unzip exefile),
    shell(zip to zipfile),
    shell(strip exefile), % can be optional
    shell(mv exe to stub),
    shell(objcopy etc.).
````
Did I miss something?

mgondan:
See https://github.com/SWI-Prolog/swipl-devel/pull/1413

JanWielemaker:
Not sure this should have any priority.   If we do though, it is rather odd to create a state consisting of the exe + zip, then unpack the whole lot and finally reassemble it using objcopy.    It is trivial to create the zip without the exe at the start by having make_header/3 do nothing.  Next, the finalize step can call objcopy.    This only if we are dealing with elf binaries and objcopy actually exists.     This route also avoids depending on zip/unzip.   Doable, but is it worth the trouble?

mgondan:
> It is trivial to create the zip without the exe at the start by having make_header/3 do nothing. Next, the finalize step can call objcopy.

Out of curiosity, where does the header for the "self-extraction" come from? More practically: This procedure would end up in a single file, plain.zip. What would the objcopy command look like?

> Doable, but is it worth the trouble?

I don't know. I just found the problem interesting. I admit that my proposed solution is a bit clumsy.

kamahen:
About self-extracting executables, something you might want to keep in mind ... at Google we had problems with self-extracting Python executables (which are essentially zip files plus a Python interpreter) because the architects of the data centers wanted to have read-only "disks" for locally-attached storage, with all writing being done using network-attached storage (this design might have changed of course). A proposal was made to modify dlopen() to allow reading from generalized streams but this was rejected by the Linux community. I don't recall how the problem was solved, but it wasn't easy. If you add things like shared-libraries (which Google was trying to eliminate for a number of maintainability reasons), the problem becomes even more difficult.

JanWielemaker:
> Out of curiosity, where does the header for the "self-extraction" come from? More practically: This procedure would end up in a single file, plain.zip. What would the objcopy command look like?

The header is either `swipl.exe` itself or, on non-Windows, a shell script if stand-alone is not activated.    It relies on the fact that zip archives are designed to allow for arbitrary data before the zip file.    If I read the command correctly, objcopy provides an alternative for combining the header executable and the zip data that adds the zip data as a normal ELF section, so strip is aware of it.

> If you add things like shared-libraries (which Google was trying to eliminate for a number of maintainability reasons), the problem becomes even more difficult.

I know.    There is basically no portable solution AFAIK.    As is, Prolog states can contain shared libraries.   If they do, these are extracted to a tmp folder and loaded.    On non-Windows we can do this fairly cleanly by loading the extension and immediately deleting it.    On Windows we have to keep the file around and hope we can remove it on program termination.

mgondan:
> The header is either swipl.exe itself or, on non-Windows, a shell script if stand-alone is not activated. It relies on the fact that zip archives are designed to allow for arbitrary data before the zip file. If I read the command correctly, objcopy provides an alternative for combining the header executable and the zip data that adds the zip data as a normal ELF section, so strip is aware of it.

I simplified the PR accordingly. It's still clumsy and work in progress, still depending on zip and unzip, but is now two lines shorter.

mgondan:
The other PR is now ready for review. If you think it's not helpful at this stage (e.g., because of Version 10), we might postpone it until there is a new development release.

kamahen:
> Prolog states can contain shared libraries

There's a difference between containing shared libraries and using shared libraries. I don't see much reason for containing shared libraries (you could just as easily statically link with `.a`  files); and using shared libraries turns out to have some nasty problems in production systems (not just "DLL hell" (which is partially solved in Linux); I can expand on the problems if you want), only a few dubious advantages, and usually little or no savings in RAM.

JanWielemaker:
This patch broke saved states on Linux.    I've pushed two commits (66cd5119b082b606580545a47b633c07082d10d7 and 88dfeb032239c548934ee19512651fedb028721a) to address the issue and add better documentation on why it works as it does.

PR #1383, opened 2025-07-30
Profiler related fixes

ridgeworks:
Two fixes:
- support for `[]` functor name in `prolog_code:predicate_label/2`
- fix predicate timing data in profiler default text display (was previously including callee cluster times)


JanWielemaker:
Thanks.   Merged after rebase and adding a ChangeLog tag to the commit messages.   Github doesn't understand that :cry:

PR #1388, opened 2025-09-09
Suppress unused warning under MSYS2

mgondan:
Unsure if you like this PR, but it suppresses a warning under MSYS2 (namely, that grace_time is unused)


JanWielemaker:
> Unsure if you like this PR, but it suppresses a warning under MSYS2 (namely, that grace_time is unused)

Not really, but I didn't see an obvious way to make it likeable either :cry:

PR #1390, opened 2025-09-13
More robust check for `git` command

ridgeworks:
On MacOS the `git` command is only present after the (optional) xcode support has been installed. Prior to that there is an "imposter" `git` which outputs text and a dialog indicating how xcode can be acquired. 

For users with no need for xcode, this change provides a more robust check for an operational `git` by additionally checking for a man page. I think this change is platform agnostic but I haven't tested it on Unix or Windows.

### DIFF
--- a/library/prolog_pack.pl
+++ b/library/prolog_pack.pl
@@ -2801,10 +2801,14 @@ version_part(Int) --> integer(Int).
 
 have_git :-
     process_which(path(git), _),
-    catch(run_process(path(man),['-w', 'git'],[output(_),error(_)]),  % silent
-          error(Err,_),
-          \+ Err = process_error(_,_)  % test succeeds if `man` cmd not present
-         ).
+    (process_which(path('xcode-select'), XSpath)  % MacOS -> check for "imposter"
+     -> catch(run_process(path('xcode-select'),['-p'],[output(_),error(_)]),  % silent
+              _Err,  % any error -> xcode not installed,
+                     % any `git` in same dir as `xcode-select` is an imposter
+              \+ (file_directory_name(Gpath,Gdir), file_directory_name(XSpath,Gdir))
+             )
+     ;  true         % no `xcode-select`, assume valid `git`
+    ).
 
 
 %!  git_url(+URL, -Pack) is semidet.


ridgeworks:
That is not going to work. Minimal Linux installations (e.g., for Docker) often have no man. Windows does not have man. 

I tried to accommodate that case. If there is no `man`, `run_process` generates an `existence error`, rather than a `process_error`. When that happens the handler will succeed and `have_git` will succeed. So no `man` command essentially defaults to the current test.

Now that doesn't handle the case where there are `man` and  `git` commands but no man page for `git`. In that case `have_git` will fail even though there's a working version of `git` (the non-git implementation is then used). And, from my understanding, that will only cause a problem when a `.git` URL is used with something other than GitHub. If that's a realistic scenario, I'll fallback to executing `git` to test, but it's a slightly inferior solution IMO (which is why I attempted to use Discourse to  generate feedback prior to generating the PR).

JanWielemaker:
Systems designed for this stuff like GNU autoconf or CMake all find the executable and for executables that have broken versions floating around they run some sanity check by executing the program in some scenario that distinguishes the fake from the real thing.    Using `--version` seems about right here if the fake version does something different that looks like a reliable test (exit code, message).   If I recall correctly that was also your first proposal.

The real git gives status 0 and writes "git version <version>", to stdout, where Apple git puts some stuff after that.

ridgeworks:
Just to clarify the discussion.

So using `--version` on a MacOS system with proper `git` command:
```
?- catch(build_tools:run_process(path(git),['--version'],[output(So),error(Se)]),Err,true).
So = [103, 105, 116, 32, 118, 101, 114, 115, 105|...],
Se = [].
```
and on a system with an "imposter" `gif`:
```
?- catch(build_tools:run_process(path(git),['--version'],[output(So),error(Se)]),Err,true).
Err = error(process_error(process(path(git), ['--version']), exit(1)), _).
```
so from the SWIP perspective, this clearly different behaviour looks good. And the options can be used to suppress any output. The problem is that the  imposter on MacOS launches another app offering to install the xcode command line tools. If there is an argument to be made that you shouldn't really need xcode to install add-on packs, at best this is annoying and, at worst, confusing. See:

https://github.com/microsoft/vscode-python/issues/20446

https://github.com/Zettlr/Zettlr/issues/5534

https://discussions.apple.com/thread/252157812?sortBy=rank

I admit that using `man` to avoid actually calling the imposter is a bit of a hack, but it does avoid the spurious launch of the "Install Command Line Developer Tools" app on MacOS. 

On the other hand, not using the `git` command has obvious deficiencies (as documented on the Discourse thread). A package manager which is operationally independent of external tools is a nice aspirational goal but one which is beyond me given my current understanding of git and any existing SWIP support, e.g., `library(git)`.  Maybe that's not a big deal - most users, especially developers, will just install xcode which is probably the right thing (i.e., most robust) to do anyway . 

Bottom line: If you still think the `--version` approach is preferred despite its deficiency, I'll proceed to make the change in the PR.

ridgeworks:
As imposter commands seems to be a uniquely MacOS issue, how about making the "man test" conditional on `current_prolog_flag(apple,true)`, e.g.,
```
have_git :-
    process_which(path(git), _),
    % check for "imposter"
    (current_prolog_flag(apple,true)
     -> catch(run_process(path(man),['-w', 'git'],[output(_),error(_)]),  % silent
              error(Err,_),
              \+ Err = process_error(_,_) % will succeed if no 'man' command
             )
     ;  catch(run_process(path(git),['--version'],[output(_),error(_)]),  % silent
              _Err, 
              fail                        % any error
            )
    ).
```
Other OS's will do the "version test". Or not do any test at all?

Still a hack, but a contained one, and it avoids the spurious launch on MacOS. I'm fairly confident xcode is unlikely to change in a way that renders the test invalid.

JanWielemaker:
This gets too messy for me.   I understand the issue is that the fake `git` starts the xCode installer?   We need something MacOS specific as there is only one such broken OS.   ChatGPT says

   - Check the path.   If /usr/bin/git, it may be a stub.  Otherwise it is a proper git.  This deals with Homebrew or other non-xcode installations of git. 
   - Now we seem to have two options.  One is to see whether `xcode-select` is present and if so, run `xcode-select -p`.  If that returns a sensible path, git is fine.   The other is to check the size.  The stub is small, while the real thing is 120Kb (on my machine)

I think that makes some sense.  Notably the xcode-select route looks promising.  For me it returns `/Applications/Xcode.app/Contents/Developer`. Using `man` seems flaky.  You may have man from Homebrew or Macports, while the user's path stiff resolves to the MacOS git binary or the other way around.

ridgeworks:
Using `git` size also seems flaky; on my machine (Intel) the real `git` is 18KB. (The imposter was overwritten, but it's 119KB on the the non-xcode machine.)

The following implementation using `xcode-select` (seems to work:
```
have_git :-
    process_which(path(git), _),
    (process_which(path('xcode-select'),_)  % MacOS -> check for "imposter"
     -> catch(run_process(path('xcode-select'),['-p'],[output(_),error(_)]),  % silent
              _Err,
              fail  % no xcode tools so not a valid `git`
             )
     ;  true        % no `xcode-select`, assume valid `git`
    ).
```
This does assume that only the xcode version of `git` is valid. If xcode is not installed but `git` has been installed from somewhere else (MacPorts, ..), it won't be recognized. OTOH, if there was another installation of `git` that did come with a man page, I think the previous implementation would succeed. So I'm not sure there's a clear win for either approach but it's your call.

FWIW, `xcode-select -p` on my machine (with xcode) prints `/Library/Developer/CommandLineTools`.

JanWielemaker:
> FWIW, `xcode-select -p` on my machine (with xcode) prints `/Library/Developer/CommandLineTools`.

But, what would it print if xcode was not enabled?

Otherwise, we should consider git commands that are not /usr/bin/git valid as it most likely indicates a git version that is not from Apple, e.g., binary installer, homebrew, macports.

ridgeworks:
But, what would it print if xcode was not enabled?

If xcode is not installed:
```
?- catch(build_tools:run_process(path('xcode-select'),['-p'],[]),Err,true).
xcode-select: error: Unable to get active developer directory. Use `sudo xcode-select --switch path/to/Xcode.app` to set one (or see `man xcode-select`)
Err = error(process_error(process(path('xcode-select'), ['-p']), exit(2)), _).
```
i.e., you actually get an error which IMO is better.

    Otherwise, we should consider git commands that are not /usr/bin/git valid as it most likely indicates a git version that is not from Apple, e.g., binary installer, homebrew, macports.

OK, so the condition for failing is: 
1. `xcode-select` exists - we're running on MacOS
2. `xcode-select -p` generates an error - xcode tools have not been installed.
3.  the current `path(git)` is in the same directory as `path(xcode-select)` - `git`'s not in the same directory as `xcode-select` are usable.

Full implementation as tested:
```
have_git :-
    process_which(path(git), Gpath),
    (process_which(path('xcode-select'), XSpath)  % MacOS -> check for "imposter"
     -> catch(run_process(path('xcode-select'),['-p'],[output(_),error(_)]),  % silent
              _Err,  % any error -> xcode not installed,
                     % any `git` in same dir as `xcode-select` is an imposter
              \+ (file_directory_name(Gpath,Gdir), file_directory_name(XSpath,Gdir))
             )
     ;  true        % no `xcode-select`, assume valid `git`
    ).
```

Still a small assumption that non-Apple versions of `git` do not overwrite the Apple imposter.

JanWielemaker:
From the discussions and internet, I think 4a25a1d6a7702bfbd8bb76cd711d9cf360192c7c is acceptable clean and takes care of MacOS.   Please test and close if you agree.

ridgeworks:
Please test and close if you agree.

Not quite. Xcode comes in two flavours: the full app (very large install) and command line tools only. Executing `xcode-select -p` outputs a string ending with `"Xcode.app/Contents"` for the full app but `"/Library/Developer/CommandLineTools"` for the command line version. (see "Verify Installation" in https://www.mikegopsill.com/posts/install-xcode-command-line-tools/ ). So if you need to check the output, you have to accommodate both options.

JanWielemaker:
:cry:  Maybe 774a4c30d7ab1f726e59ee590040ec279909dff6?

ridgeworks:
😢 Maybe https://github.com/SWI-Prolog/swipl-devel/commit/774a4c30d7ab1f726e59ee590040ec279909dff6?

Yes, that works :)

Minor issue with warning when pack manager loaded:
```
?- pack_install(b_real).
Warning: /Applications/SWI-Prolog9.3.28.app/Contents/Resources/swipl/library/prolog_pack.pl:2833:
Warning:    Clauses of prolog_pack:is_sane_git/1 are not together in the source-file
Warning:    Earlier definition at /Applications/SWI-Prolog9.3.28.app/Contents/Resources/swipl/library/prolog_pack.pl:2822
Warning:    Current predicate: prolog_pack:sane_xcode_path/2
Warning:    Use :- discontiguous prolog_pack:is_sane_git/1. to suppress this message
% Contacting server at https://www.swi-prolog.org/pack/query ... ok
Installation plan:
  Install b_real at version 1.0 from https://stoics.org.uk/~nicos/sware/packs/b_real/b_real-1.0.tgz (downloaded 72 times)
  Install lib at version 2.11 as dependency for [b_real] from https://stoics.org.uk/~nicos/sware/packs/lib/lib-2.11.tgz (downloaded 69 times)
Download packs? Y/n? 
% Downloading b_real ... 91,895 bytes
% Downloading lib ... 38,808 bytes
% Contacting server at https://www.swi-prolog.org/pack/query ... ok
true.
```

jan-swi-prolog-01:
Thanks.  Pushed a fix for that.

PR #1392, opened 2025-09-16
MSYS2: Fix unit test for saved states (PR incomplete, advice needed)

mgondan:
This adds --foreign=copy to the unit test for saved states. I am unsure if it is _really_ needed, but at least the test passes if test_installation is run outside of swipls "bin" folder.

So, this worked before:
```
cd swipl/bin
./swipl
?- test_installation.
```

Now, this also works:
```
cd swipl
bin/swipl
?- test_installation.
```

The flip side is that some of the dlls are copied to the current folder, and they remain there. Unsure how to clean up sensibly.

### DIFF
--- a/tests/save/test_saved_states.pl
+++ b/tests/save/test_saved_states.pl
@@ -157,13 +157,12 @@ set_windows_path.
 
 create_state(File, Output, Args) :-
     me(Me),
-    append(Args, ['-o', Output, '-c', File, '-f', none], AllArgs0),
     (   current_prolog_flag(msys2, true)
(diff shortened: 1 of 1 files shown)


mgondan:
The idea is to collect the names of the DLLs located in the directory in the beginning, check which ones were added by --foreign=copy, and only delete those in the end. We therefore avoid accidental deletion of, say libswipl.dll if test_installation is invoked in the bin folder.

mgondan:
Closing this, I found a simpler way.

PR #1394, opened 2025-09-18
MSYS2: Set SWI_HOME_DIR in unit test for saved state

mgondan:
The environment variable seems to be necessary so that the subprocess in the unit test finds the foreign libraries (under MSYS2).


JanWielemaker:
Thanks.   That looks like a nice clean fix :smile:

PR #1395, opened 2025-09-22
swipl.cmake: C standard 11

mgondan:
needed to cast foreign_t to pl_function_t. I proposed a similar change in pack environ, but I think it is also safe to put it here (and not too intrusive).


JanWielemaker:
Again, I do not want to fix the C standard.  13b22ede93cd764b1b5c1a25d273f18bfc75549c should fix this issue without setting the standard.

mgondan:
Thanks, it's working now.

PR #1398, opened 2025-10-01
Corrected argument for error message in pl-files.c

mgondan:
tmp_file_stream(F, S, [encoding(binary1), extension(txt)]). ERROR: Domain error: `encoding' expected, found `txt'

The txt does not make sense

### DIFF
--- a/src/os/pl-files.c
+++ b/src/os/pl-files.c
@@ -869,7 +869,7 @@ PRED_IMPL("$tmp_file_stream", 4, tmp_file_stream, 0)
     { enc = ENC_OCTET;
       mode = "wb";
     } else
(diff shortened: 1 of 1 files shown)


mgondan:
The last error message is also strange,

```
   tmp_file_stream('/', S, [encoding(binary), extension(exe)]).
ERROR: Uninstantiated argument expected, found binary (2-nd argument)
```

JanWielemaker:
Pushed a65b7469e2e51b6b046f3e70e47721ff351e1cc7 to make the error match the signature of the helper.  Not ideal, but given the fact that the helper is in the backtrace it should be acceptable.

PR #1399, opened 2025-10-10
TEST: Skip saved states if current directory is not writeable

mgondan:
The test is a bit cumbersome, also for issues of portability). The temp folder is temporarily set to "." and then a temporary file is created. If that succeeds, we are fine. 

Otherwise, an exception is reported to the debug stream.

### DIFF
--- a/tests/files/test_pipe.pl
+++ b/tests/files/test_pipe.pl
@@ -67,16 +67,17 @@ test(pwd, [condition(not(wine))]) :-
 test(cat1) :-
     current_prolog_flag(tmp_dir, Tmp),
 	current_prolog_flag(pid, Pid),
-    format(atom(File), '~w/pltest-~w.txt', [Tmp, Pid]),
     (   current_prolog_flag(windows, true)
-    ->  format(atom(Bat), '~w/pltest-~w.bat', [Tmp, Pid]),
+    ->  format(atom(File), '~w\\pltest-~w.txt', [Tmp, Pid]),
+	    format(atom(Bat), '~w/pltest-~w.bat', [Tmp, Pid]),
         setup_call_cleanup(
             open(Bat, write, Fd1),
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I doubt we want to go this way.    The directory should not only be writeable, but also allow for running executables.   Notably temp file systems increasingly deny execute access.  Of course you can test anything, but it gets hard to maintain and explain.  I don't think there is a portable test except for creating an executable, put it there and see whether it runs.   In theory we could use `swipl-ld.exe` for that except that I like to replace that at some point by a Prolog program and not all installations should have  `swipl-ld.exe`  ...    So, to be safe we must add a simple executable to the installation for testing of the installed system ...

Why do we want this?   Aren't the tests executed as part of the installation process, so we know some things about the environment?

mgondan:
> Aren't the tests executed as part of the installation process, so we know some things about the environment?

That's the thing: the tests can be executed as part of an installation process from source. But in binary installations, they are typically skipped, which is a bit of a pity. Personally, I am happy with the test for a "writable" folder. Of course, that does not include the issue with executable files—which I suggest to postpone until someone needs it.

mgondan:
I noticed a misconception at my end. The default working directory is usually writable, even on Windows. For whatever reason, I thought it is C:\Program Files\swipl, but it's not. In other words, there's really no reason to apply this patch.

Closing this. Sorry for wasting your time. And my own time too :-)

JanWielemaker:
Indeed, the default working directory for `swipl-win` is writeable.   `swipl` has no default working directory.  That is just the dir where you start it from.    Anyway, I'm happy to leave it as is.   I added this to the docs:

```
%   When  running  this  predicate  the   _working  directory_  must  be
%   writeable and allow for writing  executable   files.  This is due to
%   tests  for  file  system  interaction    and  tests  for  generating
%   stand-alone executables.
```

PR #1404, opened 2025-11-13
PORT: Allow for shell scripts swipl.sh and swipl.bat in test_saved_states

mgondan:
So far, test_saved_states allows for a swipl.sh in the current folder as a replacement for swipl. I guess this has to do with embedded systems and actually comes in quite handy if swipl should be invoked with extra arguments. I have added the same functionality for a swipl.bat under Windows. Since process_create/3 cannot invoke swipl.bat directly, but has to call cmd.exe /c swipl.bat instead, me/2 now takes an extra argument.

You can see this patch in action here, https://github.com/mgondan/rswipl/blob/main/inst/patch/05-tests.patch

### DIFF
--- a/tests/save/test_saved_states.pl
+++ b/tests/save/test_saved_states.pl
@@ -123,7 +123,7 @@ state_output(Id, State) :-
 
 %!  me(-Exe, -Args:list) is det.
 %
-%   True when Exe is the executable to   run  for creating a saved state
+%   True when Exe is the executable  to  run  for creating a saved state
 %   and Args is a  list  of  commandline   arguments  to  pass  to  Exe.
 %   Normally, this merely runs `swipl`, but it can be necessary to embed
 %   this into some sort of script.


JanWielemaker:
Sounds good.    Why not simply test write access using access_file/2?

```
?- access_file(., write).
```

mgondan:
Thanks, interesting. And very short indeed. But this PR does not test for write access anymore. I think that test is actually not needed, it was a misconception at my end.

mgondan:
I am sorry, I have to correct a tiny mistake in a second PR (a space was missing in atomic_list_concat).

JanWielemaker:
This one is merged after some small edits.  Thanks.

JanWielemaker:
Ok.  Added the space by hand.   All set and done (I hope).

mgondan:
This is a great moment, because I can now run test_installation with packages(false) from the unit test of the rolog and rswipl R packages 💪

Expect a few more PRs when I try to switch packages(true) 😊 Not soon, though

PR #1405, opened 2025-11-16
TEST: Add unit test for saved states with swipl.sh

mgondan:
Details: 
1. invoke swipl.sh with path(sh): solves a problem under R package rswipl.
2. remove SWI_HOME_DIR. This causes problems when using an external swipl with another home.
3. add unit tests for saved states with script.

### DIFF
--- a/tests/save/test_saved_states.pl
+++ b/tests/save/test_saved_states.pl
@@ -79,6 +79,11 @@ test_saved_states :-
 :- dynamic
     test_dir/1.
 
+% Invoke the swipl executable, arguments to prepend, environment variables
+% Example: 
+% swipl('R', ['-s', '-e', 'rswipl::swipl()', '--args'], []).
+:- dynamic user:swipl/3.
+
 :- prolog_load_context(directory, Dir),
    retractall(test_dir(_)),
    asserta(test_dir(Dir)).
@@ -128,7 +133,13 @@ state_output(Id, State) :-
 %   Normally, this merely runs `swipl`, but it can be necessary to embed
 %   this into some sort of script.
 
-me(Exe, Args) :-                     % swipl-win --> swipl
+me(Exe, Args, Env) :-                % User-defined hook
+    user:swipl(Exe0, Args0, Env0),
+    !,
+    Exe = Exe0,
+    Args = Args0,
+    Env = Env0.
+me(Exe, Args, Env) :-                % swipl-win --> swipl
     current_prolog_flag(executable, WinExeOS),
     prolog_to_os_filename(WinExe, WinExeOS),
     file_base_name(WinExe, WinFile),
@@ -137,26 +148,9 @@ me(Exe, Args) :-                     % swipl-win --> swipl
     file_directory_name(WinExe, WinDir),
     atomic_list_concat([WinDir, 'swipl.exe'], /, PlExe),
     prolog_to_os_filename(PlExe, Exe),
-    Args = [].
-me(Sh, Args) :-                      % Using a shell script
-    absolute_file_name('swipl.sh', MeSH,
-                       [ access(execute),
-                         file_errors(fail)
-                       ]),
-    !,
-    Sh = path(sh),
-    Args = [MeSH].
-me(Cmd, Args) :-                     % Windows: use .bat file
-    current_prolog_flag(windows, true),
-    absolute_file_name('swipl.bat', MeBat,
-                       [ access(execute),
-                         file_errors(fail)
-                       ]),
-    !,
-    Cmd = path('cmd.exe'),
-    prolog_to_os_filename(MeBat, OsName),
-    Args = ['/c', OsName].
-me(Exe, []) :-                       % What should be the normal case.
+    Args = [],
+    Env = [].
+me(Exe, [], []) :-                   % What should be the normal case.
     current_prolog_flag(executable, Exe).
 
 :- dynamic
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Can you remind me why we need all this weird stuff?    At least we can than add a decent comment explaining this ...
I vaguely recall this is because when running the tests under R the main executable is not `swipl`?   But, does creating saved state make any sense in this case?   Wouldn't it be easier to simply disable the saved state tests when running under R?

P.s.  Please rebase before making a PR rather than merging.   I.e. `git rebase origin/master` should normally do.

mgondan:
> Can you remind me why we need all this weird stuff?

Haha, my students sometimes ask the same question in the Prolog seminar :-)

> At least we can than add a decent comment explaining this ...

test_saved_states.pl allowed for some swipl.sh-hook before. I added something similar for Windows, which is nice, and now we have a unit test for it, which is also nice. As it is now, the only R-specific modification is path(sh) running swipl.sh.

The reason why I put so much effort in this is because it offers the flexibility to use swipl program in a nonstandard context (e.g., embedded in R or whatever other application comes to mind), and still be able to use 100% of its functionality. I have also seen that the unit tests of some packages do something similar as the saved state (i.e., invoke swipl from the unit test), so I hope that the unit tests in the packages can now easily be adapted. Of course, it is possible to disable the unit tests in such a scenario, but why not try to fix them instead. Finally, I really like the idea of being able to invoke the unit tests at runtime with test_installation (not at compile time with ctest).

> P.s. Please rebase before making a PR rather than merging. I.e. `git rebase origin/master` should normally do.

Will do. I should really learn git properly.

JanWielemaker:
Ok, but as I read it, we are now creating a `swipl.sh` or `swipl.bat` that merely calls what we have discovered to be the `swipl` executable.   What do we gain using that?    I'm actually wondering what this `swipl.sh` code is doing there.    I _think_ it may be related to the pre-cmake era where running `swipl` in the build environment was complicated and there was indeed a `swipl.sh` script created by `configure` that setup the environment to run `swipl` in the build environment.   CMake fixes that issue and there should also not be an issue when using `?- test_installation` as the installed version is supposed to run :smile: .

Testing saved states from an embedded SWI-Prolog is problematic as the embedding typically messes with passing the commandline arguments, which is vital for these tests.    As both creating and running the state is done by creating processes, all we need to do is to find `swipl[.exe]` and use that to create the state, no?

mgondan:
I hope it is ok if I respond within the specific context of the R package (I am still convinced that this may be generalized, e.g. to situations in which swipl is invoked with additional command line arguments).

> there should also not be an issue when using `?- test_installation` as the installed version is supposed to run 😄 .

Unfortunately, no. CRAN does not allow to include executables in R packages, so I have to delete bin/swipl and invoke `R -s -e "rswipl::swipl()" —args` … instead. That’s where the hook with swipl.sh comes in quite handy.

> embedding typically messes with passing the commandline arguments

Well, see above. The solution is a carefully constructed swipl.sh. I like to denote this as an interesting problem rather than a mess 😅

JanWielemaker:
> I hope it is ok if I respond within the specific context of the R package

Sure.  That is what we try to fix after all.

> Well, see above. The solution is a carefully constructed swipl.sh. I like to denote this as an interesting problem rather than a mess 😅

Ok.   But then I still do not see why we would create trivial swipl.sh/.bat files for the normal case.  Just finding `swipl[.exe]` should be enough.    Wrapping it in a script merely asks for trouble than fixing anything if the executable is there.   So, I think the logic should be "if `swipl[.exe]` exists, use it.   Otherwise use some script".    I could live with "If there is a script, use it.  Otherwise use `swipl[exe]`".    What problem would generating a trivial script solve though?

I could also live with a predicate that runs Prolog with a given list of arguments.   We could add e.g. process_create(prolog(swipl), ...) to run SWI-Prolog with arguments.    We can provide default rules there and a hook that allows your R code to make this work?    That way there is a general way to run Prolog itself and we can remove all this complicated code from the saved state test and I think also from some other tests and use cases.    I'm happy to add something like that and let you do the hook.

mgondan:
> I could live with "If there is a script, use it. Otherwise use `swipl[exe]`".

This is basically what we currently have, see the predicate me/2. What about the following: I remove the unit test with the script (the fourth one that I just added). Instead, I add the unit test to the documentation. Not as a unit test, but as a basic example to explain how to deal with saved states if swipl is embedded in another program.

> We can provide default rules there and a hook that allows your R code to make this work? That way there is a general way to run Prolog itself and we can remove all this complicated code from the saved state test and I think also from some other tests and use cases. I'm happy to add something like that and let you do the hook.

Do you mind postponing the decision on this last alternative? I haven't yet enough of an overview on the other tests.

JanWielemaker:
Note that we do not need a script.   Instead, through a hook, we could turn process_create(prolog(swipl), ...) into

     process_create(path('R'), ['-s', '-e',  'rswipl::swipl()', '—args'|ArgV], Options).

That should work just fine, no?   In general, the problem of running SWI-Prolog itself requires a little attention, so providing portable library support for that seems worthwhile.

mgondan:
Give me a bit of time to try it out on the different systems.

mgondan:
> That should work just fine, no?

See a proof of concept in the modified test_saved_states.pl. Is this what you mean?

I invoke it from R as follows:
````
  R <- normalizePath(file.path(R.home("bin"), "R"))
  if(.Platform$OS.type == "windows")
    R <- normalizePath(file.path(R.home("bin"), "R.exe"))
  query(call("assert", call("swipl", R, list("-s", "-e", "rswipl::swipl()", "--args"), list())))
  submit()
  clear()
````

It works, at least on Linux, but I don't foresee any problem with Windows. Obviously, the user:swipl/3 would be declared elsewhere.

JanWielemaker:
> It works, at least on Linux, but I don't foresee any problem with Windows. Obviously, the user:swipl/3 would be declared elsewhere.

This looks promising!    Just, I'd like to hook this into library(process), so there is a portable simple way run Prolog itself as a sub process.    Assuming this works, I'll design a hook and use it in the saved state generation.

JanWielemaker:
I've pushed stuff to process_create/3 and test_saved_states.pl to resolve this issue.    See https://github.com/SWI-Prolog/packages-clib/blob/d85c7865e7a421bd52fa9b0c6064d0a89de4db15/process.pl#L292 for details in how I think this should be integrated into rolog.

Hope this can now be resolved cleanly.    Closing this PR because if anything needs changing, this should be in a new PR.   Thanks for your patience.

PR #1408, opened 2025-11-22
BUILD: show options in generated file

kamahen:
This adds debug info to the generated "configure" file, so that I can verify that the options I thought I chose were actually used (that is, checking for typos).


kamahen:
There's a bug in this code. I'm preparing a PR, but I came across a few other things in the process and it's not quite ready to submit

PR #1409, opened 2025-11-22
FIXED: Make cpu_count flag return number of online processors when available

pvdabeel:
This patch improves cpu_count detection on POSIX systems.

Currently SWI-Prolog uses sysconf(_SC_NPROCESSORS_CONF) to determine the number of processors. On some systems, the number of *configured* processors may be higher than the number of *online* processors, which causes cpu_count to overestimate the available concurrency.

This change prefers sysconf(_SC_NPROCESSORS_ONLN) when the libc defines the corresponding constant. If the platform does not support _SC_NPROCESSORS_ONLN, SWI-Prolog falls back to the existing _SC_NPROCESSORS_CONF logic. No change in behaviour occurs on systems that lack ONLN.

The result is that cpu_count now reflects what the OS scheduler sees as active CPUs. On systems where CONF > ONLN this produces a more accurate cpu_count value and avoids unexpected oversubscription.

### DIFF
--- a/src/os/pl-os.c
+++ b/src/os/pl-os.c
@@ -377,13 +377,23 @@ WallTime(void)
 
 #ifdef HAVE_SC_NPROCESSORS_CONF
 
+#ifdef _SC_NPROCESSORS_ONLN
+
+int
+CpuCount(void)
+{ return sysconf(_SC_NPROCESSORS_ONLN);
+}
+
+#else
+
 int
 CpuCount(void)
 { return sysconf(_SC_NPROCESSORS_CONF);
 }
 
-#else
+#endif
 
+#else
 #ifdef PROCFS_CPUINFO
 int
 CpuCount(void)
@@ -449,7 +459,6 @@ CpuCount(void)
 #endif /*PROCFS_CPUINFO*/
 
 #endif /*HAVE_SC_NPROCESSORS_CONF*/
-
 
 void
 setOSPrologFlags(void)


JanWielemaker:
Thanks.   Merged after some cleanup to make the #ifdefs a bit more concise.

Note that I was a bit surprised that using `taskset 0x3 swipl` I still get 32 cores rather than the 2 I had hoped for.

pvdabeel:
The current change uses _SC_NPROCESSORS_ONLN, which reports the number of online CPUs -system-wide-. Taskset only modifies the affinity mask for the process, so it doesn't affect sysconf’s system-wide return value. 

If you’d like cpu_count to honour the process affinity (e.g., when using taskset or cpusets), it is possible to use sched_getaffinity(0, ...) on Linux (when available) to count the CPUs in the current process’s mask and base cpu_count on that (with _SC_NPROCESSORS_ONLN / _SC_NPROCESSORS_CONF as fallback where affinity isn’t available). It's not POSIX, and only on Linux since 2004. 

```
#ifdef __linux__
#include <sched.h>
#endif
```

With the following added to  int CpuCount(void):

```
#ifdef __linux__
  { cpu_set_t set;

    CPU_ZERO(&set);
    if ( sched_getaffinity(0, sizeof(set), &set) == 0 )
    { int cnt = 0;
      int i;

#ifdef CPU_COUNT
      cnt = CPU_COUNT(&set);
#else
      for(i=0; i<CPU_SETSIZE; i++)
      { if ( CPU_ISSET(i, &set) )
          cnt++;
      }
#endif
        return cnt;                     /* honour taskset / cpuset */
    }
  }
#endif /* __linux__ */
```

JanWielemaker:
Thanks.   Not sure that is a good idea or not.    There is already `thread_affinity/3` to query and/or set the affinity mask.

pvdabeel:
Agreed, using thread_affinity/3 is much cleaner

PR #1410, opened 2025-11-23
TEST: less strict deadlines in queue_timeout.pl

mgondan:
From time to time, my "Github Action" fails on Windows, stating "queue_timeout:abs_rel failed". I wonder if the deadlines can be changed to a bit more liberal ones.

If you dislike/want me to investigate the problem in more depth, just reject this PR. It's just a bit difficult to spot because the problem appears only rarely.

### DIFF
--- a/tests/thread/queue_timeout.pl
+++ b/tests/thread/queue_timeout.pl
@@ -50,7 +50,7 @@ test(relative,
        cleanup(message_queue_destroy(Q))
      ]) :-
 	get_time(T0),
-	\+ thread_get_message(Q, _, [timeout(0.01)]),
+	\+ thread_get_message(Q, _, [timeout(0.1)]),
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I think the problem is not so much about the times, but about verifying that the timeout happened at the right moment.   I'm not really sure which one is the problem.    I'll push some changes, turning the checks into assertion/1 calls so that we know which one is the culprit and relaxing them a little.    After all, all timeout actions should guaranteed the system waits until the timeout, so we may wait longer (with no guarantees on the deadline).   Also Windows timeouts are in milliseconds while Linux are in nanoseconds, so we may have some rounding issues here.

JanWielemaker:
Added ff8b466af8fae49c6846e263968ea12cce91d91e

PR #1412, opened 2025-11-25
Added some Cmake vars to config.h for debugging

kamahen:
You might not agree with this -- I've done it to help in understanding how some of the options are processed (I could instead create a `config_debug.h` if you  think that these changes are unnecessary. (I might find that they're not needed and would delete  them in a future PR).
 
I'll also add some similar stuff to scripts/configure and its outputs (configure and .envrc) but that PR isn't ready yet.

### DIFF
--- a/src/config.h.cmake
+++ b/src/config.h.cmake
@@ -317,3 +317,35 @@
 #cmakedefine SWIPL_TMP_DIR "@SWIPL_TMP_DIR@"
 
 #cmakedefine MSYS2 @MSYS2@
+
+/* Useful information about the compilation
+   See https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html
+
+   build types: RelWithDebInfo Release Debug
+                (defined in ../cmake/BuildType.cmake): PGO Sanitize DEB
+                not used (cmake predefined build types): MinSizeRel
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I don't see the point.   If you just want to know the values, grep through `CMakeCache.txt`.    This only would make sense if we add Prolog flags for them that are added to crash reports.   They are typically not values I'm interested in for debugging though.   Did I miss something?

kamahen:
I'm adding these so that I can output the compiler options with a benchmark test. I was thinking of adding a PL_* function to get these as a string.

But if you think it's unnecessary, I'll remove it and hope that I don't mess up my bookkeeping when I do tests. (I'm amazingly good at getting files mixed up)

JanWielemaker:
I think I prefer to avoid complicating the build for this.    As we have the relation between build dir and flags, just recording the build dir should be enough.

kamahen:
Sorry for having wasted your time. (It would have taken me a while to discover `CMakeCache.txt` on my own ... I did some web searches and all I had come up with was adding to `config.h.make` although I thought there must be another way).

JanWielemaker:
No worries :smile:

PR #1420, opened 2025-12-10
Update CMAKE.md

mgondan:
Please check (< 1 min)


JanWielemaker:
Fixed the typo and re-added the feature `make`.   I think that should deal with the issue.   E.g.,

```
mkdir build.make
cd build.make
../scripts/configure
```

PR #1421, opened 2025-12-11
FIX: UBSAN check

mgondan:
Fixes/suppresses an UBSAN issue: to reproduce, start `../scripts/configure` from build.gcc.ubsan and `make`. The issue is raised when a == NULL during QLF compilation, which happens once. 

You might perhaps check if this is desired behavior.


JanWielemaker:
Thanks.  Fixed a little different using bbd7f7a69b73fb6c8d53555c233e82dc52ec496c

PR #1423, opened 2025-12-14
UBSAN: Avoid bit shift == sizeof(x)

mgondan:
The proposed change is not pretty, but it avoids another UBSAN error if low (64 bit) is << by 64 bits. I assume that this particular piece of code does not need to run in constant time.


JanWielemaker:
> I assume that this particular piece of code does not need to run in constant time.

I doubt it makes a measurable difference.   The code is fairly time critical though as it is used in data conversion.   It is sad that the shift should be smaller than the size.   <= would have been a lot prettier.

Issue #1424, opened 2025-12-19
clang-21, ubsan


mgondan:
I get an UBSAN error with clang-21, see here:

https://win-builder.r-project.org/incoming_pretest/rswipl_10.1.0_20251218_005630/specialChecks/clang-san/process.txt

% [1/14] answer:simple ../home/hornik/tmp/CRAN_special_clang-san/rswipl.Rcheck/00_pkg_src/rswipl/src/swipl-devel/src/pl-event.c:412:11: runtime error: call to function cleanup_thread through pointer to incorrect function type 'int (*)(void *, unsigned long)'

Apparently, the clang-21 sanitizer does not like that cleanup_thread in clib/time.c is void (*)(void*) and then registered as an event handler. I am not sure either. For example, cleanup_thread has no return value. A quick fix would look like this (diff files in several places).

````
diff --git a/src/SWI-Prolog.h b/src/SWI-Prolog.h
 PL_EXPORT(bool)        PL_thread_destroy_engine(void);
-PL_EXPORT(bool)        PL_thread_at_exit(void (*function)(void *),
+PL_EXPORT(bool)        PL_thread_at_exit(int (*function)(void *, unsigned int), void *closure, bool global);
 PL_EXPORT(int) PL_thread_raise(int tid, int sig);

diff --git a/src/pl-thread.c b/src/pl-thread.c
 bool
-PL_thread_at_exit(void (*function)(void *), void *closure, bool global)
+PL_thread_at_exit(int (*function)(void *, unsigned int), void *closure, bool global)
 { GET_LD
   event_list **list = global ? &GD->event.hook.onthreadexit
                             : &LD->event.hook.onthreadexit;
-  int (*func)() = (int(*)())function;
+  // int (*func)() = (int(*)())function; // this cast is indeed a bit dubious

-  return register_event_function(list, 0, false, func, closure, 0);
+  return register_event_function(list, 0, false, function, closure, 0);
 }

diff --git a/SWI-cpp2-plx.h b/SWI-cpp2-plx.h
 PLX_EXCE(int                     , thread_destroy_engine           , (), ())
-PLX_ASIS(int                     , thread_at_exit                  , (void (*function)(void *), void *closure, int global), (function, closure, global))
+PLX_ASIS(int                     , thread_at_exit                  , (int (*function)(void *, unsigned int), void *closure, int global), (function, closure, global))
[...]

PR #1425, opened 2025-12-21
UBSAN: thread_exit function signature

mgondan:
See the problem in https://win-builder.r-project.org/incoming_pretest/rswipl_10.1.0_20251218_005630/specialChecks/clang-san/process.txt

% [1/14] answer:simple ../home/hornik/tmp/CRAN_special_clang-san/rswipl.Rcheck/00_pkg_src/rswipl/src/swipl-devel/src/pl-event.c:412:11: runtime error: call to function cleanup_thread through pointer to incorrect function type 'int (*)(void *, unsigned long)'

### DIFF
--- a/src/pl-thread.c
+++ b/src/pl-thread.c
@@ -3392,13 +3392,12 @@ still capable of running Prolog queries.
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - */
 
 bool
-PL_thread_at_exit(void (*function)(void *), void *closure, bool global)
+PL_thread_at_exit(int (*function)(void *, unsigned long), void *closure, bool global)
 { GET_LD
   event_list **list = global ? &GD->event.hook.onthreadexit
 			     : &LD->event.hook.onthreadexit;
-  int (*func)() = (int(*)())function;
 
-  return register_event_function(list, 0, false, func, closure, 0);
+  return register_event_function(list, 0, false, function, closure, 0);
 }
 
 #if O_PLMT


mgondan:
See https://github.com/SWI-Prolog/swipl-devel/issues/1424

mgondan:
See the companion PRs in clib and cpp

https://github.com/SWI-Prolog/packages-clib/pull/62
https://github.com/SWI-Prolog/packages-cpp/pull/103

(if these are merged, the test here will pass)

JanWielemaker:
This is not right.   We do not want to change the signature for PL_thread_at_exit().   We must fix the internal typing.    We must probably cast the function pointers to `void*` and to the appropriate function pointer before calling.    It is a nuisance :cry:

mgondan:
Do you mean here?

````
bool
PL_thread_at_exit(void *function, void *closure, bool global)
{ GET_LD
  event_list **list = global ? &GD->event.hook.onthreadexit
                            : &LD->event.hook.onthreadexit;
  int (*func)() = (int(*)(void*, unsigned long))function;

  return register_event_function(list, 0, false, func, closure, 0);
}
````

Edit: I guess "before calling" means that the cast is elsewhere.

JanWielemaker:
The real problem (which is indeed dubious) is that the event system calls the resgistered function as returning `int`, while it is `void`.    I think 8fc8e31d6e50cbcfe07a1a578e43af8cf9dea0ad should fix the issue.   If not, please indicate how to reproduce it.

mgondan:
After this patch, the screen is full of warnings. This is the command that I use:

````
CC="clang-21 -fsanitize=address,undefined -fno-sanitize=float-divide-by-zero -fno-sanitize=alignment -fno-omit-frame-pointer -std=gnu23" CXX="clang++-21 -fsanitize=address,undefined -fno-sanitize=float-divide-by-zero -fno-sanitize=alignment -fno-omit-frame-pointer -frtti -std=gnu++17" CFLAGS="-g -O3 -Wall -Wpedantic" LDFLAGS="-Wl,-O1" CXXFLAGS="-g -O3 -Wall -pedantic  -fpic" cmake -G "Unix Makefiles" -DUSE_GMP=OFF -S ..
````
and then `make`

The warnings (sorry for not copying them here in the first place):

````
/home/gondan/swipl-devel/src/os/pl-fmt.c:1896:3: warning: initializing 'pl_function_t' (aka 'void *') with an expression of type 'foreign_t (term_t, size_t, control_t)' (aka 'unsigned long
      (unsigned long, unsigned long, struct foreign_context *)') converts between void pointer and function pointer [-Wpedantic]
 1896 |   PRED_DEF("current_format_predicate", 2, current_format_predicate, NDET|META)
````

JanWielemaker:
Are you sure this is caused by this patch?    This patch does not change function pointer handling for foreign predicates. 

I know C does officially not support casting void* to function pointers and visa versa.    It is used all over the place though, and really hard to avoid.    Even official APIs such as dladdr() return a void * for a function pointer.   There is no C way to call this function though :cry:    Most of these could be avoided using unions, but it would break the API.

mgondan:
I think you're right, the warnings were from -Wpedantic. Sorry for the confusion.

mgondan:
The problem has indeed been solved by the above-mentioned commit. Closing this PR.

There are two more UBSAN issues, one in package utf8proc (see the issue reported there), one in cpp (I will raise it soon).

PR #1427, opened 2025-12-22
UBSAN: avoid offset to NULL pointer

mgondan:
See here, win-builder.r-project.org/incoming_pretest/rswipl_10.1.0_20251218_005630/specialChecks/clang-san/process.txt

(2nd line): % [2/11] continuation:sum ../home/hornik/tmp/CRAN_special_clang-san/rswipl.Rcheck/00_pkg_src/rswipl/src/swipl-devel/src/pl-cont.c:490:16: runtime error: applying non-zero offset 64 to null pointer


JanWielemaker:
Thanks.   Fixed a little different and did some cleanup.   See 7e855208ed8c3c37de786cf7ee38a8ca3f450774

PR #1430, opened 2025-12-29
BUILD: add CMAKE_EXTRA_FLAGS

kamahen:


kamahen:
An alternative design would be to use `$*` (or `$@` or whatever bash quotes properly) as an additional argument to the `cmake` call. This could complicate things in the future, if we want to add non-cmake parameters to `configure`; but if avoids the convention of a specially-named environment variable `$CMAKE_EXTRA_FLAGS`.

Another possibility is to make `cmake` something that can be changed by a parameter to `configure` -- but this complicates `emcmake`.

And, of course, we could have `configure --cmake_extra_flags="-DCMAKE_INSTALL_PREFIX=$HOME/.local"` but that would require adding some code to process `configure`'s args. (BTW, at Google, we had some boilerplate code for parsing args to a bash script ... maybe something similar is available as open source?)

In the end, I only want to be able to set the install prefix, and change the value of `-DTEST_PROTOBUFS_PROTOC` (which, by default, is off).

JanWielemaker:
I'm not convinced this is the route to go.   After all, configure is not the principal way to configure the system, just a shorthand to comfortably manage multiple configurations, notably for testing purposes.    Remembering the settings is too hard.  CMake itself has `CMakeUserPresets.json` for that, but that is not multi-dimensional.    The way to configure completely arbitrary systems is by using environment variables and `cmake` options (where it is pretty much undefined which combination of options actually works).

As is, there are a couple of environment variables that you can use.  I have these:

```
SWIPL_PUBLIC_NETWORK_TESTS=true
SWIPL_TEST_ODBC_DRIVER=DRIVER=SQLite3;Database=test.sqlite
SWIPL_DOC_SERVER=ON
SWIPL_INSTALL_PREFIX=/home/jan/cmake/@builddir@
SWIPL_TEST_FAIL_ON_UNLIKELY=y
SWIPL_REDIS_SERVER=localhost:6379
```

That deals the the installation location.   I normally only install the system for testing purposes.    Otherwise I use `../scripts/swipl-activate` to link a specific build directory to provide `swipl` and the other programs if I want to use a specific version for a longer time as default.

 I had a look at protobuf testing.   Basically you can configure the system using `configure` and afterwards run

      cmake -DTEST_PROTOBUFS_PROTOC=ON .

That will add this variable to `CMakeCache.txt`, making this setting permanent for that build directory.   Note that on Fedora you need to install this to make the test work:

     dnf install protobuf-compiler protobuf-devel python3-protobuf

Considering that, I'm mostly in favour of

  - Modify the tests to run in the `build` directory rather than the source
  - Use CMake or Prolog to control the test suite (no `make`, no shell as we want to maximize portability)
  - Remove `TEST_PROTOBUFS_PROTOC` and try to configure the tests unconditionally
  - Change the message if the tests cannot be configured to a normal informational message rather than a warning not to frighten people who try to install the system.

That may be too much work.  In that case just keep it as is and run the second `cmake -DTEST_PROTOBUFS_PROTOC=ON .` step.

kamahen:
Superseded by PR #1471

PR #1433, opened 2026-01-14
Use size_t for sizes (arity, no. variables, other things that follow)

mgondan:
This should suppress all warnings from -Wconversion -Wno-sign-conversion. Note that I tried to avoid casts, so I was rather liberal in changing types. "Necessary" casts are highlighted either as "safe casts" (when I was sure not to break things), or as "dubious casts" when I was not so sure.


mgondan:
Closing this.

PR #1434, opened 2026-01-14
CLEANUP: Use size_t for sizes (arity, no. variables, other things that follow)

mgondan:
This should suppress all warnings from -Wconversion (but -Wno-sign-conversion). Note that I tried to avoid casts, so I was rather liberal in changing types. "Necessary" casts are highlighted either as "safe casts" (when I was sure not to break things), or as "dubious casts" when I was not so sure.

### DIFF
--- a/src/pl-comp.c
+++ b/src/pl-comp.c
@@ -4748,7 +4748,7 @@ to skip H_VOID_N in small steps.
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - */
 
 Code
-skipArgs(Code PC, ssize_t skip, int *in_hvoid)
+skipArgs(Code PC, ssize_t skip, ssize_t *in_hvoid)
 { int nested = 0;
   Code nextPC;
 
@@ -4758,7 +4758,7 @@ skipArgs(Code PC, ssize_t skip, int *in_hvoid)
     { skip -= *in_hvoid;
       *in_hvoid = 0;
     } else
-    { (*in_hvoid) -= (int) skip; /* very dubious cast */
+    { (*in_hvoid) -= skip;
       if ( *in_hvoid == 0 )
 	return stepPC(PC);
       return PC;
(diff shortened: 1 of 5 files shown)


JanWielemaker:
Thanks.   That is a lot.    Some of it is clearly an improvement, notably type changes to local variables.    Various others are less clear.    Some make very frequently used structs larger.   It is not clear whether we want that.   Some change the types of the public API.   That is not acceptable as it does not allow merging into stable and breaks the binary API for no good reason (breaks ctype defined bindings used for e.g. Python, JavaScript and Rust).

I'm a bit unsure how to continue.    One option I see is that I review the changes to struct and function prototype declarations and changes these as I see fit.   For some of them it might be best to introduce a new type rather than  using a standard type.   That documents the issues cleanly and makes it easier adjust to changing demands.   After that, everything that depends on this needs to be changed again ...

What do you think?

mgondan:
In a first round, I’ll remove all changes in SWI-Prolog.h, that’s the public API, isn’t it?

Then we can move on with the discussion. There’s no rush.

JanWielemaker:
> In a first round, I’ll remove all changes in SWI-Prolog.h, that’s the public API, isn’t it?

I think these are minor and I might opt for new defined types (binary compatible with the current).    I think it is better if I decide on the types for (mostly) public functions and struct fields.   Just, after that more changes are required :cry:   Are you willing to look at that?

mgondan:
Of course, go ahead! It’s an honor

JanWielemaker:
Just for the record so we won't forget: the benchmark `bench/programs/sieve.pl` is broken.   It still runs, but 10 times slower and all extra time seems to goes into retract/1.     The result looks correct.

mgondan:
Something in pl-index.c/pl-index.h. I'll try to find it.

mgondan:
sieve is working again.

````
   key ^= key >> shift;
-  return (unsigned int) (fib64*key) >> shift; /* dubious cast */
+  return (unsigned int) ((fib64*key) >> shift); /* less dubious */
 }
````
"Der Mensch bleibt ein Laie." (Max Frisch)

JanWielemaker:
Thanks!   That saves me some time :smile:

mgondan:
Fixed by https://github.com/SWI-Prolog/swipl-devel/commit/9cb544583d24c9fffffa2769b9d79a06f9b4b93c

PR #1438, opened 2026-01-31
Fix uninitialized variable 'dts' in pl-trie.c

EricGT:
## Summary

Fixes MSVC Debug Runtime Check #3 error: "The variable 'dts' is being used without being initialized"

The issue occurs in `add_choice()` in `src/pl-trie.c` at lines 2243-2250, where `popSegStack()` return value was ignored. When the stack is empty at a `TN_KEY_POP` marker, `popSegStack` returns false and `dts` remains uninitialized, but the code was accessing `dts.term` and `dts.size` regardless.

## Changes

- Added conditional check around `dts` usage to only access fields when `popSegStack()` returns true
- This prevents undefined behavior when the stack is empty at a pop marker

## Tests Affected

Previously triggered MSVC Debug error dialog on:
- `xsb_sub:rrtc8` - Right-recursive transitive closure with compound terms
- `xsb_sub:drtc8` - Double-recursive transitive closure with compound terms

## Test Results

After fix, all 36 XSB subsumption tests pass without errors:

100% tests passed, 0 tests failed out of 1
Test #27: swipl:xsb/sub_tests ..............   Passed    1.88 sec

## Platform

- Windows 11
- MSVC (Visual Studio 2026)
- Debug build configuration
- All tests pass with no runtime check failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pl-trie.c
+++ b/src/pl-trie.c
@@ -2242,11 +2242,12 @@ add_choice(DECL_LD trie_gen_state *state, descent_state *dstate, trie_node *node
 
 	  if ( IS_TRIE_KEY_POP(children.key->key) && dstate->compound )
 	  { desc_tstate dts;
-	    popSegStack(&dstate->stack, &dts, desc_tstate);
-	    dstate->term = dts.term;
-	    dstate->size = dts.size;
-	    DEBUG(MSG_TRIE_GEN,
-		  Sdprintf("Popped %p, left %zd\n", dstate->term, dstate->size));
+	    if ( popSegStack(&dstate->stack, &dts, desc_tstate) )
+	    { dstate->term = dts.term;
+	      dstate->size = dts.size;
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Ok.  Fixed with b40bcb11df3fac5e79a799b9424e5037d425715e.   This follows your patch, after some cleanup on bool types and layout while scanning through the code.    The main difference is that in debug mode I set the dstate variables to a value that will cause a crash when accessed.    That should guarantee that in the case there is also a semantic issue involved we'll crash rather that producing the wrong answer.  I didn't manage to fully understand why the popSegStack() can fail, i.e., why there is no guarantee something is still on the stack.

EricGT:
>  I didn't manage to fully understand why the popSegStack() can fail

The attached file might offer a clue.

PR #1443, opened 2026-02-01
WIP: Partial mitigation for shared dynamic clause race condition

EricGT:
IMPORTANT: This does NOT fully fix the underlying race condition.

This PR is being shared for visibility and discussion of the approach, but is not expected to be merged as-is.

This change reduces the likelihood of race condition failures by:
- Setting generation values BEFORE adding clause to indexes
- Using memory barriers to ensure proper visibility ordering
- Writing erased before created with release semantics

The race can still occur under certain timing conditions. A complete fix requires further synchronization work.

### DIFF
--- a/src/pl-proc.c
+++ b/src/pl-proc.c
@@ -1412,9 +1412,6 @@ assertDefinition(DECL_LD Definition def, Clause clause, ClauseRef where)
     return PL_no_memory(),NULL;
   }
 
-  clause->generation.created = max_generation(def);
-  clause->generation.erased  = 1;
-
   LOCKDEF(def);
   acquire_def(def);
   if ( !def->impl.clauses.last_clause )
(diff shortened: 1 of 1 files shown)


JanWielemaker:
I'm not convinced this fixes anything.    I know there are (pretty rare) issues on some platforms.    The logic now is though to set the created generation to the max and the erased to 1.   That means the clause is invisible.   Next we set the created to the real generation, meaning the clause is still invisible and finally we set the erased to the max generation, making the clause atomically visible.   Possibly we do need a MEMORY_RELEASE() between setting the created and erased.   

I don't think I'm going to rely on AI here ...

EricGT:
> I'm not convinced this fixes anything.

It doesn’t.

You asked to share my work, and this was one of the items on the list.
Had I not posted about the ASan problem
<https://swi-prolog.discourse.group/t/windows-native-build-with-asan-claude-ai-notes-asan-found-a-heap-use-after-free-bug/9500>,
I would have continued investigating and fixing bugs quietly.
>  I don't think I'm going to rely on AI here ...

Fully agree.

This is one of those cases where working directly with you, the domain
expert, is more effective than relying on AI.
---

 When running test_shared_dynamic.pl using Windows MSVC generated code, the
test would often fail multiple times within a single test run:

   29: .failed(26554)
   29: failed(52451)
   29: failed(170254)
   29: failed(180441)
   29: Script c:/dev-msvc-pr/swipl-devel/tests/thread/test_shared_dynamic.pl
failed
   29: ..

These changes are not even present in the version I use for submitting PRs,
specifically to avoid introducing additional variables or potential
problems.

For comparison, I ran the same test using the downloaded Windows swipl.exe,
and it did not report a single failure.

Exe path:  C:\Program Files\swipl\bin\swipl.exe
Test path:  C:\dev-MSVC-PR\swipl-devel\tests\thread\test_shared_dynamic.pl

C:\dev-MSVC-PR\swipl-devel>"C:\Program Files\swipl\bin\swipl.exe" -g
"test_shared_dynamic(1, 2000000)" -t halt
C:\dev-MSVC-PR\swipl-devel\tests\thread\test_shared_dynamic.pl

C:\dev-MSVC-PR\swipl-devel>"C:\Program Files\swipl\bin\swipl.exe" -g
"test_shared_dynamic(10, 2000000)" -t halt
C:\dev-MSVC-PR\swipl-devel\tests\thread\test_shared_dynamic.pl
% The following threads wouldn't die: [gc]

C:\dev-MSVC-PR\swipl-devel>"C:\Program Files\swipl\bin\swipl.exe" -g
"test_shared_dynamic(10, 2000000)" -t halt
C:\dev-MSVC-PR\swipl-devel\tests\thread\test_shared_dynamic.pl
% The following threads wouldn't die: [gc]

C:\dev-MSVC-PR\swipl-devel>"C:\Program Files\swipl\bin\swipl.exe" -g
"test_shared_dynamic(20, 2000000)" -t halt
C:\dev-MSVC-PR\swipl-devel\tests\thread\test_shared_dynamic.pl
% The following threads wouldn't die: [gc]

---

Feel free to close the PR.🙂

JanWielemaker:
Although I do not agree with the PR, it still may hold important clues.   Luckily I added your repo as remote to my development tree, so I can still access that branch.

EricGT:
> Luckily I added your repo as remote to my development tree, so I can still access that branch.

One benefit about Claude is that if you set the model on Opus (Anthropic most advanced model), turn thinking on and give it lots of specific details such as feedback when debugging the code with Visual Studio IDE it is pretty consistent about the code changes it suggest.

FYI

Currently stepping though the code with the cleanup. Nothing yet but have not been solely focused on this.

PR #1444, opened 2026-02-02
Exclude LeakSanitizer code on Windows

EricGT:
MSVC ASan does not include LSan. Guard LSan headers and functions with !defined(__WINDOWS__).

### DIFF
--- a/src/pl-fli.c
+++ b/src/pl-fli.c
@@ -71,7 +71,8 @@
 #include "pl-nt.h"
 #endif
 
-#ifdef __SANITIZE_ADDRESS__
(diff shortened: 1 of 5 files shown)


JanWielemaker:

https://swi-prolog.discourse.group/t/windows-native-build-with-asan-claude-ai-notes-asan-found-a-heap-use-after-free-bug/9500/44

JanWielemaker:
Ok.  Merged, though using a cmake check for the lsan header rather than hard encoding (for example, clang on Windows has this header).   Also left `__asan_default_options()`  unconditionally.  It is just a function that is used by some versions of Asan to get configuration defaults.    If it exists but is not known to your asan it should do no harm.

PR #1445, opened 2026-02-08
PORT: Fix MSVC C5286 enum conversion warnings in ar_maxr/ar_minr

EricGT:
Cast cmpReals() return to int in switch expressions to avoid implicit
enum conversion warnings (cmp_t to cmpex_t) on MSVC.


JanWielemaker:
Fixed by using cmpex_t enum values (as it should be).

EricGT:
PS

Commit used:
https://github.com/SWI-Prolog/swipl-devel/commit/fd522f14bbe1d64a2aaed64e6649e0544f4b2904

PR #1446, opened 2026-02-08
CLEANUP: Fix MSVC C4133 type mismatch warnings in pl-dict.c

EricGT:
Change return type of `compare_dict_entry()` and the outer
`compare_term_refs()` wrapper from `cmp_t` to `int` to match the
`sort_r()` function pointer signature declared in `pl-rsort.h`.

The `cmp_t` enum values `{-1, 0, 1}` fit in `int`, and this matches
the pattern already used by `cmp_dict_index()` in the same file.

Silences on MSVC:
```
pl-dict.c(274,3): warning C4133: 'function': incompatible types - from 'cmp_t (__cdecl *)(...)' to 'int (__cdecl *)(...)'
pl-dict.c(319,39): warning C4133: 'function': incompatible types - from 'cmp_t (__cdecl *)(...)' to 'int (__cdecl *)(...)'
```

### DIFF
--- a/src/pl-dict.c
+++ b/src/pl-dict.c
@@ -251,7 +251,7 @@ dict_ordered(DECL_LD Word data, size_t count, Word dupl)
 }
 
 
-static cmp_t
+static int
 compare_dict_entry(const void *a, const void *b, void *arg)
 { Word p = (Word)a+1;
   Word q = (Word)b+1;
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Fixed with 01ef10953fa5fff8ee75d035b3ad4d25881da501.  Should be the same, but all the merge commits make this PR hard to merge.

PR #1447, opened 2026-02-10
PORT: Fix MSVC C4334 warning in pl-buffer.c

EricGT:
Cast `1` to `size_t` before left-shifting by `k` to avoid implicit 32-bit to 64-bit conversion when used as an array index on x64.

```
pl-buffer.c(161,46): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
```

### DIFF
--- a/src/os/pl-buffer.c
+++ b/src/os/pl-buffer.c
@@ -158,7 +158,7 @@ popStringBuffer(string_stack *stack)
   if ( __builtin_popcount(stack->top) == 1 && stack->top > 4 )
   { unsigned int i;
     unsigned int k = MSB(stack->allocated);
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Merged, only the pl-buffer.c change.   Note that you can set personal git ignore as described in https://stackoverflow.com/questions/5724455/can-i-make-a-user-specific-gitignore-file

Your PRs have a lot of merge commits.   Probably your history is not the same as mine any more.    Make sure you preserve (know the hashes of) commits you want to preserve and then run

     git checkout master
     git fetch
     git reset --hard origin/master

That looses local changes, but gets your master in sync with mine.  Then re-apply local changes (preferable in a branch).

PR #1448, opened 2026-02-10
BUILD: Set CMAKE_MSVC_RUNTIME_LIBRARY globally for MSVC builds

EricGT:
Static libraries built as subdirectories (e.g., `libstemmer` in
packages-nlp, `edit` in winlibedit) were compiled without an explicit
CRT setting, defaulting to `/MT` (static CRT). The plugin DLLs linking
them use `/MD` (dynamic CRT), causing LNK4098 linker warnings:

```
LINK : warning LNK4098: defaultlib 'LIBCMT' conflicts with use of other libs
```

Set `CMAKE_MSVC_RUNTIME_LIBRARY` globally in `cmake/port/Windows.cmake`
to ensure all targets use the dynamic CRT consistently. Enable policy
CMP0091 (NEW) so the variable is respected by all subprojects,
including those with `cmake_minimum_required` below 3.15.


JanWielemaker:
Merged (squashed).    I assume this makes the packages-nlp PR redundant.

PR #1453, opened 2026-02-11
PORT: Use C99 variadic macro for Sdprintf in O_DEBUG mode

EricGT:
## Summary
- Replace GCC-style variadic macro syntax (`fmt...`) with C99 standard (`...` / `__VA_ARGS__`) for the `Sdprintf` macro in `pl-builtin.h`
- The GCC extension `fmt...` is not supported by MSVC, causing error C2608 when compiling with `O_DEBUG` enabled

## Files Changed
- `src/pl-builtin.h` — Change `#define Sdprintf(fmt...)` to `#define Sdprintf(...)`

## Test Plan
- [x] Build with `--config Sanitize` on MSVC completes without errors
- [x] Tests pass under Sanitize configuration

### DIFF
--- a/src/pl-builtin.h
+++ b/src/pl-builtin.h
@@ -517,7 +517,7 @@ is also printed if stdio is not available.
 
 /* We want to use the version of Sdprintf with the debug channel, if possible */
 #undef Sdprintf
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Was already applied

PR #1454, opened 2026-02-11
PORT: Use C99 variadic macro for Sdprintf in pl-stream.c

EricGT:
## Summary
- Replace GCC-style variadic macro syntax (`fmt...`) with C99 standard (`...` / `__VA_ARGS__`) for the `Sdprintf` macro in `pl-stream.c`
- The GCC extension `fmt...` is not supported by MSVC, causing error C2608 when compiling with `O_DEBUG` enabled

## Files Changed
- `src/os/pl-stream.c` — Change `#define Sdprintf(fmt...)` to `#define Sdprintf(...)`

## Test Plan
- [x] Build with `--config Sanitize` on MSVC completes without errors
- [x] Tests pass under Sanitize configuration

### DIFF
--- a/src/os/pl-stream.c
+++ b/src/os/pl-stream.c
@@ -2831,7 +2831,7 @@ Sdprintf_ex(const char *channel, const char *file, int line, const char *fm, ...
   return rval;
 }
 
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Was already applied

PR #1457, opened 2026-02-11
BUILD: Fix MSVC MSB8065 warnings for custom build outputs

EricGT:
Touch OUTPUT files on MSVC to satisfy MSBuild's requirement that custom command outputs must exist.

### DIFF
--- a/cmake/LibIndex.cmake
+++ b/cmake/LibIndex.cmake
@@ -2,8 +2,10 @@ include(QLF)
 
 # Rebuild all library indexes. First build is   always  ok, but as we do
 # not know all dependencies for each library index and these may change,
-# we wish to re-run these on every build  run. We do this by setting the
-# OUTPUT to a dummy file that is not really created.
+# we wish to re-run these on every build run. The OUTPUT is set to a dummy
+# __INDEX.pl file (touched after build) while the real INDEX.pl is declared
+# as a BYPRODUCT. This forces CMake to rebuild indexes while avoiding
+# MSB8065 warnings on Visual Studio.
 
 function(library_index)
 
@@ -18,6 +20,7 @@ foreach(dir ${ARGN})
 	NOINSTALL
 	QUIET
 	OUTPUT ${SWIPL_BUILD_HOME}/${dir}/__INDEX.pl
+	BYPRODUCTS ${SWIPL_BUILD_HOME}/${dir}/INDEX.pl
 	COMMAND "make_library_index('${SWIPL_BUILD_HOME}/${dir}')"
 	COMMENT "Build home/${dir}/INDEX.pl")
     install(FILES ${SWIPL_BUILD_HOME}/${dir}/INDEX.pl
--- a/cmake/QLF.cmake
+++ b/cmake/QLF.cmake
@@ -65,16 +65,38 @@ function(add_swipl_target name)
     set(ENV_SEP)
   endif()
 
-  add_custom_command(
-      OUTPUT ${my_OUTPUT}
-      BYPRODUCTS ${my_BYPRODUCTS}
-      COMMAND ${CMAKE_COMMAND} -E env --unset=DISPLAY SDL_VIDEODRIVER=dummy
-	      ${ENV_SEP} ${PROG_SWIPL} ${options} ${my_OPTIONS}
-      COMMENT "${my_COMMENT}"
-      WORKING_DIRECTORY "${my_WORKING_DIRECTORY}"
-      DEPENDS core prolog_home
-              ${SWIPL_COMMAND_DEPENDS} "${my_DEPENDS}"
-      VERBATIM)
+  # On MSVC, add a second command to touch the OUTPUT file to avoid MSB8065 warnings
+  # (some OUTPUT files are dummy files that are not actually created by the command)
+  if(MSVC)
+    # Convert relative OUTPUT paths to absolute (resolved relative to CMAKE_CURRENT_BINARY_DIR)
+    list(GET my_OUTPUT 0 output_file)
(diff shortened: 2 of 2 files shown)


EricGT:
Let me know if there are any problems so I can fix the PR format.
Thanks, appreciate it. :)

JanWielemaker:
I don't see any problems with the PR itself :smile:   Should have read through before the previous reply ...  But, I think this fixes nothing.    Now we just have an additional file that we do not want to have.     We can achieve the same by not using the dummy file at all and simply assign the real file as output, which also gets rid of the dummy file.   Then there is no warning, but also no proper update.

To get that we need to get the dependencies for creating `INDEX.pl`.   That requires more serious rewrites.   It might be doable since each package has its own library rather than all packages adding files to the same library as it was some years ago.    Simplest would be if you could express that `INDEX.pl` depends on all `[a-z]*.pl` files in the directory, but I think that is not possible.   If you want to have a look at that, be my guest :smile:

EricGT:
> I think this fixes nothing.

It did fix about 40 warnings all like

warning MSB8065: Custom build for item "C:\dev-MSVC-PR\swipl-devel\build\CMakeFiles\8652276d6a607c7134934d525631549b\__INDEX.pl.rule" succeeded, but specified output "c:\dev-msvc-pr\swipl-devel\build\home\library\ext\pdt\__index.pl" has not been created. This may cause incremental build to work incorrectly. [C:\dev-MSVC-PR\swipl-devel\build\packages\PDT\library_index_library_ext_PDT.vcxproj]

Rebuilt the code and reran to ensure that the warnings were gone. As it is a warning, not losing sleep if we have to look for better ways to handle this. 

> Simplest would be if you could express that `INDEX.pl` depends on all `[a-z]*.pl` files in the directory, but I think that is not possible. If you want to have a look at that, be my guest 😄

Thanks.

I'll put that on the to do list.
Will pull out the change on my side as it is only a warning, so that I can work on a more important error.

I'll close this without accepting the changes as it is just a warning.

JanWielemaker:
Actually, this had a pretty easy fix.   Try 988b6496a00497d1ac8d6258aca11673bc139dac

EricGT:
Excellent timing. I just finished cleaning up all of the Git loose ends on
my side and was in sync with
https://github.com/SWI-Prolog/swipl-devel/commits/master/ except for this
change.

Will resync and test all the versions (release, debug, ...) to see if the
warnings are gone. The debug version still has an error in the cmake make
step but the warnings appear then, not in ctest.

EricGT:
Your fix removed all but one.

---

PrepareForBuild:
  Creating directory "x64\Release\library_qlf\".
  Structured output is enabled. The formatting of compiler diagnostics will
reflect the error hierarchy. See https://aka.ms/cpp/structured-output for
more details.
  Creating directory "x64\Release\library_qlf\library_qlf.tlog\".
InitializeBuildStatus:
  Creating "x64\Release\library_qlf\library_qlf.tlog\unsuccessfulbuild"
because "AlwaysCreate" was specified.
  Touching "x64\Release\library_qlf\library_qlf.tlog\unsuccessfulbuild".
CustomBuild:
  "The build of
'C:\dev-MSVC-PR\swipl-devel\build\CMakeFiles\aa4ecf01a6e1497583939298e7055cd0\library_qlf.rule'
depends on 'C:\DEV-MSVC-PR\SWIPL-DEVEL\BUILD\SRC\__LIBRARY_QLF_FILES__'
which is produced by the build of
'C:\dev-MSVC-PR\swipl-devel\build\CMakeFiles\2db26cfb4f2282474051a5b13bb16753\__library_QLF_files__.rule'.
The items cannot be built in parallel."
  QLF compile library
  Building Custom Rule C:/dev-MSVC-PR/swipl-devel/src/CMakeLists.txt
C:\Program Files\Microsoft Visual
Studio\18\Community\MSBuild\Microsoft\VC\v180\Microsoft.CppCommon.targets(254,5):
warning MSB8065: Custom build for item
"C:\dev-MSVC-PR\swipl-devel\build\CMakeFiles\2db26cfb4f2282474051a5b13bb16753\__library_QLF_files__.rule"
succeeded, but specified output
"c:\dev-msvc-pr\swipl-devel\build\src\__library_qlf_files__" has not been
created. This may cause incremental build to work incorrectly.
[C:\dev-MSVC-PR\swipl-devel\build\src\library_qlf.vcxproj]
FinalizeBuildStatus:
  Deleting file
"x64\Release\library_qlf\library_qlf.tlog\unsuccessfulbuild".
  Touching
"x64\Release\library_qlf\library_qlf.tlog\library_qlf.lastbuildstate".
Done Building Project
"C:\dev-MSVC-PR\swipl-devel\build\src\library_qlf.vcxproj" (default
targets).

---

When I was creating the patch I sent, I also had just one that needed more
work. Can not be sure if this was the one but do remember it took a bit
more work to get them all.

This warning occurred doing a CMake --build for  release .
Will keep an eye on the fix for the other types of builds.

JanWielemaker:
That one is about generating the .QLF files.    That suffers from similar problems than the INDEX.pl files did, but needs a rather different approach to solve it.    Fixing it the way you tried is (again) no option as that exactly causes incremental builds to fail.   What this tries to do is to run some process after everything else is completed by claiming that this process creates some file but it does not, so the build system keeps running it.    Possibly modern CMake has a more elegant solution.   As is, in invokes one Prolog task and gives Prolog the responsibility to generate all outdated .qlf files in the right order.

EricGT:
If this last one still remains after fixing the errors I will come back to
it and give Claude and ChatGPT some more info and deeper thinking to see
what they can come up with.

Issue #16, opened 2014-10-07
Undefined procedure: prolog_ide_stack:predsort/3


cbarrick:
I get this error when using `gtrace/0` to debug code on Mac OS X in version 7.1.23.

Query:

``` prolog
?- gtrace, main.
```

Error:

```
ERROR: prolog_ide_stack:display_levels/3: Undefined procedure: prolog_ide_stack:predsort/3
```

Goal stack:

```
 [65] generate_current_predicate(_G1929, _G1926, predsort)
 [58] setof(_G1926:predsort, '$dwim':current_predicate(_G1929, _G1926:predsort), _G2222)
 [54] setup_call_catcher_cleanup('$bags':'$new_findall_bag', '$bags':findall_loop(_G1904, '$messages':dwim_predicate(prolog_ide_stack:predsort, _G1904), _G1914, []), _G2222, '$bags':'$destroy_findall_bag')
 [44] print_message(error, error(existence_error(procedure, prolog_ide_stack:predsort/3), context(prolog_ide_stack:display_levels/3, _G1871)))
 [42] pce_principal:send(@17582044230985/prolog_debugger, show_stack([frame(708, call), frame(683, 52), frame(658, 52), frame(633, 52), frame(608, 52), frame(..., ...)|...], []))
```

C stack:

```
C-stack trace labeled "INT":
  [0] [...]/lib/swipl-7.1.23/lib/x86_64-darwin14.0.0/libswipl.dylib(save_backtrace+0x65) [0x104b1169f]
  [1] [...]/lib/swipl-7.1.23/lib/x86_64-darwin14.0.0/libswipl.dylib(interruptHandler+0x230) [0x104acdaac]
  [2] [...]/lib/swipl-7.1.23/lib/x86_64-darwin14.0.0/libswipl.dylib(dispatch_signal+0x371) [0x104ac969c]
  [3] [...]/lib/swipl-7.1.23/lib/x86_64-darwin14.0.0/libswipl.dylib(handleSignals+0x84) [0x104ac92c2]
  [4] [...]/lib/swipl-7.1.23/lib/x86_64-darwin14.0.0/libswipl.dylib(PL_next_solution+0x3c78) [0x104a70845]
  [5] [...]/lib/swipl-7.1.23/lib/x86_64-darwin14.0.0/libswipl.dylib(PL_call_predicate+0x1c) [0x104a7f30e]
[...]

PR #1458, opened 2026-02-12
TEST: Windows pipe test by using absolute batch file path

EricGT:
The pipe:cat1 test was failing on Windows because the batch file path was relative. cmd /c couldn't locate pltest-XXXX.bat. Fixed by using working_directory/2 to get absolute path and quoting it in the cmd invocation. Test results: All 89 tests now pass.

### DIFF
--- a/tests/files/test_pipe.pl
+++ b/tests/files/test_pipe.pl
@@ -69,12 +69,14 @@ test(cat1) :-
     format(atom(File), 'pltest-~w.txt', [Pid]),
     (   current_prolog_flag(windows, true)
     ->  format(atom(Bat), 'pltest-~w.bat', [Pid]),
+        working_directory(CWD, CWD),
+        atomic_list_concat([CWD, '/', Bat], BatPath),
         setup_call_cleanup(
             open(Bat, write, Fd1),
             writeln(Fd1, '@findstr .* > %1'),
             close(Fd1)),
-        format(atom(Cmd), 'cmd /c ~w ~w', [Bat, File]),
-        debug(pipe, 'Created BAT script ~q', [Bat])
+        format(atom(Cmd), 'cmd /c "~w" ~w', [BatPath, File]),
+        debug(pipe, 'Created BAT script ~q', [BatPath])
     ;   format(atom(Cmd), 'cat > ~w', [File])
     ),
     Text = 'Hello World',
@@ -89,8 +91,8 @@ test(cat1) :-
         close(Fd3)),
     debug(pipe, 'Read ~q to "~s"', [File, String]),
     delete_file(File),
-    (   nonvar(Bat)
-    ->  delete_file(Bat)
+    (   nonvar(BatPath)
+    ->  delete_file(BatPath)
     ;   true
     ),
     !,


EricGT:
Used a Claude skill this time. It did the checks and created a much simpler explanation.

Left me know if there are any problems so I can update the skill as needed.

JanWielemaker:
Merged after replacing working_directory/2 logic with absolute_file_name/2.

EricGT:
The reason for the change is that I created a skill to run the various builds directly from within a Claude conversation.

Most AI vendors on Windows now default to PowerShell for script execution rather than `.bat` files. The skill therefore relies on a PowerShell script to perform the bulk of the work. This also reduces token usage, since the AI does not need to walk through each individual step interactively.

I am not yet certain whether your modification will function as intended, but I will test and report back.

JanWielemaker:
> Left me know if there are any problems so I can update the skill as needed.

The PR is well formulated.   It is not clear to me why we need an absolute path, but if it works for you I'm fine with it.  Just checked my cross-compilation build.  That is still in good shape.


> I am not yet certain whether your modification will function as intended, but I will test and report back.

They should.  It does more or less the same, except that it does path normalization, so absolute_file_name('../../x', F) gives you a path without ../.. in the middle.

EricGT:
> It is not clear to me why we need an absolute path

The ctest error was

16: % [2/3] pipe:cat1 ..'pltest-30424.bat' is not recognized as an internal or external command,
16: operable program or batch file.
16: ............................... **FAILED (0.045 sec)
16: ERROR: [Thread main] c:/dev-msvc-pr/swipl-devel/tests/files/test_pipe.pl:67:
16: ERROR: [Thread main]     test pipe:cat1: received error: open/3: source_sink `'pltest-30424.txt'' does not exist (No such file or directory)
16: % [3/3] pipe:cat2 ................................... passed (0.040 sec)
16: ERROR: [Thread main] 1 test failed
16: Script c:/dev-msvc-pr/swipl-devel/tests/files/test_pipe.pl failed
16:  done
16: *** 1 tests failed ***
16/89 Test #16: swipl:files ......................***Failed    0.85 sec

---

Up to this most recent test, the ctest commands were always executed from a Windows cmd.exe session initialized with:

"C:\Program Files\Microsoft Visual Studio\18\Community\VC\Auxiliary\Build\ [vcvars64.bat](https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170)"

This ensures that the MSVC build environment is correctly configured before invoking CMake or CTest.

In the recent run with the error, the build is initiated from a PowerShell script launched by Claude via a skill. It is not clear how Claude configures the execution environment or sandbox for that PowerShell session. 

The PowerShell script then initializes the MSVC environment using [Launch-VsDevShell.ps1.](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=visualstudio), which subsequently invokes ctest.

This represents a change in both the shell and environment initialization path compared to previous runs, and that difference may be relevant to the observed behavior.

EricGT:
The code works via the Claude -> PowerShell sequence.

16: % [1/3] pipe:pwd .................................... passed (0.037 sec)
16: % [2/3] pipe:cat1 ................................... passed (0.069 sec)
16: % [3/3] pipe:cat2 ................................... passed (0.037 sec)

PR #1460, opened 2026-02-13
CLEANUP: Avoid unreferenced variable warning in blocked_unload_shared_objects()

EricGT:
Scope 'char *s' inside #ifndef __WINDOWS__ where it is used, fixing MSVC warning C4101.


JanWielemaker:
Merged after some simplification.

PR #1462, opened 2026-02-14
FIXED: Debug bootfile library autoload dependency

EricGT:
Adds prolog_home dependency to bootfile step. Fixes library autoload failures in Debug builds where O_DEBUG causes boot/tabling.pl to require library(debug).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -367,6 +367,13 @@ add_custom_command(
 )
 add_custom_target(bootfile DEPENDS ${SWIPL_BOOT_FILE} ${SWIPL_ABI_FILE})
 
+# Debug builds define O_DEBUG, which sets the prolog_debug flag,
+# causing boot/tabling.pl to autoload library(debug).  Without this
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Merged after some shortening and adding another small issue with using library(debug) in a boot file.

EricGT:
> Merged after some shortening

Thanks.

Will update my skill so that C code comments should be short one line
comments, preferably on the same line after the code.
They should be more of a memory jog than an explanation.

JanWielemaker:
The commit message and inline comments should be fairly brief.   The PR should have a proper description for me to judge it though.

EricGT:
> The PR should have a proper description for me to judge it though.

I was hoping the test would help. I wanted to put the fix and the test
together but you like separate PRs.

Do you prefer test(s) with the fix in the same PR?

JanWielemaker:
> Do you prefer test(s) with the fix in the same PR?

Yes.   That is, if they are related of course.

PR #1463, opened 2026-02-14
FIXED: PThreads4W exit crash in MSVC Debug builds

EricGT:
Prevents DllMain from accessing freed thread state during process exit. Details in commit message.

### DIFF
--- a/src/pl-init.c
+++ b/src/pl-init.c
@@ -1918,6 +1918,16 @@ emergency:
     memset(&PL_local_data,  0, sizeof(PL_local_data));
   }
 
+#if defined(__WINDOWS__) && defined(O_PLMT)
+  /* Explicitly clean up PThreads4W before process exit.  Without
+     this, PThreads4W's DllMain(DLL_PROCESS_DETACH) accesses freed
+     thread data, causing ACCESS_VIOLATION in Debug builds where the
+     CRT fills freed heap with 0xDD.  Calling this after all Prolog
+     threads have terminated and memory has been reclaimed makes
+     DllMain a no-op. */
+  pthread_win32_process_detach_np();
+#endif
+
 #if defined(__SANITIZE_ADDRESS__) && defined(HAVE_SANITIZER_LSAN_INTERFACE_H)
   char *s;
 


JanWielemaker:
Combined with the test PR.   Also moved detection of this function to CMake as the MinGW thread runtime does not define this and this it breaks the build.   Finally adjusted the conditions for running this test as the `clib` and `http` packages need not be there.

EricGT:
FYI

Synced code up to
https://github.com/SWI-Prolog/swipl-devel/commit/a97b0b1c844cc5db455767ac0bcc0548be94202a

A release build did not have any build errors, debug now has a build error,
as such the CTest did not run.

JanWielemaker:
> debug now has a build error

What is the error?   I need to check, but traditionally we did not use the generated `config.h` (when we were using GNU autoconf) because Windows could not run that.  With CMake it is generated, but I think we still use the hand crafted version.   I'll have a look.

JanWielemaker:
> I'll have a look.

Seems ok.    It includes `config/wincfg.h`, which sets some stuff, but also includes the normal `config.h`.   So, in theory ( :smile:  ) it should work ...

EricGT:
What is the error?

Claude narrowed it down to the check for PThread4W, the Code is fine.

It suggested a change which worked to build a release and debug version

if(WIN32 AND PThreads4W_FOUND)
  set(HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP 1)else()
  check_function_exists(pthread_win32_process_detach_np
HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)endif()

which I did not like.

In checking CMake documentation this was suggested.

Now testing with

check_symbol_exists(pthread_win32_process_detach_np "pthread.h"
HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)

JanWielemaker:
> which I did not like.

Agree

> check_symbol_exists

If that works I'm ok with it.

EricGT:
>  What is the error?

If you need the details

Using
cmake .. -G Visual Studio 18 2026 -A x64
-DCMAKE_TOOLCHAIN_FILE=C:\dev\vcpkg/scripts/buildsystems/vcpkg.cmake
-DPython_ROOT_DIR=C:\Users\Eric\AppData\Local\Programs\Python
-DBDB_LIBRARY=C:\dev\vcpkg/installed/x64-windows/lib/libdb48.lib
cmake --build . --config Debug --verbose

Just this one error

"C:\dev-MSVC-PR\swipl-devel\build\ALL_BUILD.vcxproj" (default target) (1) ->
"C:\dev-MSVC-PR\swipl-devel\build\doc.vcxproj" (default target) (172) ->
"C:\dev-MSVC-PR\swipl-devel\build\man_index.vcxproj" (default target) (320)
->
(CustomBuild target) ->
  C:\Program Files\Microsoft Visual
Studio\18\Community\MSBuild\Microsoft\VC\v180\Microsoft.CppCommon.targets(254,5):
error MSB8066: Custom build for
'C:\dev-MSVC-PR\swipl-devel\build\CMakeFiles\ef26a058d73045c5cc7a626820c0b1c6\manindex.db.rule;C:\dev-MSVC-PR\swipl-devel\build\CMakeFiles\3756dba4d1b419f94b58a6a1dab534db\man_index.rule;C:\dev-MSVC-PR\swipl-devel\CMakeLists.txt'
exited with code -1073741819.
[C:\dev-MSVC-PR\swipl-devel\build\man_index.vcxproj]

JanWielemaker:
> If you need the details

That is merely the resulting crash during process termination.    Anyway, do you make a PR or should I simply update the check in Config.cmake as you propose?    If that works, I'm ok with it.   Normally I'd try to figure out why it does not work to use the function check by examining the CMake error log.    The difference is that the symbol exists merely checks (I think) the header and verifies the symbol is there, while the function check writes a minimal program and verifies this can be compiled and linked.

EricGT:
The noted fix is not working as expected.

Currently testing this variation

# PThreads4W from vcpkg needs its include dir for
check_symbol_existsif(TARGET PThreads4W::PThreads4W)
  get_target_property(_ptw32_inc PThreads4W::PThreads4W
INTERFACE_INCLUDE_DIRECTORIES)
  list(APPEND CMAKE_REQUIRED_INCLUDES ${_ptw32_inc})endif()
check_symbol_exists(pthread_win32_process_detach_np "pthread.h"
HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)

EricGT:
> the function check writes a minimal program and verifies this can be
compiled and linked.

That was done at one point by Claude in my research yesterday.

I can note this to Claude and see if that works. Not sure if this will work
with MSYS2 builds.

EricGT:
This did a successful debug build.

# PThreads4W: check_symbol_exists needs both the include dir and library.
# vcpkg's PThreads4W uses imported targets, so CMAKE_REQUIRED_INCLUDES
# and CMAKE_REQUIRED_LIBRARIES don't automatically include them.
if(TARGET PThreads4W::PThreads4W)
  set(_ptw32_saved_libs ${CMAKE_REQUIRED_LIBRARIES})
  set(_ptw32_saved_incs ${CMAKE_REQUIRED_INCLUDES})
  list(APPEND CMAKE_REQUIRED_INCLUDES ${PThreads4W_INCLUDE_DIR})
  list(APPEND CMAKE_REQUIRED_LIBRARIES ${PThreads4W_LIBRARY_RELEASE})
  check_symbol_exists(pthread_win32_process_detach_np "pthread.h"
HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)
  set(CMAKE_REQUIRED_LIBRARIES ${_ptw32_saved_libs})
  set(CMAKE_REQUIRED_INCLUDES ${_ptw32_saved_incs})
else()
  check_symbol_exists(pthread_win32_process_detach_np "pthread.h"
HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)
endif()

Will hold off on sending a PR at this time.

---

Next I will try doing function check, writing a minimal program and
verifying this can be compiled and linked.

EricGT:
This is what Claude created

# PThreads4W: check_function_exists() fails with vcpkg's imported targets
# because the include directory and library are not in CMAKE_REQUIRED_*.
# Use check_c_source_compiles() to verify the function can be compiled and
# linked, which is a stronger guarantee than a symbol-exists check.
if(TARGET PThreads4W::PThreads4W)
  set(_ptw32_saved_libs ${CMAKE_REQUIRED_LIBRARIES})
  set(_ptw32_saved_incs ${CMAKE_REQUIRED_INCLUDES})
  list(APPEND CMAKE_REQUIRED_INCLUDES ${PThreads4W_INCLUDE_DIR})
  list(APPEND CMAKE_REQUIRED_LIBRARIES ${PThreads4W_LIBRARY_RELEASE})
  check_c_source_compiles(
      "#include <pthread.h>
       int main(void) { pthread_win32_process_detach_np(); return 0; }"
      HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)
  set(CMAKE_REQUIRED_LIBRARIES ${_ptw32_saved_libs})
  set(CMAKE_REQUIRED_INCLUDES ${_ptw32_saved_incs})
else()
  check_function_exists(pthread_win32_process_detach_np
                        HAVE_PTHREAD_WIN32_PROCESS_DETACH_NP)
endif()

It is not what I was expecting, is this what you would expect?

PR #1464, opened 2026-02-14
TEST: Verify clean process exit after HTTP server lifecycle

EricGT:
Uses HTTP server start/stop pattern to reliably reproduce threading cleanup issues. Details in commit message.


JanWielemaker:
Integrated into #1463

PR #1466, opened 2026-02-14
FIXED: Revert unify_trie_path to return int

mgondan:
The function may return a non-boolean error code from unify_key.


JanWielemaker:
Thanks.   Never go back though :smile: .  Fixed with ce5f5df5fc1e9857b47a8c7137656d8ba4118174 which uses the `boolex_t` enum to clarify the types.

PR #1469, opened 2026-02-15
TEST: Skip test_thread_exit if alert signal == 0

mgondan:


mgondan:
I open this for review. 

1. It solves a problem in R package rswipl. 
2. I am still a bit puzzled, though. Under linux, patch _not_ included: test_thread_exit passes even if I invoke swipl --sigalert=0.

So, why does this work at all:

````
~/swipl/bin/swipl --sigalert=0 -f none --no-packs -g "use_module(library(http/th
read_httpd)), use_module(library(http/http_dispatch)), http_server(http_dispatch,[port(Port)]), http_stop_server(Port,[]
), prolog_alert_signal(Sig, Sig), writeln(sigalert=Sig)" -t "halt(1)"
% Started server at http://localhost:37973/
sigalert=0
````
..., given that http_stop_server uses thread_signal?

JanWielemaker:
Ok. Thanks.   Merged after some cleanup.   I think that even with ``--sigalert=0`` it can work if the timing is right and the http server is not yet fully initialized when it is asked to stop.    It won't work after the server started waiting for connections (that is, the stop will be honoured on the next connection attempt rather than immediately).

PR #1470, opened 2026-02-15
PORT: Double Debug C-stack to 16MB on Windows

EricGT:
Fixes deep recursion in assert:cyclic_head test.


JanWielemaker:
That is a bit too much code to my liking.   3ba68687e36a863ca529dde8e5a1edf3e9029e8c should address this as it is already done for Asan and when for the WASM versions.

PR #1473, opened 2026-02-17
ADDED: msvc_debug prolog flag for MSVC Debug builds

EricGT:
MSVC Debug builds have higher stack usage due to disabled optimizations. Reduce test depth when running on MSVC Debug to prevent stack overflow in test_ch_shift.


JanWielemaker:
That looks better!   I replaced this with 6738a952959292dcbbeb1eb511c5a1b4aaa1b9b7 and 925e7c4da8e07707cab1e9e8dedeccbef3029029.   The first introduces a Prolog flag `build_type` from CMake.   I want to avoid platform specific flags as much as possible.   Second rewrites the test to use PlUnit and extend the stack depth limitation based on the new flag.

PR #1474, opened 2026-02-17
TEST: added tests for library(record)

kamahen:
I was going to add something to the module but realized it wasn't needed -- but by then I had added some tests.

I've also simplified the code generated for is_<constructor>/1 (I disassembled the two variations and count the same number of opcodes, if a backtrack is counted as an opcode)


JanWielemaker:
Thanks.   Merged after rebase and some layout.   Haven't used this library much since we have dicts :smile:

kamahen:
Does it make sense to provide an operator for records that behaves similarly to dicts? (E.g., `op(100, yfx, (..))` and more-or-less clone the term expansion for `.`). There are some advantages to records, in terms of documentation and, I suspect, performance.

JanWielemaker:
Not that much.  I consider the library outdated.  The PIP meetings work towards an alternative that is more portable based on Ciao and ECLiPSe work and most of its usage is obsolete as I see it now.

PR #1475, opened 2026-02-18
ENHANCED: error/warning messages for swipl-ld

kamahen:
This PR was prompted by some misunderstandings I had. :)

The documentation could still be improved, but it's a bit better.
Note: I've forgotten `nroff` commands, so please check the changes I made to `swipl-ld.1`.

You might not like the way `fmtmsg()` outputs; I can easily change this to use old-fashioned `fprintf()`.

### DIFF
--- a/src/swipl-ld.1
+++ b/src/swipl-ld.1
@@ -15,9 +15,13 @@ information from the SWI-Prolog executable
 after which it scans the arguments and breaks them into several
 categories.  It then calls the C-compiler to create an executable
 containing the user's C-code and the SWI-Prolog kernel.  After this,
+if there are any Prolog files on the command line,
 it will call the development environment to create a Prolog saved
-state from the prolog files and finally it will create the target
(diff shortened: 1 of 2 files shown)


JanWielemaker:
> You might not like the way `fmtmsg()` outputs; I can easily change this to use old-fashioned `fprintf()`.

The main reason for writing swipl-ld in C was portability to Windows.  It seems Windows does not provide fmtmsg() ...

The whole thing should probably be rewritten in Prolog ...   Originally it was also used to build the Prolog packages so we had a bootstrapping issue.   That is no longer the case and it is now only used for user add-ons.

kamahen:
OK, I'll redo it without `fmtmsg()` (which is supposed to be part of the C standard, but whatever).

If this code is redone in Prolog (I'm not volunteering), the changes I've made should help a little bit in understanding the options. :)

JanWielemaker:
Merged (unnoticed by github).

PR #1478, opened 2026-02-25
DONT MERGE: Add assertions

mgondan:
to find undefined behavior

### DIFF
--- a/src/pl-arith.h
+++ b/src/pl-arith.h
@@ -140,6 +140,10 @@ argvArithStack(DECL_LD int n)
   assert(LD->arith.stack.base);
   DEBUG(0, assert(LD->arith.stack.top - n >= LD->arith.stack.base));
 
+  // Note that when the assertion fails, the function is invoked with
+  // n=0, so something like
+  // if(!n) return LD->arith.stack.top;
+  // _may_ be an option
   return LD->arith.stack.top - n;
 }
 


JanWielemaker:
I'm a little lost.    Always important is the original issue that triggered this.  How can I reproduce that?

mgondan:
I guess with an Apple M1 and UBSAN. I tried it here, with no luck 

https://github.com/mgondan/rswipl/actions/runs/22416039750

The other one in pl-arith is easy

JanWielemaker:
For both I have not yet found the original message and compilation flags/platform.   I do have an M1, so that is no problem.    I'm probably blind :smile:

mgondan:
I’ll give you more advice in the later afternoon

mgondan:
For pl-arith.h: Just add the assertion from the PR, and then, under Linux,

cmake ..
make
ctest

You should see the assertion fail in swipl:tabling.

mgondan:
For bf_gmp.c, the cmake flags are found here,

https://github.com/mgondan/rswipl/blob/main/src/Makevars (unsure if USE_GMP=ON or OFF).

Some more configuration options

https://www.stats.ox.ac.uk/pub/bdr/M1-SAN/README.txt

The error is found here

https://www.stats.ox.ac.uk/pub/bdr/M1-SAN/rswipl/tests/testthat.Rout

To build the R package:

````
$ R
> install.packages(c("Rcpp", "testthat")) [maybe more packages are needed]
> quit()
$ git clone https://github.com/mgondan/rswipl
$ cd rswipl
$ git submodule update --init --remote --recursive [unsure if remote is needed]
$ cd ..
$ R CMD build rswipl
$ R CMD INSTALL rswipl_10.1.4.tar.gz
$ R CMD check rswipl_10.1.4.tar.gz
````

JanWielemaker:
Ok.  Both should be fixed.   The one in pl-arith.h is harmless.   The one in bf_gmp.c is a real bug, but given the usage of this function, it only affects generating large random numbers.    It probably does impact the distribution of the random numbers.

PR #1482, opened 2026-03-13
BUILD: add flag for building protoc plug-in

kamahen:
It'll be a little while before I finish the PR that uses this flag, but it'd be nice to have it already available when I finish the PR.


kamahen:
This PR became quite out-of-sync during last month (when I was away), so I ended up doing something like https://xkcd.com/1597/ ... so, it would be nice if this PR can be merged, as one less thing to remember. (The changes to `packages/protobufs` will take a bit of time because I'm also documenting the bootstrap process, which I'm having to review because of the the protobufs v34.0 changes.)

PR #1486, opened 2026-04-23
Clicking a pl-file in  Windows 11 Task Bar and current directory becomes c:\windows\system32 problem

xpxaxsxi:
Not sure if all Windows users have this problem but I do. The problem mentioned propably is a NSIS installer problem. I did chat with plain ChatGPT about this fix, and I am pretty convinced this would fix it. 

I am thinking of testing the installer  with the nightly build, so that I won't have to build the whole swipl to fix a small installer problem.


xpxaxsxi:
I tested this  manually editing registry, this does work. If installer writes registry ok, then this will work

PR #1488, opened 2026-05-07
Make library(macros) works with dicts

kwon-young:
I have added a test file to show the effect of this patch.


JanWielemaker:
Thanks.   Looks like a good move.   Will look at the details after current work stabalizes ...

JanWielemaker:
Thanks.  Merged after ditching the test and use `if->then;else` rather than `(a,!;b)`.

PR #1506, opened 2026-07-11
DOC: update sample calc.c to match doc

kamahen:
This replaces #1505.

### DIFF
--- a/man/calc.c
+++ b/man/calc.c
@@ -1,19 +1,30 @@
+#include <stdio.h>
+#include <stdlib.h>
 #include <string.h>
 #include <SWI-Prolog.h>
 
-#define MAXLINE 1024
-
 int
 main(int argc, char **argv)
-{ char expression[MAXLINE];
-  char *e = expression;
+{ char *expression;
   char *program = argv[0];
-  char *plav[2];
-  int n;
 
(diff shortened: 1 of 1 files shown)


JanWielemaker:
Thanks for taking care of this AI slob.    The file is used nowhere, so I decided for a simpler fix by deleting it :smile: Commit d6ef2138f03f8e07d5d73bc394a58f02910a346c

kamahen:
My plan was to have copies of the example code, both to distribute and to run at least a smoke test to verify that there weren't typos. But I  hadn't kept things in sync. :( Or added a smoke test. :( I guess that whoever modifies the documentation should verify that the code behaves as expected -- if we had a lot of examples, something that strips out the samples and runs them  would be nice but that seems like too much work.

On the other, does LaTex have an "include verbatim" facility? If so, we could split out the samples and avoid the duplication problem. (And maybe add some smoke tests)

JanWielemaker:
> does LaTex have an "include verbatim" facility

I think something like that exists.   Would require some serious annotation to automatically test the examples.   Surely worth considering, but not on my shortllist.

kamahen:
The problem is -- when I update an example, I often forget to test it. :(  I suppose my code reviewer needs to be more vigilant. :)

PR #1510, opened 2026-07-15
Optionally use CORE-MATH correctly rounded elementary functions

ridgeworks:
Under control of CMake option USE_CRMATH (default=OFF), build a local copy of library containing a subset of CORE-MATH functions from repo https://github.com/indigobio/core-math-binary64.

If option is OFF, use existing explicitly rounded functions currently in pl-arith.c. (under control of O_ROUNDING_UP_DOWN). 

Caveats:
1. Only tested on MacOS, X86 architecture.
2. compile flags for local crmath library set to "-O3 -ffp-contract=on -fno-math-errno" as per external repo. Is this right for all architectures/compilers. (Note: no support for different FP rounding in WASM).
3. Limited CMake experience so additions should be reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,9 @@ option(USE_SIGNALS
 option(USE_GMP
        "Use GNU MP Bignum library (LGPL)"
        ON)
+option(USE_CRMATH
+       "Use correctly rounded CORE_MATH elementary functions"
+       OFF)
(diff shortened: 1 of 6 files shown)


JanWielemaker:
Merged after some refactoring.  Notably reuse the external dependency with as little as possible own code and pin the version by git HASH for security and reproducibility purposes.

PR #1522, opened 2026-09-11
CMAKE: Fix C/C++ flags for release/debug/sanitize

mgondan:
The changes are pretty straightforward, but you might wish to check one thing: Whenever I saw $ENV{CXXFLAGS} in statements like

````
  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CXX_OPTFLAGS} ${CC_DBGFLAGS} $ENV{CXXFLAGS}"
      CACHE STRING "CXXFLAGS for a RelWithDebInfo build" FORCE)
````

I added the same to the C_FLAGS version:

````
 set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CC_OPTFLAGS} ${CC_DBGFLAGS} $ENV{CFLAGS}"
    etc.
````

This _may_ duplicate the CFLAGS if they are already into CC_OPTFLAGS in the beginning:

````
if(DEFINED ENV{CFLAGS})
  string(REGEX MATCH "-O" matchc $ENV{CFLAGS})
endif()

if(matchc)
  set(CC_OPTFLAGS $ENV{CFLAGS})
else()
````

I think it is harmless, but if it isn't, that copy-logic should be replaced by some other mechanism.


mgondan:
https://github.com/SWI-Prolog/swipl-devel/commit/4073cf478cff1e0dd45cf41febdbd96517eb282e

Closing this

Issue #1526, opened 2026-09-13
MSYS2: ctest fails in core_lang


mgondan:
The problem seems to appear between https://github.com/SWI-Prolog/swipl-devel/commit/ba4d1d320bd4e9a58b36becdd2718739717daaba and https://github.com/SWI-Prolog/swipl-devel/commit/94c6691f6e8353d595750ccb30e979bc4067782b (when pl-bisim.c and .h were rewritten).

````
% [6/68] bisim:oracle_bisimilar ..................... passed (0.000 sec)
% [7/68] bisim:oracle_hydra ......................... passed (0.000 sec)
% [8/68] bisim:oracle_distinct ...................... passed (0.000 sec)
% [9-1/68] bisim:quotient ..

0% tests passed, 1 tests failed out of 1
Total Test time (real) =   1.32 sec
The following tests FAILED:
          4 - swipl:core_lang (Exit code 0xc0000374)
````

The problem seems to be in term_automaton, e.g. a simple call

?- term_automaton(f(a, b), A).

crashes the program.

mgondan:
````
warning: Critical error detected c0000374

Thread 1 received signal SIGTRAP, Trace/breakpoint trap.
0x00007ffaed0fd583 in ntdll!RtlIsZeroMemory () from C:\windows\SYSTEM32\ntdll.dll
(gdb) bt
#0  0x00007ffaed0fd583 in ntdll!RtlIsZeroMemory () from C:\windows\SYSTEM32\ntdll.dll
#1  0x00007ffaed10665a in ntdll!.misaligned_access () from C:\windows\SYSTEM32\ntdll.dll
#2  0x00007ffaed10693a in ntdll!.misaligned_access () from C:\windows\SYSTEM32\ntdll.dll
#3  0x00007ffaed112935 in ntdll!.misaligned_access () from C:\windows\SYSTEM32\ntdll.dll
#4  0x00007ffaed02c32d in ntdll!RtlGetCurrentServiceSessionId () from C:\windows\SYSTEM32\ntdll.dll
#5  0x00007ffaed02b041 in ntdll!RtlFreeHeap () from C:\windows\SYSTEM32\ntdll.dll
#6  0x00007ffaeaa3364b in ucrtbase!_free_base () from C:\windows\System32\ucrtbase.dll
[...]

PR #1527, opened 2026-09-13
Update pl-bisim.c

mgondan:
This PR illustrates the point made in 

https://github.com/SWI-Prolog/swipl-devel/issues/1526


mgondan:
Case closed

